PR #45, opened 2020-02-28
Analytics

nsolov:


nsolov:
fix all, except "config keyword may be omited for get and set commands"
we already have 1 command (module-action) that could be omited

src/cls/_ZPM/PackageManager/Client/REST/PackageManagerClient.cls
```diff
@@ -106,6 +110,60 @@ Method GetModuleManifest(pModuleReference As %ZPM.PackageManager.Core.ResolvedMo
...
+        $$$ThrowStatus($$$ERROR($$$GeneralError,"Error Sending Analytics Data"))
+    }
+}
+
+Method SendAnalyticsToGoogle(action As %String, data As %DynamicObject)
+{
+    Set TrackingID = ##class(%ZPM.PackageManager.Settings).GetAnalyticsTrackingId()
+    Quit:(TrackingID = "") 0
+    Set req = ##class(%Net.HttpRequest).%New()
+    Set req.Server = "www.google-analytics.com"
```
daimor: should not it use SSL?

nsolov: I've added SSL, and fix one more bug invoking nonexistent class

PR #55, opened 2020-03-24
Support multiple test resources and running individual test cases/suites

timleavitt:
This fixes a bug I introduced in #25 (and a possible cross-OS compatibility issue that would cause issues running a targeted test suite on Linux)

Running a targeted test suite/case will now work, running the test exactly once, even if there are multiple test resources present. (The incorrect behavior was to not run it at all, vs. running it twice before the changes in #25.)

I've tested with a simple repo (objectscript-math) as well as one I have coming that has two separate unit test resources (both running in the test phase).


timleavitt:
@evshvarov I found this while working on the Developer Community article I've been promising you for a while

evshvarov:
Thanks Tim!
That shows again how extremely helpful DC is!)

вт, 24 марта 2020 г. в 19:59, Tim Leavitt <notifications@github.com>:

PR #65, opened 2020-04-10
transfer parameters

nsolov:


src/cls/_ZPM/PackageManager/Developer/Processor/Abstract.cls
```diff
@@ -224,6 +225,26 @@ Method %Evaluate(pAttrValue) As %String [ Internal ]
...
+	Set key = ""
+	For {
+		Set attribute = pAttributes.GetNext(.key)
+		Quit:(key="")
+		Do pAttributes.SetAt(##class(%ZPM.PackageManager.Developer.ModuleSetting.Default).EvaluateAttribute(attribute,.customParams),key)
+		// try to set a Property
+		Try {
+			Set val = $Property($this,key)
+			Continue:($IsObject(val)||$Listvalid(val))
+			Set $Property($this,key) = ..%Evaluate(pAttributes.GetAt(key))
```
daimor, line 240: So, you just set some else value?
$Property($this,key) had already some value, but you set there something from `pAttributes.GetAt(key)`
looks like it should change the value only in case if it found

Issue #68, opened 2020-04-12
Restore verify lificycle phase for zpm itself


daimor:
command `zpm verify` should successfully test ZPM itself with integration tests.

zpm verify should be also included in CI

PR #69, opened 2020-04-12
#68 Restored verify lifecycle for ZPM

daimor:


timleavitt:
@daimor this looks good to me other than the issues on 2020.1; any idea what's up there? Might be worth adding verbose output? We also support junit test result reporting - not sure what the best way to pick up test results would be in GitHub Actions though.

daimor:
Yeah just polishing it for GitHub actions, locally it works well now.
JUnit is cool, yes, it needs some investigation on how to deal with here.

PR #257, opened 2021-09-13
Added an ability to extend repo command by external packages

daimor:
With this update, it is now possible to make a new external package, which would be able to add a new type of repository

As an example, good to look at [RemoteServer](https://github.com/intersystems-community/zpm/blob/5575365b85/src/cls/_ZPM/PackageManager/Client/RemoteServerDefinition.cls) implementation

The class of a new realization requires extending `%ZPM.PackageManager.Client.ServerDefinition` and  XData Commands which will extend the existing repo command 

```
XData Commands [ XMLNamespace = "http://www.intersystems.com/PackageManager/CLI" ]
{
<commands>
<command name="repo">
<group name="GitHub repositories">

<modifier name="git" aliases="github" description="Create/update a github server repository" />
<modifier name="token" value="true" description="Token to access private repositories" />

<example description="Switch to test registery">
repo -git -n github -token sometoken
</example>

</group>
</command>
</commands>
}
```

Implements #251


isc-tleavitt:
This is excellent.

daimor:
version for testing is available here
https://github.com/intersystems-community/zpm/releases/tag/v0.3.1-dev.257.1

PR #298, opened 2022-02-11
Renaming to %IPM

daimor:


isc-tleavitt:
Work in the v1 branch covers the same ultimate result.

PR #313, opened 2022-05-02
HSIEO-5485: Add Generic classes back to IPM after adopting name change

Beiming-Zhang:
Source-of-truth of class renaming for IPM package: https://docs.google.com/spreadsheets/d/1zyvoBVxKC9zBa6m9yy_XkW3O-jxapJWq/edit#gid=1974853703


isc-kiyer:
@Beiming-Zhang Please link the google doc here as well

src/%IPM/ResourceProcessor/Default/LocalizedMessages.cls
```diff
@@ -60,10 +60,29 @@ Method OnConfigureMappings(ByRef pParams) As %Status
 		Set tGlobalScope = ..ResourceReference.Module.GlobalScope && '$Get(pParams("Reload","ForceLocalScope"),0)
 		If 'tGlobalScope {
 			Set tSourceDB = ##class(%IPM.Utils.Module).GetRoutineDatabase($Namespace)
-			Set tSC = ##class(%IPM.Utils.Module).AddRoutineMapping($namespace,tName,,tSourceDB)
+			// Check if current file exists in the the routine database of the current namespace exists in current namespace('s database)
```
isc-kiyer, line 63: Again use %SYS.Namespaces methods for this

src/%IPM/Utils/Module.cls
```diff
@@ -1149,11 +1149,11 @@ ClassMethod LoadNewModule(pDirectory As %String, ByRef pParams, pRepository As %
 		If $Get(pParams("DeveloperMode"),0) {
 			Set tCurrentSourceControl = ##class(%Studio.SourceControl.Interface).SourceControlClassGet()
 			If (tCurrentSourceControl = "") {
-				Set tCompositeSCClass = "%IPM.PackageManager.Developer.Extension.Composite"
+				Set tCompositeSCClass = "%IPM.ExtensionBase.Composite"
 				$$$ThrowOnError(##class(%Studio.SourceControl.Interface).SourceControlClassSet(tCompositeSCClass))
-				Set tPrimaryExtension = "%IPM.PackageManager.Developer.Extension.SourceControl.ISC"
-				Set tSecondaryExtensions = $ListBuild("%IPM.PackageManager.Developer.Extension.PackageManager")
-				$$$ThrowOnError(##class(%IPM.PackageManager.Developer.Extension.Composite).SetExtensionClasses(tPrimaryExtension,tSecondaryExtensions))
+				Set tPrimaryExtension = "%IPM.Extension.SourceControl.Perforce.Main"
```
isc-kiyer, line 1154: Should not reference these classes here as they are in zpm-perforce and zpm-studio-ui repos

isc-kiyer (changes requested):
My comments are up

src/%IPM/Main.cls
```diff
@@ -991,9 +991,9 @@ ClassMethod Init(ByRef pCommandInfo) [ Internal ]
 	Set tSCClassChanged = 0
 	
 	Set tPrimaryClass = ""
-	Set tCompositeSCClass = "%IPM.PackageManager.Developer.Extension.Composite"
-	Set tPackageMgrSCClass = "%IPM.PackageManager.Developer.Extension.PackageManager"
-	Set tSCInterface = "%IPM.PackageManager.Developer.Extension.SourceControl.Interface"
+	Set tCompositeSCClass = "%IPM.ExtensionBase.Composite"
+	Set tPackageMgrSCClass = "%IPM.Extension.ServerMenu.MenuItems"
```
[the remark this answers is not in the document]
Beiming-Zhang: Yeah this class is for setting the package manager extension, it is used as the default extension when ipm is initializing, and secondary extension if another extension has been set primary. I think we can keep it since SetExtensionClass() can ignore non-existing class. But can also create another jira to completely remove the default logic from this class to perhaps an installer class of ipm-studio-ui?

src/%IPM/ExtensionBase/CompositeMethodOverrides.cls
```diff
@@ -0,0 +1,367 @@
...
+			}
+			If '$Data(tClassNames(tClassName)) {
+				Set tClassNames = tClassNames_$ListBuild(tClassName)
+				Set tClassNames(tClassName) = ""
+			}
+		}
+		If $$$ISERR(tStatus) {
+			Quit
+		}
+		Set @..#SUBEXTENSIONGBL = pClassNames
```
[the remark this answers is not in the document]
Beiming-Zhang: Ah yes I already made this change in (Port recent HSPM changes) review, but making in here as well

src/%IPM/ExtensionBase/SourceControl/Interface.cls
```diff
@@ -1,18 +1,20 @@
 /// Additional inteface for source control extensions to implement to plug in to module lifecycles.
-Class %IPM.PackageManager.Developer.Extension.SourceControl.Interface [ Abstract ]
+Class %IPM.ExtensionBase.SourceControl.Interface [ Abstract ]
 {
 
 /// Name of page to help configure source control on a per-module basis
 Parameter SETTINGSPAGE;
 
+Parameter NOPREFIXCHARCODE = 8;
```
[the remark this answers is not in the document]
Beiming-Zhang: This is from %IPM.Extension.SourceControl.Perforce.Main, which is in zpm-perforce, but there are places referring it in zpm which I already changed, so should move the parameter to this class as well
image (In %IPM.Lifecycle.Base.cls)

src/%IPM/ResourceProcessor/Default/Class.cls
```diff
@@ -7,20 +7,40 @@ Parameter DESCRIPTION As STRING = "Standard resource processor for class definit
...
 	Set tSC = $$$OK
 	Try {
 		Set tName = $Piece(..ResourceReference.Name,".",1,*-1)
 		Set tGlobalScope = ..ResourceReference.Module.GlobalScope && '$Get(pParams("Reload","ForceLocalScope"),0)
 		If 'tGlobalScope {
-			Set tPackage = $p(tName,".",1,*-1)
 			Set tSourceDB = ##class(%IPM.Utils.Module).GetRoutineDatabase($Namespace)
-			Set tSC = ##class(%IPM.Utils.Module).AddPackageMapping($namespace,tPackage,tSourceDB)
+			Set tPackage = $p(tName,".",1,*-1)
+			// Check if current file exists in the the routine database of the current namespace exists in current namespace('s database)
```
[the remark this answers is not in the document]
Beiming-Zhang: Created https://usjira.iscinternal.com/browse/HSIEO-6334 to handle this as I want to do some testing to make sure nothing breaks after we removing this logic

src/%IPM/Main.cls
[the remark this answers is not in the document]
Beiming-Zhang: EDIT: Resolved now by adding super class %IPM.ExtensionBase.ServerMenu.Interface

src/%IPM/Utils/Module.cls
Beiming-Zhang: Created https://usjira.iscinternal.com/browse/HSIEO-6334 to handle this

src/%IPM/ResourceProcessor/Default/LocalizedMessages.cls
Beiming-Zhang: Created https://usjira.iscinternal.com/browse/HSIEO-6334 to handle this

Beiming-Zhang:
@evshvarov  @daimor  This PR has been reviewed by HSOPS team. 
Since this work is Time-Critical and this review has been up for more than a month with no feedback from community PM team, I will merge this Pull Request by the end of tomorrow.
Please give feedback before then if you would like to. Thank you!

Beiming-Zhang:
> It's not so good, to have it merged so quickly

We have this review up for more than a month, and did not receive any feedback on it. Our fall release need it, and so does DBP's pending work.
Please do some review and provide feedback if you would like to. Otherwise, we need to merge it by the end of tomorrow. As this is really time-restricted.
Cc @isc-kiyer

Installer.cls
```diff
@@ -18,14 +18,11 @@ XData PM [ XMLNamespace = INSTALLER ]
 <Default Name="REGISTRY" Value="https://pm.community.intersystems.com" />
 <Default Name="TRACKINGID" Value="UA-126752118-1" />
 
-<Namespace Name="%ALL" Code="%DEFAULTDB" Data="%DEFAULTDB" Create="yes" Ensemble="0">
+<Namespace Name="${CURNS}" Code="${CURNSRoutineDB}" Data="${CURNSGlobalDB}" Create="no" Ensemble="0">
```
daimor, line 21: Package Manager should be available from any Namespace, there is no reasons at all to have multiple versions of zpm in different namespaces

Installer.cls
```diff
@@ -48,6 +45,13 @@ ClassMethod setup(ByRef pVars, pLogLevel As %Integer, pInstaller As %Installer.I
   #; Let our XGL document generate code for this method. 
   Do %code.WriteLine(" If $Data(pVars, path)#10,path'="""" Set pVars(""MODDIR"")=path")
   Do %code.WriteLine(" Set pVars(""CURNS"")=$namespace")
+  Do %code.WriteLine(" Set prevNS = $namespace")
+  Do %code.WriteLine("  New $Namespace")
+  Do %code.WriteLine("  Set $Namespace = ""%SYS"" ")
+  Do %code.WriteLine("  kill properties")
+  Do %code.WriteLine("  Set sc = $ClassMethod(""Config.Namespaces"", ""Get"", prevNS, .properties)")
```
daimor, line 52: really, what's wrong with this way?
`##class(Config.Namespaces).Get(prevNS, .properties)`

evshvarov:
Hi Beiming!
Sorry for the delay.
But we really need to sort out the idea of different versions of IPM in
different namespaces. Can we provide any use case when we 100% need it?
Another issue is to install IPM for every new namespace - this could slow
down the IRIS launch e.g. for serverless scenarios where every second
becomes meaningful.
Can we discuss it again, please?

Installer.cls
daimor: Looks like, community opinion means nothing
why would you need our review then?

daimor: There is no way, that it will help current zpm users, would like to use this project
no migration, too many changes for them, they would need to delete all their packages installed to install it again

src/%IPM/Utils/Module.cls
```diff
@@ -1118,15 +1118,15 @@ ClassMethod LoadNewModule(pDirectory As %String, ByRef pParams, pRepository As %
 		Set tSC = $system.OBJ.Load(pDirectory_"module.xml",$Select(tVerbose:"d",1:"-d"),,.tLoadedList)
 		$$$ThrowOnError(tSC)
 		
-		Set tFirstLoaded = $$$lcase($Order(tLoadedList("")))
+		Set tFirstLoaded = $Order(tLoadedList(""))
```
daimor, line 1121: why it was changed here?

src/%IPM/CLI.cls
```diff
@@ -66,279 +65,273 @@ ClassMethod %Help(ByRef pCommandInfo) [ Final ]
...
 		If 'tVerbose {
-			Write !, "For more detail, run:", !?2, "help <command-name>", !, "or", !?2, "help -v"
+			Write !!, "For more detail, run:", !?2, "help <command-name>", !, "or", !?2, "help -v"
 		}
 	}
 	Write !
 }
 
-ClassMethod %HelpForCommand(pCommandName As %String, ByRef pCommandStruct, pDetailed As %Boolean = 0, pMarkdownFormat As %Boolean = 0)
+ClassMethod %HelpForCommand(pCommandName As %String, ByRef pCommandStruct, pDetailed As %Boolean = 0, pMarkdownFormat As %Boolean = 0, pMaxLength As %Integer = 0)
```
daimor, line 80: really, all these changes really necessary?
and removing added features, also necessary?

src/%IPM/Utils/Module.cls
Beiming-Zhang: This is to make sure that packages are not recognized all in lower case. They should be kept in their original text when installed by IPM

daimor: zpm packages have to be lowercase only, there are no reasons to change it, there were a few reasons, why we come to this decision

daimor:
So, it's I see that these changes needed for InterSystems, and with no care at all about the existing hundreds of packages
many braking changes, makes it to have a lot of pain for many developers who already used zpm

src/%IPM/Utils/Module.cls
Beiming-Zhang: Why zpm packages needs to be lowercase only?
We have a good reason to change it since package name themselves should be kept in their original form.

Beiming-Zhang:
> So, it's I see that these changes needed for InterSystems, and with no care at all about the existing hundreds of packages many braking changes, makes it to have a lot of pain for many developers who already used zpm

Dmitry, we do CARE about existing packages, and we are working on upgrade steps and testings to make sure that existing users would have a smooth adoption. The feature changes we make are the necessary steps for supporting internal customers (IRIS, HS, I4H) with IPM, and in the long term, they will definitely benefit external users.

src/%IPM/Utils/Module.cls
daimor: naming convention
every where it should be normalized to one form, for search, for store any typo, should still be one case
dependency and so on
any reasons to break it?

src/%IPM/CLI.cls
```diff
@@ -66,279 +65,273 @@ ClassMethod %Help(ByRef pCommandInfo) [ Final ]
...
+					Write $$$EscapeOrFormat("[", $$$Red)
+					Set tEndingBrackets = tEndingBrackets + 1
+				}
+				Write $$$EscapeOrFormat("<"_tParamName_">", $$$Red)
 			}
 			
-			Write " "
-			// Non-required parameters are surrounded with [] brackets
-			If 'pCommandStruct("parameters", tParamName, "required") {
-				Write $$$EscapeOrFormat("[", ..TerminalPromptColor())
```
daimor, line 134: > I don't think we removed any features from Community PM other than rewrite codes that are not aligned.

yes, you removed, and not only in this place

src/%IPM/ResourceProcessor/Abstract.cls
```diff
@@ -225,35 +225,40 @@ Method OnGetUniqueName(Output pUniqueName)
...
+		Set attrValue = ##class(%IPM.Utils.Module).%RegExReplace(attrValue,"(?i)\{\$root\}",root)
 	}
-
-  Quit ..ResourceReference.Module.%Evaluate(tAttrValue)
+	If $Locate(attrValue,"(?i)\{name\}") {
+		Set name = ..ResourceReference.Name
+		Set attrValue = ##class(%IPM.Utils.Module).%RegExReplace(attrValue,"(?i)\{name\}",name)
+	}
+	Return attrValue
 }
```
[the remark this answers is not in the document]
daimor: Important to remove what was here before you?
this is really breaking change.
just removed important parts of code

src/%IPM/Utils/Module.cls
Beiming-Zhang: The naming convention should be kept to its original name, without casting to lower case. If any package's name breaks things by casting to lower case, that's a bug in other places that we need to fix, not to fix the name itself.

daimor: Who is decided so?

daimor: This project was driven by Community, and rules was set by community, by reasons,
And InterSystems, just don't care about it

isc-tleavitt (changes requested):
I've offered some unsolicited feedback, hopefully bridging the HealthShare and community perspectives a bit.

Installer.cls
isc-kiyer: We plan to add some code to the IPM installer to make migration easier (migrating all the storage globals as needed). The goal is to have a smooth transition from ZPM to IPM with no additional work required from consumers of the package manager.

Installer.cls
isc-kiyer: I agree. @Beiming-Zhang In generated code, can always use direct ##class

src/%IPM/CLI.cls
```diff
@@ -50,7 +50,6 @@ ClassMethod %Help(ByRef pCommandInfo) [ Final ]
 		// List commands
 		If 'tMarkdown {
 			Write !, $$$FormattedLine($$$Underlined, "Available commands:")
-			Write !, "NOTE: [] around a parameter indicates it is optional"
```
isc-kiyer, line 53: @Beiming-Zhang For this class in particular (and perhaps some other classes as well), for this merge, could you revert to using the existing indentation so it is easier to see the diffs? Could fix just the indentation across the repo in a separate PR

src/%IPM/CLI/Commands.cls
```diff
@@ -20,136 +20,141 @@ XData Commands [ XMLNamespace = "http://www.intersystems.com/PackageManager/CLI"
 }
 
 /// This is the Schema which defines the form of the Commands XData block
-XData Schema [ Internal ]
+XData Schema [ Internal, MimeType = application/xml ]
 {
 <?xml version="1.0"?>
-<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema" >
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">
```
isc-kiyer, line 26: @Beiming-Zhang  Again please revert to existing indentation for the merge

src/%IPM/DependencyAnalyzer.cls
```diff
@@ -2,6 +2,14 @@ Include %IPM.Common
...
 Class %IPM.DependencyAnalyzer
 {
+/// Prefix to indentify .inc files
+Parameter INCPREFIX = "!!!INC.";
+
+/// Suffix for different file types
+Parameter INCSUFFIX = ".INC";
+
+Parameter CLSSUFFIX = ".CLS";
+
```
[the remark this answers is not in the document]
isc-kiyer: These are meant to be used by the dependency analyzer. @Beiming-Zhang Did all the dependency analyzer changes not make it into this merge?

src/%IPM/ResourceProcessor/Abstract.cls
```diff
@@ -225,35 +225,40 @@ Method OnGetUniqueName(Output pUniqueName)
...
+/// 	<li>{name} - the name of the resource </li>
+/// </ul>
+/// In addition to these, look at expressions supported by %EvaluateSystemExpression
+/// in class <class>%IPM.Utils.Module</class>.
+Method %Evaluate(pAttrValue As %String) As %String [ Internal ]
 {
-	Set tAttrValue = pAttrValue
-	If $Data(..Params) {
-		Merge tParams = ..Params
-		Do ##class(%IPM.Storage.ModuleSetting.Default).GetCustomParameters(.customParams,..ResourceReference.Module, .tParams)
```
isc-kiyer, line 247: This seems like a merge issue. Lost this logic at the beginning of the method before %Evaluate. As it was, custom parameters should get evaluated before the defaults below of the OOTB parameters/expressions

src/%IPM/Storage/ResourceReference.cls
```diff
@@ -120,7 +120,8 @@ ClassMethod GetChildren(pResourceName As %String, pModuleName As %String, pCheck
 					Set tResourceCache(tCacheResult.%Get("UniqueName")) = ""
 				}
 			}
-			
+			// include hidden files
+			Set ^%SYS("Studio","ShowHidden") = 1
```
isc-kiyer, line 124: If this is a global setting, we should not be setting it and leaving it as is. Instead, should record the previous value, set it to 1 before the query and reset it to the previous value after the query

src/%IPM/Common.inc
```diff
@@ -6,10 +6,10 @@ ROUTINE %IPM.Common [Type=INC]
 #define ZPMStudioDocumentModule %ZPMStudioDocumentModule
 
 #; Local % Variable to track modules already handled for a given end goal
-#define ZPMHandledModules %ZZPMHandledModules
+#define ZPMHandledModules %ZPMHandledModules
```
isc-kiyer, line 9: The values here can all be changed to IPM instead of ZPM and IRIS.Temp instead of CacheTemp (the names of the macros can remain as ZPM to avoid extra work of replacing all those references for now).

Installer.cls
daimor: When it will happen?
History with IRIS says, it requires more than a year, to get some migration tool

may we keep at least globals the same,
at this time, users with installed packages from ZPM, after upgrade will loose control over installed packages, and will require to reinstall everything

src/%IPM/CLI.cls
daimor: This class is just a help generation, there should not be many important changes from HS
But, you just completely replaced it, with no care of changes

src/%IPM/ResourceProcessor/Default/Package.cls
```diff
@@ -28,10 +28,29 @@ Method OnConfigureMappings(ByRef pParams) As %Status
 		Set tGlobalScope = ..ResourceReference.Module.GlobalScope && '$Get(pParams("Reload","ForceLocalScope"),0)
 		If 'tGlobalScope {
 			Set tSourceDB = ##class(%IPM.Utils.Module).GetRoutineDatabase($Namespace)
-			Set tSC = ##class(%IPM.Utils.Module).AddPackageMapping($namespace,tName,tSourceDB)
+			// Check if current file exists in the the routine database of the current namespace exists in current namespace('s database)
+			// If yes, skip creating unnecessary mappings
+			// (Spec,Dir=1,OrderBy=1,SystemFiles=1,Flat,NotStudio=0,ShowGenerated=0,Filter,RoundTime=0,Mapped=0)
+			// Set mapped=0 since we only want to check whether the file exists in the current routine DB
+			Set tResult = ##class(%Library.RoutineMgr).StudioOpenDialogFunc("tName.*",1,1,1,,,,,,0)
```
daimor, line 35: "tName.*"?

src/%IPM/ResourceProcessor/Default/Routine.cls
```diff
@@ -27,10 +27,29 @@ Method OnConfigureMappings(ByRef pParams) As %Status
 		Set tGlobalScope = ..ResourceReference.Module.GlobalScope && '$Get(pParams("Reload","ForceLocalScope"),0)
 		If 'tGlobalScope && '..LoadAsDirectory {
 			Set tSourceDB = ##class(%IPM.Utils.Module).GetRoutineDatabase($Namespace)
-			Set tSC = ##class(%IPM.Utils.Module).AddRoutineMapping($namespace,tName,,tSourceDB)
+			// Check if current file exists in the the routine database of the current namespace exists in current namespace('s database)
+			// If yes, skip creating unnecessary mappings
+			// (Spec,Dir=1,OrderBy=1,SystemFiles=1,Flat,NotStudio=0,ShowGenerated=0,Filter,RoundTime=0,Mapped=0)
+			// Set mapped=0 since we only want to check whether the file exists in the current routine DB
+			Set tResult = ##class(%Library.RoutineMgr).StudioOpenDialogFunc(tName_".*",1,1,1,,,,,,0)
```
daimor, line 34: "tName.*"
The same piece of code, used multiple times, if we really need it, it has to be moved outside.

src/%IPM/ExtensionBase/ServerMenu/Interface.cls
```diff
@@ -0,0 +1,25 @@
+/// Additional inteface for server menu source control extensions to implement.
+Class %IPM.ExtensionBase.ServerMenu.Interface Extends %Studio.Extension.Base [ Abstract ]
```
daimor, line 2: Why we still add Studio support?

src/%IPM/ExtensionBase/Composite.cls
```diff
@@ -0,0 +1,339 @@
+/// This class provides the API and interface to work with multiple Studio extensions
+/// It directs the action and lifecycle calls to corresponding primary/secondary extension class
+/// Set Inheritance to right, so that on conflict definitions, CompositeMethodOverrides.cls will take over %Studio.Extension.Base.cls
+/// This way, CompositeMethodOverrides.cls can override final methods in %Studio.* methods
+Class %IPM.ExtensionBase.Composite Extends (%Studio.Extension.Base, %IPM.ExtensionBase.CompositeMethodOverrides) [ Inheritance = right, Owner = {%Developer} ]
```
daimor, line 5: Why this is part of IPM?

src/%IPM/DependencyAnalyzer/Problem.cls
```diff
@@ -52,10 +55,13 @@ Storage Default
...
 <Value name="8">
-<Value>ReferencedClass</Value>
+<Value>ReferencedItem</Value>
 </Value>
 <Value name="9">
-<Value>ReferencedClassExists</Value>
+<Value>ReferencedFileType</Value>
+</Value>
+<Value name="10">
+<Value>ReferencedItemExists</Value>
```
daimor, line 64: Why the order of properties is changed?
ReferencedClassExists was replaced by ReferencedItemExists
and place is changed from 9 to 10

src/%IPM/Lifecycle/Base.cls
```diff
@@ -492,8 +492,7 @@ Method %Reload(ByRef pParams) As %Status
 			Set tSC = $$$ERROR($$$GeneralError,"module.xml is malformed.")
 			Quit
 		}
-		set first = $$$lcase(first)
-    If ($Piece(first,".",*) '= "zpm") {
+    	If ($Piece(first,".",*) '= "ZPM") {
```
daimor, line 495: return back lcase

src/%IPM/Repo/Filesystem/Definition.cls
```diff
@@ -39,6 +39,7 @@ repo -name LocalFiles -snapshots 1 -fs -depth 2 -path C:\MyWorkspace\RootModuleD
 </commands>
 }
 
+
```
daimor, line 42: ?

src/%IPM/Repo/Filesystem/Definition.cls
```diff
@@ -259,7 +260,7 @@ Method GetModuleStreamFromFile(pFilename As %String, Output pStream As %Stream.G
 		$$$ThrowOnError(tSC)
 		
 		Set tMetaStream.LineTerminator = $Select($$$isWINDOWS:$c(13,10),$$$isUNIX:$c(10))
-		Set pName = $ZConvert(tMetaStream.ReadLine(), "L")
+		Set pName = tMetaStream.ReadLine()
```
daimor, line 263: why?

src/%IPM/Repo/Definition.cls
```diff
@@ -3,7 +3,7 @@ Include (%syPrompt, %IPM.Common)
 Class %IPM.Repo.Definition Extends (%Persistent, %ZEN.DataModel.Adaptor, %IPM.CLI.Commands) [ Abstract ]
 {
 
-Parameter DEFAULTGLOBAL = "^%IPM.Repo.Definition";
+Parameter DEFAULTGLOBAL = "^IPM.Repo.Definition";
```
daimor, line 6: why?

src/%IPM/ResourceProcessor/Abstract.cls
```diff
@@ -225,35 +225,46 @@ Method OnGetUniqueName(Output pUniqueName)
...
-/// {name} - the name of the resource
-/// These special expressions are case-insensitive.
-Method %Evaluate(pAttrValue) As %String [ Internal ]
+/// <ul>
+/// 	<li>{$root} - the resource's module's root directory </li>
+/// 	<li>{name} - the name of the resource </li>
+/// </ul>
+/// In addition to these, look at expressions supported by %EvaluateSystemExpression
+/// in class <class>%IPM.Utils.Module</class>.
+Method %Evaluate(pAttrValue As %String) As %String [ Internal ]
```
daimor, line 244: needs to be restored what was here before

src/%IPM/Main.cls
```diff
@@ -155,75 +188,106 @@ repo -n LocalFiles -delete
...
+<description>
+Loads a module from the specified directory into the current namespace.
+Dependencies are also loaded automatically, provided that they can be found in
+repositories configured with the 'repo' command.
+</description>
+
+<!-- Examples -->
 <example description="Loads the module described in C:\module\root\path\module.xml">
 load C:\module\root\path\
-load C:\module\root\path\module-0.0.1.tgz
```
daimor, line 182: why, this was deleted?

src/%IPM/Main.cls
```diff
@@ -258,95 +322,189 @@ run C:\Temp\MyCommands.json, where contents of the file are as follows:
...
+</description>
+
+<!-- Examples -->
+<example description="Shows all installed modules in tree format.">
+	list-installed -tree
+</example>
+
+<!-- Modifiers -->
 <modifier name="tree" aliases="t" description="If specified, show dependency tree for installed modules." />
-<modifier name="description" aliases="d" dataAlias="Desc" dataValue="1" description="Additional information is displayed for each module." />
```
daimor, line 316: Why deleted?

src/%IPM/Main.cls
```diff
@@ -362,11 +520,14 @@ default-modifiers -delete
 </example>
 </command>
 
-<command name="version" aliases="ver">
```
daimor, line 365: why aliases deleted?

src/%IPM/Main.cls
Beiming-Zhang: Thank you Dmitry and good catch! I miss delete it in a merge, now it's added back now!

src/%IPM/Main.cls
Beiming-Zhang: Thank you Dmitry and good catch! I miss delete it in a merge, now it's added back now!

src/%IPM/Main.cls
```diff
@@ -991,9 +1155,44 @@ ClassMethod Init(ByRef pCommandInfo) [ Internal ]
...
 	Set tPrimaryClass = ""
-	Set tCompositeSCClass = "%IPM.PackageManager.Developer.Extension.Composite"
-	Set tPackageMgrSCClass = "%IPM.PackageManager.Developer.Extension.PackageManager"
-	Set tSCInterface = "%IPM.PackageManager.Developer.Extension.SourceControl.Interface"
+	Set tCompositeSCClass = "%IPM.ExtensionBase.Composite"
+	Set tServerMenuExtClass = ""
+
+	// set server menu extension class (tServerMenuExtClass)
+	If tQuiet {
+		If $$$HasModifier(pCommandInfo,"menuextension") {
```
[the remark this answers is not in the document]
Beiming-Zhang: Hi Dmitry! This is a case where the user initialize IPM with a full command including server menu extension, which will probably be a very common use case for our studio/vscode developers. Besides, the menu extension class is a very generic class for those who want to set server menu extension, and we've provided such as dummy class as the super class: %IPM.ExtensionBase.ServerMenu.Interface. Therefore, I think is should be part of IPM.

src/%IPM/ResourceProcessor/Abstract.cls
Beiming-Zhang: Hi Dmitry! The %Evaluate method will call %EvaluateSystemExpression, which includes the current functionality of the old version of %Evaluate, so I don't see a very strong reason of reverting back to the old version of %Evaluate(), since all the existing calls to %Evaluate() will have the preserved functionality. Besides, it is an internal method, so I don't see a strong reason to have to revert it. You can review the %EvaluateSystemExpression() method to see if there's any missing features from the old version of %Evaluate(). Thank you!

src/%IPM/DependencyAnalyzer/Problem.cls
Beiming-Zhang: Hi Dmitry, good question! This is because we've re-named the ReferenceClass property to ReferenceItem property and replaced all references to that property. ReferenceItem is a better name since it includes not only classes but also include files, etc. Additionally, we added ReferenceFileType property, that's why the place is changed from 9 to 10

src/%IPM/Lifecycle/Base.cls
Beiming-Zhang: Hi Dmitry! There is an on-going discussion about upper case and lower case! So please hold before we all reach a final agreement. Thank you!

src/%IPM/Repo/Filesystem/Definition.cls
Beiming-Zhang: Hey Dmitry, good catch! Have removed the unnecessary empty line.

src/%IPM/Repo/Definition.cls
Beiming-Zhang: Hi Dmitry, good question! This is to make sure Repo definition's storage global are namespace specific. ^% is instance-wide globals and we don't want the repo definition to exist in other databases, so change it accordingly!

src/%IPM/ExtensionBase/Composite.cls
Beiming-Zhang: Good question! This is the abstract super class for %IPM.Extension.ServerMenu.MenuItems to be used during init, in order to avoid invalid references. Itself is just an abstract class but our studio ui extension, or user defined ui extension, can extend this class. So we are not adding Studio support, but just provide users with super class to add their studio support.

src/%IPM/ExtensionBase/ServerMenu/Interface.cls
Beiming-Zhang: Good question! This is the abstract super class for %IPM.Extension.ServerMenu.MenuItems to be used during init, in order to avoid invalid references. Itself is just an abstract class but our studio ui extension, or user defined ui extension, can extend this class. So we are not adding Studio support, but just provide users with super class to add their studio support.

src/%IPM/ResourceProcessor/Default/Routine.cls
Beiming-Zhang: Now moved to command helper class in %IPM.ResourceProcessor.Default.Document.cls!

src/%IPM/Repo/Definition.cls
daimor: We still don't have an agreement about this. I'm sure that this should be system-wide

src/%IPM/Main.cls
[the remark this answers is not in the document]
Beiming-Zhang: Good point and I agree! However, this is generic server menu extension, not Studio extension, which I believe is generic enough to keep it in IPM.
@isc-kiyer  can weigh in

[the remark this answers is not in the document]
isc-tleavitt: IPM includes an IDE (not Studio-specific) extension to help with Package Manager-specific stuff. This (and related classes about which you had questions) support enabling multiple extensions at the same time without needing to write a separate extension to manage all of them. Ideally this would be in IRIS but since it isn't we can't exactly put it in a dependency of the package manager. (I think we kept the package manager extension, right?)

src/%IPM/Lifecycle/Base.cls
daimor: This is not a part of that discussion at all, here is just normalization for content of module.xml
with this change, it will not support AnyModule.zpm, only AnyModule.ZPM will be allowed.

for instance
https://github.com/intersystems-community/restforms-angular/blob/master/module.xml

src/%IPM/Main.cls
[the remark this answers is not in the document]
isc-tleavitt: Clarifying on what I believe we've agreed on: Specific source control extension implementations (such as ISC's related to Perforce) should be separate packages, not part of IPM itself. The package manager extension should be part of the package manager, as well as the ability to use it seamlessly with arbitrary source control extensions (and to make source control extensions aware of events from a package manager perspective that are relevant to them).

src/%IPM/ExtensionBase/Composite.cls
daimor: I still see no reasons, of having it as part of IPM
it can be just an extra module, which then can be used by extensions developers as a dependency

isc-tleavitt: User story:
As a developer who is currently using a source control extension and *not* the package manager (yet), I want to be able to install the package manager and have it "just work" (including the Package Manager's own extension) without it breaking my development environment.

src/%IPM/DependencyAnalyzer/Problem.cls
Beiming-Zhang: Thanks Dmitry, this is a very solid point and I totally agree! However, this class, %IPM.DependencyAnalyzer.Problem, will only be used in running dependency analyzer, which will always create %New() instance of Problem item. That been said, I don't see a reason that is will corrupt existing data; even that's the case, those old data of Problem class will never be used for twice, since dependencyanalyzer always use new instance of Problem and heve %Save an old data.

src/%IPM/Repo/Filesystem/Definition.cls
Beiming-Zhang: Let's hold on this as well for the uppercase-lowercase package name discussion! Thanks

src/%IPM/ExtensionBase/Composite.cls
daimor: Looks like, there is an issue with IPM then, why it's even possible to break it. Why it should touch user's SourceControl class at all? I still don't get, the reasons of having own Studio's support at all

src/%IPM/DependencyAnalyzer/Problem.cls
daimor: It's not about corrupt existing data, I understand, that it will not happen
it's about consistency, any changes in Storage should be correct, it does not matter if data is not used. 
Just best. practice

src/%IPM/Main.cls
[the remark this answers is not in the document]
Beiming-Zhang: @isc-tleavitt  Thanks for clarifying! Yes, specific-source-control should be in independent package, for example zpm-perforce. But we do provide interface for those (%IPM.ExtensionBase.SourceControl.Interface) which should be fine.

src/%IPM/ExtensionBase/ServerMenu/Interface.cls
Beiming-Zhang: @isc-tleavitt  This class is to 1: provide a dummy super class for user-defined server menu extension class + 2: avoid invalid references to server menu extension in IPM. For example, when initializing IPM, if user do not have a command line prompt for server menu extension, we should query for classes that extend this interface class. 
Based on the query result, if there is exactly one match, we set that as the extension using the composite method. 
If more than one match, and if not in quiet mode, we let the user select which one to set as the extension in the Shell interface.
If no match, do not set.
So that this behaves similar to the current behave of PM extension. Does it make sense?

isc-tleavitt: @Beiming-Zhang why not just query for non-abstract subclasses of %Studio.Extension.Base? Also, mere existence of such a class is not sufficient to indicate that it should be activated as the source control class for the namespace. We just want to overlay ##class(%Studio.SourceControl.Interface).SourceControlClassGet() (if non-empty) with the PM extension on init.

Beiming-Zhang: @isc-tleavitt  We can query on %Studio.Extension.Base, however we will then need to split out other source-control classes such as Perforce.Main which I think would be easier to just separate out the interface for SourceControl extension(%IPM.ExtensionBase.SourceControl.Interface) vs. ServerMenu extension(%IPM.ExtensionBase.ServerMenu.Interface).

src/%IPM/ExtensionBase/Composite.cls
isc-kiyer: Server side source control hooks need to map files to different modules instead of to a single root ^Sources value which is why we have custom source control classes for working with IPM

evshvarov:
Why do you need source-control classes in IPM? Couldn’t this be an ad-hoc
IPM dev package installed as a dev tool?

Ср, 15 июня 2022 г. в 20:00, isc-kiyer ***@***.***>:

src/%IPM/Utils/Module.cls
```diff
@@ -1980,4 +2110,90 @@ ClassMethod GetFileLines(pFileName As %String, Output pOutput) As %Status [ Inte
...
+	Kill pProperties
+	New $Namespace
+	Set $Namespace = "%SYS"
+	Set sc = $ClassMethod("Config.Namespaces", "Get", pNamespace, .pProperties)
+}
+
+ClassMethod %RegExReplace(pString, pRegEx, pValue) As %String [ Internal ]
+{
+	Quit ##class(%Regex.Matcher).%New(pRegEx,pString).ReplaceAll($Replace(pValue,"\","\\"))
+}
```
daimor, line 2196: Move this method from %IPM.Storage.Module
That method is universal for both `${var}` and `{$var}`

And move variables valuation from that class
and the code will be more readable, as well, as less possible errors
```
	Set tAttrValue = ..%RegExReplace(tAttrValue, "namespace", $Namespace)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "ns",        $Namespace)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "mgrdir",    tMgrDir)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "cspdir",    tCSPDir)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "root",      tRoot)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "bindir",    tBinDir)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "libdir",    tLibDir)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "verbose",   tVerbose)
	Set tAttrValue = ..%RegExReplace(tAttrValue, "webroot",   urlRoot)
```

src/%IPM/ResourceProcessor/Abstract.cls
```diff
@@ -225,35 +225,47 @@ Method OnGetUniqueName(Output pUniqueName)
...
-
-  Quit ..ResourceReference.Module.%Evaluate(tAttrValue)
+	Set attrValue = ##class(%IPM.Utils.Module).%EvaluateSystemExpression(tAttrValue)
+	// for backward compatibility, supporting both {$root} and ${root}
+	If $Locate(attrValue,"(?i)\{\$root\}|\$\{root\}") {
+		Set root = ..ResourceReference.Module.Root
+		If (root '= "") {
+			Set root = ##class(%Library.File).NormalizeDirectory(root)
+		}
+		Set attrValue = ##class(%IPM.Utils.Module).%RegExReplace(attrValue,"(?i)\{\$root\}|\$\{root\}",root)
```
daimor, line 262: no need in $locate, and with fix in RegExReplace, it will be just `%RegExReplace(attrValue, "root",root)`

src/%IPM/Storage/ModuleSetting/Default.cls
```diff
@@ -45,7 +45,7 @@ ClassMethod EvaluateAttribute(attribute As %String, ByRef customParams) As %Stri
     For {
         Set name = $Order(customParams(name))
         Quit:(name="")
-        Set regex=##class(%Regex.Matcher).%New("\$\{"_name_"\}", attribute)
+        Set regex=##class(%Regex.Matcher).%New("\$\{"_name_"\}|\{\$"_name_"\}", attribute)
         Set attribute = regex.ReplaceAll(customParams(name))
```
daimor, line 49:
```suggestion
        Set name = $Order(customParams(name),1,value)
        Quit:(name="")
        Set attribute = ##class(%IPM.Utils.Module).%RegExReplace(attribute, name, value)
```

src/%IPM/Utils/Module.cls
Beiming-Zhang: @daimor  Comments resolved. Also added the new properties to handle display in case-sensitive (storage will be in all lower case as before)

Beiming-Zhang:
Closing temporarily

PR #332, opened 2022-07-31
Deployment Enabled

nsolov:
Added multiple remote registry support

`-publish 1`  required in repo command to enable publishing into the registry

`repo -r -n registry -url https://your-registry.com -publish 1`

### DIFF
--- a/src/%ZPM/PackageManager.cls
+++ b/src/%ZPM/PackageManager.cls
@@ -118,7 +118,7 @@ This command is an alias for `module-action module-name verify`
 <modifier name="verbose" aliases="v" dataAlias="Verbose" dataValue="1" description="Produces verbose output from the command." />
 </command>
 
-<command name="publish">
+<command name="publish" dataPrefix="D">
 <description>
 This command is an alias for `module-action module-name publish`
 </description>
--- a/src/%ZPM/PackageManager.cls
+++ b/src/%ZPM/PackageManager.cls
@@ -168,6 +168,8 @@ repo -n AppModules -list-modules
 <modifier name="publish" value="true" valueList="0,1" description="When configuring a remote repository, specifies that publishing of packages to the repository is allowed. When configuring the current namespace with -enable, specifies that publishing is also enabled. "/>
 
 <!-- General properties -->
+<modifier name="enable" value="false" description="For any repository, specifies that it can be used." />
+<modifier name="disable" value="false" description="For any repository, specifies that it can not be used." />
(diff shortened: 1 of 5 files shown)


isc-tleavitt:
As it stands this doesn't allow multiple servers, it allows two, one with deployment enabled and one not.
DeploymentEnabled should be allowed to be 1 or null instead of 1 or 0 - the unique index will allow one server with it set to 1 (as intended), and many with it set to null.

daimor:
> As it stands this doesn't allow multiple registries, it allows two, one with deployment enabled and one not. DeploymentEnabled should be allowed to be 1 or null instead of 1 or 0 - the unique index will allow one server with it set to 1 (as intended), and many with it set to null.

look here, if 0 it makes it null, so, it's ok
```
Method %OnBeforeSave(insert As %Boolean) As %Status [ Private, ServerOnly = 1 ]
{
	// Leave DeploymentEnabled null to make unique index work right
	// (i.e., deployment can only be enabled for one remote server.)
	Set:..DeploymentEnabled=0 ..DeploymentEnabled=""
	Quit $$$OK
}
```

isc-tleavitt:
> > As it stands this doesn't allow multiple registries, it allows two, one with deployment enabled and one not. DeploymentEnabled should be allowed to be 1 or null instead of 1 or 0 - the unique index will allow one server with it set to 1 (as intended), and many with it set to null.
> 
> look here, if 0 it makes it null, so, it's ok
> 
> ```
> Method %OnBeforeSave(insert As %Boolean) As %Status [ Private, ServerOnly = 1 ]
> {
> 	// Leave DeploymentEnabled null to make unique index work right
> 	// (i.e., deployment can only be enabled for one remote server.)
> 	Set:..DeploymentEnabled=0 ..DeploymentEnabled=""
> 	Quit $$$OK
> }
> ```

Ah - didn't realize we already had code for that, great!

daimor:
Well, I've decided to change some logic a bit. I don't want to force users to use publish flag, WWE already have two ways of authorizations on registry, which means that the server with that settings may be used for publishing.  So, I've ordered the list of configured registries in order

- Server with DeploymentEnabled = 1
- Server with Token 
- Server with Username and Password
- Any other servers
So, `publish 1` just raise thee priority

Updated the login on changing settings by repo command, so, already defined properties may be limited, as well as the type of registry
```
zpm "repo -n registry -publish 1"
zpm "repo -n registry -token sometoken"
zpm "repo -n registry -username demo -password demo"
```

daimor (approved):

PR #120, opened 2020-07-14
33 system requirements

nsolov:
specify required IRIS version
and production enabled namespace for a module

PR #338, opened 2022-08-23
deployed modules support

daimor:
Support for deployed modules.
--
Example of module.xml
```xml
<?xml version="1.0" encoding="UTF-8"?>
<Document name="demo.ZPM">
<Module>
 <Name>demo</Name>
  <Version>1.0.0</Version>
  <Description>DEMO.</Description>
  <Keywords>demo</Keywords>
  <SourcesRoot>src</SourcesRoot>
  <Resource Name="Sample.Demo.PKG" Deploy="true"/>
</Module>
</Document>
</Export>
```

Load module, will be loaded in the developer mode my default with the command `load`
```
zpm "load /path/to/module"
```

package as tgz file
```
zpm "package demo"
```
or publish it to registry 
```
zpm "publish demo"
```

Then install tgz file or from that registry, in a new place
```
zpm "install /path/to/demo-1.0.0.tgz"
zpm "install demo"
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,12 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = false
+insert_final_newline = false
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE=containers.intersystems.com/intersystems/irishealth-community:2022.2.0.293.0
+ARG BASE=containers.intersystems.com/intersystems/irishealth-community:2022.2.0.311.0
 FROM ${BASE}
 
 ARG REGISTRY=https://pm.community.intersystems.com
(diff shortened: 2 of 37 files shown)


daimor:
Yes, sure, this is very important. And I would think of it as a next step, and that should be done together with support form registry.

isc-tleavitt:
https://github.com/intersystems-community/zpm/pull/120 may be helpful for this

daimor:
> #120 may be helpful for this

Yeah, this is a good place. And multi-version support on registry. So, we can directly request build for particular version of IRIS

isc-tleavitt:
Information on the current platform version needs to be pushed into search criteria sent to the registry (of course).

To be clear - I'm a HUGE fan of this feature and have wanted to see it for a while. :)

isc-tleavitt:
For the initial scope, I think module.xml in the .tgz only (not in source) should have SystemRequirements generated based on the platform version used for packaging.

daimor:
> For the initial scope, I think module.xml in the .tgz only (not in source) should have SystemRequirements generated based on the platform version used for packaging.

Agree

daimor:
So, added support on Registry too. And It will check the version of IRIS during installation
At the moment, check only major and minor e.g. 2022.2, in hope it should be enough in most cases. Not sure that  we have to be too strict

isc-tleavitt:
I think checking major + minor is enough.

PR #348, opened 2022-10-05
Universal display for lists of packages

daimor:
Attempt to merge all outputs of packages in one universal way, so, it always will look the same, no matter what the list is shown

Affected commands
- list - list of installed packages (except tree view)
- search - search for packages in registries
- zn - switch to namespace
- ver - display versions of `zpm` and `zpm-registry` (if installed)


daimor:
[zpm-0.4.1-dev.348.1.zip](https://github.com/intersystems-community/zpm/files/9722728/zpm-0.4.1-dev.348.1.zip)

PR #380, opened 2023-04-07
Support for Atelier/VSCode format files

daimor:
Added support for LUT in both formats, it checks if XML is not in XML `/Export/Document` format, then loads it through `%RoutineMgr`

integration test does not work on 2023.1 at the moment due to changes with `##class(%EnsembleMgr).IsHealthShareInstalled()`, no idea how to fix it yet
@timleavitt


isc-tleavitt (changes requested):
The most pressing issue is supporting non-XML %Studio.AbstractDocument types. One simple solution would be switching the logic around so that for user types we use $System.OBJ.Load if we can find /Export/Document and using %RoutineMgr if anything goes wrong (e.g., document isn't XML, document doesn't contain /Export/Document). For non user types we'd keep just using $System.OBJ.Load.

src/%ZPM/PackageManager/Developer/Processor/Default/Document.cls
```diff
@@ -399,8 +399,31 @@ Method OnLoad(pFullResourcePath As %String, pVerbose As %Boolean, pCompile As %B
       Set pLoadedList = ""
       Set tSC = $$Import^%apiRTN(pFullResourcePath,"replace,compile="_pCompile,,,.tResult,$Select(pVerbose:"d",1:"-d"))
     } Else {
-      Set tSC = $System.OBJ.Load(pFullResourcePath,tFlags,,.pLoadedList)
-      $$$ThrowOnError(tSC)
+      Set tProcessed = 0
+      if ##class(%RoutineMgr).UserType(..ResourceReference.Name, .docclass, .doctype) {
```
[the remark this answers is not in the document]
daimor: I personally don't like that `..Format` separation, and trying to make it as much simple as possible and universal. So, it'll keep loading whatever is acceptable despite the Format part, which is mostly confusing.
Packaging is supposed to be only for the installation by IPM itself. The case is only to support files in formats stored in SourceControl by developers.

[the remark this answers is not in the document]
isc-tleavitt: @daimor that's fair. The intent of Format is primarily to support taking existing source controlled code and define a package around it without losing history / needing to move things around. I'm fine with load being clever/permissive, as long as it actually permits all the cases it should. I still think a change is needed to your implementation for user types to use $System.OBJ.Load if we find /Export/Document and %RoutineMgr if anything goes wrong with that.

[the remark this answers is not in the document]
daimor: yep, I'll do it

daimor:
@isc-tleavitt  do you have some other files that can be used in the test? 

https://github.com/intersystems/ipm/blob/87c5bf7e7d89891375c1fa0a9514cd574e0242bb/tests/integration_tests/Test/PM/Integration/_data/resource-test/module.xml

src/%ZPM/PackageManager/Developer/Processor/Default/Document.cls
```diff
@@ -399,8 +400,40 @@ Method OnLoad(pFullResourcePath As %String, pVerbose As %Boolean, pCompile As %B
       Set pLoadedList = ""
       Set tSC = $$Import^%apiRTN(pFullResourcePath,"replace,compile="_pCompile,,,.tResult,$Select(pVerbose:"d",1:"-d"))
     } Else {
-      Set tSC = $System.OBJ.Load(pFullResourcePath,tFlags,,.pLoadedList)
-      $$$ThrowOnError(tSC)
+      Set tProcessed = 0
+      set fileExt = $$$UPPER($Piece(pFullResourcePath, ".", *))
+      if '$LISTFIND($LB("CLS", "MAC"), fileExt), ##class(%RoutineMgr).UserType(..ResourceReference.Name, .docclass, .doctype) {
```
isc-tleavitt, line 405: Why do we need to check fileExt? UserType should return 0 for both of those.

isc-tleavitt (commented):
One question here - maybe needs a fix, maybe not.

isc-tleavitt:
@daimor I don't have samples offhand without using document type implementations I can't share. I'll try to kick the tires on this a bit today but from code inspection I think it should work.

src/%ZPM/PackageManager/Developer/Processor/Default/Document.cls
daimor: ResourceReference.Name and pFullResourcePath are different
we have a case with `bpl`, and it can be as `cls` file

evshvarov: What about DFI?

daimor: This one in the test at the moment
https://github.com/intersystems/ipm/blob/87c5bf7e7d/tests/integration_tests/Test/PM/Integration/_data/resource-test/dfi/Sample_Operational_Reports/Auditing_Overview.dashboard.xml

daimor:
Updated test, checked with: hl7, x12, lut, dfi
including, the way it's exported by VSCode, with the original extension different from XML

daimor:
Got some issues, making it working on Linux, with case-sensitive filesystem. Luckily CI is on real Linux, so, it caught that and I managed to solve it.
For that case-sensitive case added new File Exists method, which checks if the file exists no matter what case it case, is in a few steps, to make it faster in some scenarios.

evshvarov:
@isc-tleavitt  @isc-rkuszews  please review? This is a very demanded change. Thank you!

isc-tleavitt:
@evshvarov it's on my list to review today.

src/%ZPM/PackageManager/Developer/File.cls
```diff
@@ -182,4 +182,35 @@ ClassMethod FindFiles(pPath As %String, pWildcards As %String = "", Output pList
 	Return pList
 }
 
+/// Check if file exists but case insensetive
```
isc-tleavitt, line 185: Inconsequential typo: insensitive

isc-tleavitt (approved):
This looks good to me.

PR #420, opened 2023-06-22
Set the ^UnitTestRoot node when building the container

gjsjohnmurray:
This is a follow-up to #415. It switches to using an `iris.script` file for the sequence of ObjectScript commands that need to be run when the container is being built, and adds the one that completes integration of InterSystems Testing Manager.


gjsjohnmurray:
Pinging @isc-rkuszews as a potential reviewer.

gjsjohnmurray:
Pinging @isc-rkuszews @daimor about getting this merged.

daimor:
I'm not a maintainer anymore

PR #437, opened 2024-01-25
Fold all updates since May from master and internal repo into v1

isc-tleavitt:
This also includes some in-progress changes to the packages workflow (will be handled separately, this at least moves the ball forward there although it isn't done yet).

This includes through internal revision 6487010.


isc-bzhang (approved):

isc-rkuszews (approved):

isc-kiyer (approved):

isc-tleavitt:
@isc-shuliu in case you get to the CI failures before I do, there are three classes of issues:
* Failures early because of our processing of tags - we're turning e.g. latest-em and latest-cd into "latest" which doesn't exist. Need to be smarter here.
* Failures early because of community edition licenses expiring - we should detect and fail early (without an error) in this case. If there's some kind of "skipped" option for the job that is neither success nor failure that would be ideal, if possible.
* We only want the highest value of z among builds (year.x.y.z)
* Failures later in the process because of the ZPM->IPM rename (this is a bigger can of worms I don't expect you to open just yet ;))

isc-tleavitt:
@isc-shuliu here's an updated checklist after the work I did earlier this week:

- [ ] Be smarter about expired community edition licenses; could have a min version flag or a file with a list of tags to ignore?
- [ ] Issues in the clean install step (this is likely to be the bigger can of worms, but we can at least investigate + fix the missing container part)
- [ ] Fix unit tests (currently have a failure due to a code change) - we might need to talk about the code change
- [ ] Fix "verify" phase (currently failing due to ZPM/IPM differences)
- [ ] Ensure "verify" passes
- [ ] Make test/verify failures fatal again
- [ ] Really fix clean install + upgrade (this one is on me and is likely enormous)

Issue #451, opened 2024-04-12
CI fixes for v1 branch


isc-tleavitt:
Splitting this out from https://github.com/intersystems/ipm/pull/437 as a separate issue:

- [x] Be smarter about expired community edition licenses; could have a min version flag or a file with a list of tags to ignore? (need this in master and v1)
- [x] Fix failing compilation on IRIS for Health prior to 2024.1
- [x] Issues in the clean install step (this is likely to be the bigger can of worms, but we can at least investigate + fix the missing container part)
- [x] Fix unit tests (currently have a failure due to a code change) - we might need to talk about the code change
- [x] Fix "verify" phase (currently failing due to ZPM/IPM differences; see also #428)
- [x] Ensure "verify" passes
- [x] Make test/verify failures fatal again
- [x] Really fix clean install + upgrade (this will require other work tracked under other issues, but #449 was a good first step)

isc-tleavitt:
#457 handles all but the clean install + upgrade.

isc-tleavitt:
#466 incorporates a change to clean install + upgrade (assuming our upcoming v0.7.1 release as a basis)

PR #447, opened 2024-03-19
fix: IPM works with delimited identifiers disabled

isc-tleavitt:
Also picks up a fix for a stray failing unit test - ${verbose} / {$verbose} weren't working in expression evaluation

Fixes #440

PR #452, opened 2024-04-12
Run on fewer container versions

isc-tleavitt:
This will make CI faster and still provide sufficient coverage of relevant versions. We now only run on the major.minor container tags (latest maint release) and latest-preview. This skips over a bunch we were trying before that had expired community edition licenses.

This change partially addresses #451 but doesn't provide the entire solution.

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -26,7 +26,7 @@ jobs:
         run: |
           images=""
           for n in $name; do
-            tags=$(curl -su ":" https://containers.intersystems.com/v2/intersystems/${n}/tags/list | jq -r '.tags[]' | awk '!/(-linux)|([1-4]-preview)|(-em)|(-cd)/' | awk '!/\.[1-4]\./' | uniq)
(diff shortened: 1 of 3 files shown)


isc-tleavitt:
There's a fix for the failing unit test (not reported as a failing check still, that's another part of #451) in https://github.com/intersystems/ipm/pull/447

.github/workflows/images.yml
```diff
@@ -26,8 +26,16 @@ jobs:
         run: |
           images=""
           for n in $name; do
-            tags=$(curl -su ":" https://containers.intersystems.com/v2/intersystems/${n}/tags/list | jq -r '.tags[]' | awk '!/(-linux)|([1-4]-preview)/' | awk '!/\.[1-4]\.0\./' | uniq)
-            for tag in $tags; do images+='"'${n}:${tag}'",'; done
+            tags=$(curl -su ":" https://containers.intersystems.com/v2/intersystems/${n}/tags/list | jq -r '.tags[]' | awk '!/(-linux)|([1-4]-preview)|(-em)|(-cd)/' | awk '!/\.[1-4]\./' | uniq)
```
isc-shuliu, line 29: Maybe add a `| sort` before `| uniq` in case future API doesn't return sorted tags?

isc-tleavitt: Good idea!

isc-shuliu (approved):

PR #458, opened 2024-04-15
Remove duplicate flags in iris.script

isc-shuliu:


isc-kiyer (approved):

CHANGELOG.md
```diff
@@ -45,7 +45,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -
 
 ### Removed
-- 
+- Removed duplicate flags in iris.script for building Docker containers
```
isc-tleavitt, line 48: This doesn't belong in the "removed" section - the intent of "removed" is to capture features that used to exist and no longer do (see https://keepachangelog.com/en/1.0.0/). This either falls under "Fixed" or perhaps, because it's purely internal to running in docker for development, doesn't need a changelog at all. If you put it under fixed, it's worth also mentioning the issue that the change fixes (although there isn't one in this case). I'd honestly vote no changelog for this.

isc-tleavitt (changes requested):

isc-tleavitt (approved):

PR #461, opened 2024-04-17
fix: Dockerfile uses 2024.1

isc-tleavitt:
Previous version has expired community edition license.


isc-tleavitt:
@isc-shuliu FYI, if you're not using Docker for your local dev/testing you probably should be, it makes running integration tests much easier.

isc-kiyer (approved):

PR #466, opened 2024-04-25
v0.7.x to v0.9.x upgrade path and storage migration

### LINKED ISSUE #454: Automatically migrate %ZPM -> %IPM storage on install
When upgrading from v0.7.x to 0.9.0, all data should be migrated automatically so that metadata regarding currently-installed packages remains valid.

isc-tleavitt:
Fixes #454

**Upgrade Path**
zpm 0.7.1 (once released) will be able to install zpm 0.9.0 (once released).

**Storage Migration**
Moves `^ZPM.Dev.ModuleD` to `^IPM.Storage.ModuleD`

In `^IPM.Storage.ModuleD(<ID>,"Defaults",n)` maps classname to equivalent - `%ZPM.PackageManager.Developer.ModuleSetting` -> `%IPM.Storage.ModuleSetting`

Migrates `%ZPM.PackageManager.Developer.Document.ModuleStream` to `%IPM.StudioDocument.ModuleStream`

For custom lifecycle classes, we're less picky and allow the legacy naming too. (I'm probably the only one using one of these anyway, in https://github.com/intersystems/isc-ipm-js )

Also incorporates the final CI fixes for #451.


isc-tleavitt:
Leaving this as a draft until CI is all passing. (I don't have time to run it all locally in [act](https://github.com/nektos/act) today, so let's see what happens!)

isc-tleavitt:
Well, that was quick. Also trying to fix the last lingering task in #451 here and that's where the CI issue is currently.

isc-tleavitt:
This looks like it's working. Note, it builds on #457 so review that one first.

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,108 @@
...
+
+            Set name = $ListGet(moduleData,2)
+            If ##class(%IPM.Storage.Module).NameExists(name) {
+                Write:verbose !,"Package already found in new storage: ",name
+                Continue
+            }
+
+            Write:verbose !,"Migrating storage for ",name," ... "
+
+            // Bring everything over - storage definitions are fortunately compatible.
```
isc-kiyer, line 42: What if we introduce an incompatibility later and someone hasn't yet migrated? I think it would be safer to use object copy over each property. More verbose but safer in the long run

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,108 @@
...
+
+            Set newId = $Increment(^IPM.Storage.ModuleD)
+            Merge ^IPM.Storage.ModuleD(newId) = moduleData
+
+            // Make sure loading the object works
+            Set newObj = ##class(%IPM.Storage.Module).%OpenId(newId,,.sc)
+            $$$ThrowOnError(sc)
+
+            // Save object to validate
+            $$$ThrowOnError(newObj.%Save())
```
isc-kiyer, line 63: If validation fails, I think it's best to revert the transaction for this one module and continue, attempting to migrate the remaining at least. For the modules that failed, perhaps accumulate the actual errors so users can report them as GitHub issues but indicate in the message that users could manually uninstall and reinstall the module if needed.

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        $$$ThrowOnError(##class(%IPM.Storage.Module).%BuildIndices(,1,1))
+
+        // Migrate %ZPM.PackageManager.Developer.Document.ModuleStream to %IPM.StudioDocument.ModuleStream
+        // ^ZPM.Dev.ModuleDocD -> ^IPM.StudioDoc.ModuleStreamD
+        // Uses objects for the actual data migration to more cleanly handle the stream.
+        Set oldId = ""
+        For {
+            Set oldId = $order(^ZPM.Dev.ModuleDocD(oldId),1,moduleDocData)
```
isc-kiyer, line 75: Rather than separate for loops, as IDs are identified in the module storage, also migrate its corresponding studio document?

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,108 @@
...
+            Merge ^IPM.Storage.ModuleD(newId) = moduleData
+
+            // Make sure loading the object works
+            Set newObj = ##class(%IPM.Storage.Module).%OpenId(newId,,.sc)
+            $$$ThrowOnError(sc)
+
+            // Save object to validate
+            $$$ThrowOnError(newObj.%Save())
+
+            Write:verbose "done."
```
isc-kiyer, line 65: Should the old globals be killed as migration succeeds?

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,108 @@
...
+            Set newObj = ##class(%IPM.StudioDocument.ModuleStream).%New()
+            Set newObj.Name = oldObj.Name
+            Set newObj.Hash = oldObj.Hash
+            Set newObj.LastModifiedTimestamp = oldObj.LastModifiedTimestamp
+            $$$ThrowOnError(newObj.Contents.CopyFrom(oldObj.Contents))
+            $$$ThrowOnError(newObj.%Save())
+
+            Write:verbose !,"Migrated document stream data for "_newObj.Name
+        }
+
```
isc-kiyer, line 94: Do we also need to migrate repos?

tests/sandbox/Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG BASE=containers.intersystems.com/intersystems/iris-community:2024.1
```
isc-kiyer, line 1: Nit: General comment - why not store the IRIS version in one place in a file somewhere and extract it for use in various other test files? Otherwise need to find/replace every time

isc-kiyer (changes requested):
@timleavitt a few comments. a few questions. I don't entirely understand the GitHub workflow syntax so haven't reviewed that. Would be good to get a second set of eyes on that bit

src/cls/IPM/Utils/Migration.cls
isc-tleavitt: Good call

tests/sandbox/Dockerfile
isc-tleavitt: I'm pretty new to docker - will look to see if there's an easy way to do that. (If you know of one off the top of your head, please tell me!)

isc-tleavitt:
> @timleavitt a few comments. a few questions. I don't entirely understand the GitHub workflow syntax so haven't reviewed that. Would be good to get a second set of eyes on that bit

@isc-shuliu I'd appreciate if you could review the GitHub workflow parts of this, since you've worked with it more than anyone else reviewing. (And probably more than I have!)

tests/sandbox/Dockerfile
isc-tleavitt: @isc-shuliu do you have any thoughts on this from a Docker perspective?

src/cls/IPM/Utils/Migration.cls
isc-tleavitt: Example of something we might do that would defeat objects-based copying: make a property private. (I want to do this for a few properties on repos.) On the other hand storage definition should retain compatibility.

tests/sandbox/Dockerfile
isc-tleavitt: On further thought: it's reasonable that we'd want to use different versions in different places, especially for the new "sandbox" container. (I think this is something it's likely a dev would briefly change locally, say to use irishealth-community or a different version instead.)

We also want to be intentional about bumping versions (generally). So the Dockerfile is a reasonable place for this.

isc-tleavitt:
This is ready for review (again).

isc-tleavitt:
Bumping this back up - needs re-review.

@isc-kiyer @isc-bzhang @isc-eneil @isc-shuliu

.github/workflows/main.yml
```diff
@@ -165,12 +165,41 @@ jobs:
...
+          curl http://localhost:52773/registry/packages/zpm/ | jq
+          ASSET_NAME='zpm-0.7.1-beta.4.xml'
+          ASSET_URL=`wget --header "Authorization: token ${GITHUB_TOKEN}" -qO- https://api.github.com/repos/intersystems/ipm/releases | jq -r ".[].assets[] | select(.name == \"${ASSET_NAME}\") | .browser_download_url"`
+          wget $ASSET_URL -O /tmp/zpm.xml
+          docker run --network zpm --rm -d ${{ steps.image.outputs.name }} ${{ steps.image.outputs.flags }}
+          CONTAINER=`docker ps -lq`
+          docker cp /tmp/zpm.xml $CONTAINER:/home/irisowner/zpm.xml
+          docker cp tests/migration/v0.7-to-v0.9/. $CONTAINER:/tmp/test-package/
+          sleep 5; docker exec $CONTAINER /usr/irissys/dev/Cloud/ICM/waitISC.sh
+          /bin/echo -e '' \
```
isc-shuliu, line 183: Personally I find using `heredoc` more readable and less error-prone. It doesn't require escaping line breaks (or potentially single quotes)

.github/workflows/main.yml
```diff
@@ -165,12 +165,41 @@ jobs:
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl http://localhost:52773/registry/packages/-/all | jq
+          curl http://localhost:52773/registry/packages/zpm/ | jq
+          ASSET_NAME='zpm-0.7.1-beta.4.xml'
+          ASSET_URL=`wget --header "Authorization: token ${GITHUB_TOKEN}" -qO- https://api.github.com/repos/intersystems/ipm/releases | jq -r ".[].assets[] | select(.name == \"${ASSET_NAME}\") | .browser_download_url"`
+          wget $ASSET_URL -O /tmp/zpm.xml
+          docker run --network zpm --rm -d ${{ steps.image.outputs.name }} ${{ steps.image.outputs.flags }}
+          CONTAINER=`docker ps -lq`
```
isc-shuliu, line 179: Alternatively, can also use `CONTAINER=$(docker run xxx)`. This way, even if the order of lines is changed, or a new `docker run` command is inserted between the above `docker run` and `docker ps -lq`, we still get a hash of the correct container. If you want to log the container ID for debugging purposes, you can run an `echo $CONTAINER` afterwards.

.github/workflows/main.yml
```diff
@@ -165,12 +165,41 @@ jobs:
       #       'halt\n' \
       #     | docker exec -i $CONTAINER iris session iris -UUSER
       #     docker stop $CONTAINER
+      - name: Clean install v0.7.x and test migration to latest.
+        timeout-minutes: 15
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          curl http://localhost:52773/registry/packages/-/all | jq
+          curl http://localhost:52773/registry/packages/zpm/ | jq
```
isc-shuliu, line 174: Are we assuming localhost:52773 is running the correct kit here?

.github/workflows/main.yml
isc-shuliu: It would look like this
```bash
docker exec -i $CONTAINER iris session iris -UUSER << EOF 
  set sc = ##class(%SYSTEM.OBJ).Load("/home/irisowner/zpm.xml", "ck")
  zpm "list":1
  zpm "repo -r -name registry -url ""http://registry:52773/registry/"" -username admin -password SYS":1
  zpm "repo -list"
  zpm "search":1
  zpm "install sslclient":1
  zpm "install zpm -v":1
  zpm "load -dev /tmp/test-package/":1
  zpm "test ipm-migration-v0.7-to-v0.9 -only -verbose":1
  halt
EOF
```

.github/workflows/main.yml
isc-tleavitt: Huh... I got this from the above section, but it makes sense neither place. Good catch.

isc-tleavitt: Actually - on further review, there's a binding of port 52773 on the runner to the same port on the registry container, so this is OK. (I guess it would fail early if there's something wrong with the registry?)

module.xml
```diff
@@ -2,8 +2,7 @@
 <Export generator="Cache" version="25">
 <Document name="ZPM.ZPM"><Module>
   <Name>ZPM</Name>
-  <GlobalScope>true</GlobalScope>
-  <Version>0.9.0+snapshot</Version>
+  <Version>0.9.0-SNAPSHOT</Version>
```
isc-shuliu, line 5: Questions: 
What does `snapshot` typically mean in package definitions. Does it mean "development in progress"?
What is the reason for changing the casing and switching from `+build` to `-prerelease`?

isc-tleavitt: Yes, it means "this is not a released version" - although our usage and rules around this are a little different from standard semantic versioning.

The standard in Maven (for Java, one of IPM's inspirations) is to use -SNAPSHOT for the current in-development version. See e.g. https://stackoverflow.com/questions/5901378/what-exactly-is-a-maven-snapshot-and-why-do-we-need-it

You don't want to just call your current in-development version "0.9.0" because if an artifact for it ends up somewhere, it could be mistaken for the real 0.9.0. In the initial package manager implementation I chose to break with the semver standard and distinguish unreleased software via build metadata (+snapshot) rather than a prerelease (-SNAPSHOT). I thought there was some good reason for this at the time, but I don't remember it and now think it's silly to use +snapshot. The automation in GitHub actions for IPM assumes -SNAPSHOT so I've changed it back.

cc @isc-kiyer @isc-bzhang

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,190 @@
...
+    Write:verbose !,"Migrating repo "_name_"... "
+
+    Set $List(data,1) = "~%IPM.Repo.Remote.Definition~"
+    Set data("RepoDefinition") = data
+    Merge data("RemoteRepoDefinition") = data("RemoteServerDefinition")
+    ZKill data
+    ZKill data("RemoteServerDefinition")
+
+    Set newId = $Increment(^IPM.Repo.DefinitionD)
```
isc-kiyer, line 177: Similarly perhaps worth adding locking for repo definitions

isc-tleavitt:
Need to update integration tests to interoperability-enable the VERIFY namespace.

shuheng-liu (approved):

shuheng-liu:
Looks all good to me!

isc-tleavitt:
@shuheng-liu FYI, you switched profiles ;)

src/cls/IPM/Utils/Migration.cls
```diff
@@ -0,0 +1,209 @@
...
+
+    Set sc = $$$OK
+    Set successPackageList = ""
+    Set failedPackageList = ""
+
+    // Intentionally runs for current namespace only
+    Try {
+        // ^ZPM.Dev.ModuleD -> ^IPM.Storage.ModuleD
+        For lockName = $Name(^ZPM.Dev.ModuleD),$Name(^IPM.Storage.ModuleD) {
+            Set locks(lockName) = $System.AutoLock.Lock(lockName,,2)
```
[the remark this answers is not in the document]
isc-tleavitt: @isc-kiyer I think I might have written LockManager (at least I've written things like it before) - it may be cleaner to manage locks in a single object rather than a bunch of objects, so I wouldn't necessarily throw it out just yet. Although perhaps it could be updated to use AutoLock under the hood.

PR #484, opened 2024-05-21
fix: exact prerelease match (HSIEO-10520)

isc-tleavitt:
Merged from internal repo.

Note: this gets GitHub up-to-date with internal work, though I believe @isc-eneil also has work in progress that will need to merge. Other than that in-progress work, this was the last straggler.


isc-kiyer (approved):

isc-tleavitt:
CI fail isn't concerning on this. It was in a funky state so I cancelled, seems like most successfully completed anyway.

isc-shuliu (approved):

PR #485, opened 2024-05-22
Support enabling IPM instance-wide

### LINKED ISSUE #470: Review and improve installation modes
With v0.9, IPM gains the ability to be installed namespace-specific, but is currently restricted to explicitly *not* run in %SYS.

Really the main goal had been to get it out of IRISSYS, not to prevent legitimate use cases (meaning, currently: not modifying IRISLIB) from running in the %SYS namespace.

Need to review %IPM.Main:EnableNS to better handle enabling instance-wide. This might be able to use %ALL (which would be valid), need to test; otherwise it'll be a ton of namespace-specific mappings.

Upon migration from v0.7, instance-wide should be the default.

isc-tleavitt:
This is done via new options in the existing zpm "enable" command that add package mappings.

We looked into using %ALL but ultimately that doesn't work, because namespace-specific mappings aren't allowed to override %ALL mappings (!?).

Fixes #470

### DIFF
--- a/preload/cls/IPM/Installer.cls
+++ b/preload/cls/IPM/Installer.cls
@@ -44,10 +44,10 @@ XData PM [ XMLNamespace = INSTALLER ]
 XData MapOnly [ XMLNamespace = INSTALLER ]
 {
 <Manifest>
-  <Namespace Name="${CURNS}" Code="${CURNSRoutineDB}" Data="${CURNSGlobalDB}" Create="overwrite" Ensemble="0">
+  <Namespace Name="${CURNS}" Code="${CURNSRoutineDB}" Data="${CURNSGlobalDB}" Create="no" Ensemble="0">
     <Configuration>
         <ClassMapping Package="%IPM" From="${CURNSRoutineDB}" />
-        <RoutineMapping Routines="%IPM.*" Type="ALL" From="${CURNSRoutineDB}"/> 
+        <RoutineMapping Routines="%IPM.*" Type="ALL" From="${CURNSRoutineDB}" /> 
     </Configuration>
   </Namespace>
 </Manifest>
@@ -80,6 +80,7 @@ ClassMethod Map(ByRef pVars, pLogLevel As %Integer, pInstaller As %Installer.Ins
   Do %code.WriteLine(" Set sc = ##class(Config.Namespaces).Get(prevNS, .properties)")
   Do %code.WriteLine(" Set pVars(""CURNSRoutineDB"")=$Get(properties(""Routines""))")
   Do %code.WriteLine(" Set pVars(""CURNSGlobalDB"")=$Get(properties(""Globals""))")
+  Do %code.WriteLine(" If ##class(Config.MapPackages).Exists(prevNS,""%IPM"") { Quit $$$OK }")
   Quit ##class(%Installer.Manifest).%Generate(%compiledclass, %code, "MapOnly")
 }
 
--- a/preload/cls/IPM/Installer.cls
+++ b/preload/cls/IPM/Installer.cls
@@ -35,7 +35,6 @@ XData PM [ XMLNamespace = INSTALLER ]
     <Arg Value="${MODDIR}"/>
   </Invoke>
   <Invoke Class="IPM.Installer" Method="ZPMCompile" CheckStatus="true" />
-  <Invoke Class="IPM.Installer" Method="MapIfLegacy" CheckStatus="true" />
 </Namespace>
 
 </Manifest>
--- a/preload/cls/IPM/Installer.cls
+++ b/preload/cls/IPM/Installer.cls
@@ -35,6 +35,7 @@ XData PM [ XMLNamespace = INSTALLER ]
     <Arg Value="${MODDIR}"/>
   </Invoke>
   <Invoke Class="IPM.Installer" Method="ZPMCompile" CheckStatus="true" />
+  <Invoke Class="IPM.Installer" Method="MapIfLegacy" CheckStatus="true" />
 </Namespace>
 
 </Manifest>
@@ -43,10 +44,10 @@ XData PM [ XMLNamespace = INSTALLER ]
 XData MapOnly [ XMLNamespace = INSTALLER ]
 {
 <Manifest>
-  <Namespace Name="${CURNS}" Code="${CURNSRoutineDB}" Data="${CURNSGlobalDB}" Create="no" Ensemble="0">
+  <Namespace Name="${CURNS}" Code="${CURNSRoutineDB}" Data="${CURNSGlobalDB}" Create="overwrite" Ensemble="0">
     <Configuration>
         <ClassMapping Package="%IPM" From="${CURNSRoutineDB}" />
-        <RoutineMapping Routines="%IPM.*" Type="ALL" From="${CURNSRoutineDB}" /> 
+        <RoutineMapping Routines="%IPM.*" Type="ALL" From="${CURNSRoutineDB}"/> 
     </Configuration>
   </Namespace>
 </Manifest>
@@ -79,7 +80,6 @@ ClassMethod Map(ByRef pVars, pLogLevel As %Integer, pInstaller As %Installer.Ins
   Do %code.WriteLine(" Set sc = ##class(Config.Namespaces).Get(prevNS, .properties)")
   Do %code.WriteLine(" Set pVars(""CURNSRoutineDB"")=$Get(properties(""Routines""))")
   Do %code.WriteLine(" Set pVars(""CURNSGlobalDB"")=$Get(properties(""Globals""))")
-  Do %code.WriteLine(" If ##class(Config.MapPackages).Exists(prevNS,""%IPM"") { Quit $$$OK }")
   Quit ##class(%Installer.Manifest).%Generate(%compiledclass, %code, "MapOnly")
 }
 
@@ -113,6 +113,14 @@ ClassMethod ZPMLoad(pDirectoryName)
 ClassMethod ZPMCompile()
 {
   Quit ##class(%IPM.Main).Shell("ZPM compile")
+}
+
+ClassMethod MapIfLegacy()
+{
+  If ##class(%IPM.Utils.Migration).HasLegacyZPMPackage() {
+    Quit ##class(%IPM.Main).Shell("enable -globally -map")
+  }
+  Quit $$$OK
 }
 
 ClassMethod EndCompile(qstruct) As %Status
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -2697,6 +2697,8 @@ ZPM(pArgs...)
  	Do ##class(%IPM.Main).Shell(pArgs...) Quit
  } Else {
 	// TODO: Needs to support enabling IPM from here, also need to decide what level of customization to provide
+	Set tQuitOnError = $Get(pArgs(2))
+	Set tHaltOnComplete = $Get(pArgs(3))
 	Write !, "IPM is not enabled in this namespace."
 	&sql(SELECT %DLIST(Nsp) INTO :namespaces FROM %SYS.Namespace_List())
 	If SQLCODE '= 0 {
@@ -2704,18 +2706,32 @@ ZPM(pArgs...)
 	} Else {
 		New $Namespace
 		Set ptr = 0
+		Set found = 0	
 		While $ListNext(namespaces, ptr, ns) {
 			Set $Namespace = $Zstrip(ns, "<>WC")
 			If $System.CLS.IsMthd("%IPM.Main", "Shell") {
 				Write !, "Change namepace to one of the following to run the ""zpm"" command"
 				Do ##class(%IPM.Main).Shell("version")
-				Return ""
+				Set found = 1
+				Quit
 			}
 		}
 		// Shouldn't happen since %ZLANGC00.mac and %ZLANGF00.mac are present
-		Write "No namespace found with IPM enabled."
+		If 'found {
+			Write !, "No namespace found with IPM enabled."
+		}
 	}
-	Quit ""
+
+	If tQuitOnError {
+		Write ! // Add a newline before quitting, otherwise the shell prompt will be on the same line as the error message
+		Do $System.Process.Terminate($Job, 1)
+	}
+	If tHaltOnComplete {
+		Write ! // Add a newline before halting
+		Halt
+	}
+	// The $$$ERROR macro is not available for use in language extension
+	Return $$Error^%apiOBJ(5001,"IPM is not enabled in this namespace.")
  }
 }
 
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -611,7 +611,7 @@ generate /my/path -export 00000,PacketName2,IgnorePacket2^00000,PacketName3,Igno
 		<!-- Modifiers -->
 		<modifier name="version" aliases="v" value="true" description="A special version of IPM can be provided. If not specified, the latest version from the registry will be installed (hence is not required if quiet flag is set)."/>
 		<modifier name="namespaces" aliases="ns" value="true" description="Comma-separated namespaces in which IPM needs to be enabled."/>
-		<modifier name="globally" aliases="g" description="Will install IPM in all non-%SYS explicit namespaces that currently do not have IPM installed. By default, this modifier is not set and will not install globally."/>
+		<modifier name="globally" aliases="g" description="Will install IPM in all explicit namespaces that currently do not have IPM installed. Unless using it together with the -map option, will not include %SYS. By default, this modifier is not set and will not install globally."/>
 		<modifier name="local-only" description="If specified, only local artifacts will be used for installation. By default, this modifier is not set and will not limit to local artifacts." />
 		<modifier name="allow-upgrade" description="If specified, will also check for IPM version in specified namespaces and upgrade if version is lower than the target version. By deafult, this modifier is not ste and will not allow upgrade." />
 		<modifier name="map" aliases="m" description="If specified, will map IPM code from the current namespace-default code database rather than installing a separate copy." />
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -2648,8 +2648,25 @@ ZPM(pArgs...)
  If $System.CLS.IsMthd("%IPM.Main", "Shell") { 
  	Do ##class(%IPM.Main).Shell(pArgs...) Quit
  } Else {
-	// TODO: Decide what to do here; definitely needs to support enabling IPM from here as well,
-	// but need to decide what level of customization to provide
+	// TODO: Needs to support enabling IPM from here, also need to decide what level of customization to provide
+	Write !, "IPM is not enabled in this namespace."
+	&sql(SELECT %DLIST(Nsp) INTO :namespaces FROM %SYS.Namespace_List())
+	If SQLCODE '= 0 {
+		Write !, "Error getting namespace name. SQLCODE =", SQLCODE
+	} Else {
+		New $Namespace
+		Set ptr = 0
+		While $ListNext(namespaces, ptr, ns) {
+			Set $Namespace = $Zstrip(ns, "<>WC")
+			If $System.CLS.IsMthd("%IPM.Main", "Shell") {
+				Write !, "Change namepace to one of the following to run the ""zpm"" command"
+				Do ##class(%IPM.Main).Shell("version")
+				Return ""
+			}
+		}
+		// Shouldn't happen since %ZLANGC00.mac and %ZLANGF00.mac are present
+		Write "No namespace found with IPM enabled."
+	}
 	Quit ""
  }
 }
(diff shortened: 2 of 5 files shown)


src/cls/IPM/Main.cls
```diff
@@ -2627,6 +2669,59 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
+				Set key = $Order(list(key))
+				Quit:key=""
+				Set namespaces = namespaces _ $ListBuild(key)
+			}
+		} Else {
+			Set namespaces = $ListFromString(namespaces)
+		}
+		Set pointer = 0
+		While $ListNext(namespaces,pointer,namespace) {
+			Set $Namespace = namespace
```
isc-kiyer, line 2695: Nit: Defensive coding to string leading/trailing white space and control characters from namespace name before trying to switch to it

src/cls/IPM/Main.cls
```diff
@@ -2627,6 +2669,59 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
+
+	If map {
+		If globally {
+			Set namespaces = ""
+			Do ..GetListNamespace(.list)
+			Kill list("%ALL")
+			Set key = ""
+			For {
+				Set key = $Order(list(key))
+				Quit:key=""
```
isc-kiyer, line 2687: Nit: Avoid using post conditionals

src/cls/IPM/Main.cls
```diff
@@ -2627,6 +2669,59 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
+	}
+	If map && 'globally && (namespaces = "") {
+		$$$ThrowOnError($$$ERROR($$$GeneralError,"If mapping from the current namespace's routine database with -map, must specify either -globally or a list of namespaces with -ns"))
+	}
+
+	If map {
+		If globally {
+			Set namespaces = ""
+			Do ..GetListNamespace(.list)
+			Kill list("%ALL")
```
isc-kiyer, line 2683: Nit: Add a comment as to why removing %ALL so we don't forget the context.

src/cls/IPM/Main.cls
```diff
@@ -2627,6 +2669,59 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
+				Quit:key=""
+				Set namespaces = namespaces _ $ListBuild(key)
+			}
+		} Else {
+			Set namespaces = $ListFromString(namespaces)
+		}
+		Set pointer = 0
+		While $ListNext(namespaces,pointer,namespace) {
+			Set $Namespace = namespace
+			If $$$comClassDefined("%IPM.Storage.Module") && ##class(%IPM.Storage.Module).NameExists($$$IPMModuleName) {
```
isc-kiyer, line 2696: Move existence check for IPM in a specific namespace to a utility helper method so we can keep the checks consistent (will also be helpful for some HS utilities to call probably)

src/cls/IPM/Main.cls
```diff
@@ -1150,6 +1156,38 @@ ClassMethod Version(ByRef pCommandInfo) [ Internal ]
 	Quit $$$OK
 }
 
+ClassMethod AddIPMMappedNamespaces(ByRef list) [ Private ]
+{
+	Do ..GetListNamespace(.allNS,"*")
+	For i=1:1:+$Get(list) {
```
isc-kiyer, line 1162: Nit: Use $ListNext instead

src/cls/IPM/Main.cls
```diff
@@ -1150,6 +1156,38 @@ ClassMethod Version(ByRef pCommandInfo) [ Internal ]
 	Quit $$$OK
 }
 
+ClassMethod AddIPMMappedNamespaces(ByRef list) [ Private ]
```
isc-kiyer, line 1159: Nit: Document expected format of list

src/cls/IPM/Main.cls
```diff
@@ -1150,6 +1156,38 @@ ClassMethod Version(ByRef pCommandInfo) [ Internal ]
...
+		}
+		Set sourceDB = ##class(%SYS.Namespace).GetPackageDest(namespace,"%IPM")
+		Set mappedFrom(sourceDB) = namespace
+		Merge mappedFrom(sourceDB,"info") = list(i)
+	}
+
+	Set namespace = ""
+	For {
+		Set namespace = $Order(allNS(namespace))
+		Quit:namespace=""
```
isc-kiyer, line 1175: Nit: use if instead

isc-kiyer (changes requested):
@isc-tleavitt a few comments but looks good!

gjsjohnmurray:
> namespace-specific mappings aren't allowed to override %ALL mappings (!?).

This always struck me as an unfortunate design choice. Too late to get it changed now, but maybe a second pseudo-namespace %DEFAULT could be added whose mappings defer to namespace specific ones.

src/cls/IPM/Main.cls
```diff
@@ -611,11 +611,16 @@ generate /my/path -export 00000,PacketName2,IgnorePacket2^00000,PacketName3,Igno
...
 		<modifier name="allow-upgrade" description="If specified, will also check for IPM version in specified namespaces and upgrade if version is lower than the target version. By deafult, this modifier is not ste and will not allow upgrade." />
+		<modifier name="map" aliases="m" description="If specified, will map IPM code from the current namespace-default code database rather than installing a separate copy. Can be used with -globally to add a %ALL namespace mapping." />
 		<modifier name="quiet" aliases="q" description="Quiet mode. By default, this modifier is not set and will display the contents onto the terminal/caller command line." />
+		<modifier name="preview" aliases="p" description="Preview only - for -map option, shows the namespaces that would be enabled." />
 		<!-- Examples -->
+		<example description="Make IPM available in all namespaces (including %SYS) by mapping the version in the current namespace default routine database to the %ALL psuedo-namespace. Namespace-specific installation will override this.">
```
gjsjohnmurray, line 621: Is this description inaccurate since %ALL could not be used?

src/cls/IPM/Main.cls
```diff
@@ -611,11 +611,16 @@ generate /my/path -export 00000,PacketName2,IgnorePacket2^00000,PacketName3,Igno
...
-		<modifier name="globally" description="Will install IPM in all non-%SYS explicit namespaces that currently do not have IPM installed. By default, this modifier is not set and will not install globally."/>
+		<modifier name="globally" aliases="g" description="Will install IPM in all non-%SYS explicit namespaces that currently do not have IPM installed. By default, this modifier is not set and will not install globally."/>
 		<modifier name="local-only" description="If specified, only local artifacts will be used for installation. By default, this modifier is not set and will not limit to local artifacts." />
 		<modifier name="allow-upgrade" description="If specified, will also check for IPM version in specified namespaces and upgrade if version is lower than the target version. By deafult, this modifier is not ste and will not allow upgrade." />
+		<modifier name="map" aliases="m" description="If specified, will map IPM code from the current namespace-default code database rather than installing a separate copy. Can be used with -globally to add a %ALL namespace mapping." />
```
gjsjohnmurray, line 617: Is this description inaccurate since %ALL could not be used?

isc-tleavitt:
> > namespace-specific mappings aren't allowed to override %ALL mappings (!?).
> 
> This always struck me as an unfortunate design choice. Too late to get it changed now, but maybe a second pseudo-namespace %DEFAULT could be added whose mappings defer to namespace specific ones.

@gjsjohnmurray thanks for chiming in! I like the idea and will float it, but we won't be able to make it a prerequisite for this work.

isc-shuliu:
Not sure why the CIs are failing.

Tried to run the tests locally on [9b19985](https://github.com/intersystems/ipm/pull/485/commits/9b19985eb7ea388246c31063d49a73321b2d0d33) it still fails although CIs passed in May ...

isc-shuliu:
It appears the CI failure stems from this block https://github.com/intersystems/ipm/blob/bf8ed87be9f4abab369e93b217d5e42dc5d23bc6/src/cls/IPM/Lifecycle/Base.cls#L1069

isc-tleavitt:
Next steps here for @isc-shuliu who's carrying this work forward:
* Revert the changes to the MapOnly installer XData block and associated method, as they're causing the CI issues and are generally problematic
* Reevaluate against criteria in https://github.com/intersystems/ipm/issues/470

isc-shuliu:
CIs are failing specifically in I4H because migration was not successful (see [log](https://github.com/intersystems/ipm/actions/runs/11040492268/job/30668625597#step:10:1363)). This failure was introduced by changes in [bf6e699](https://github.com/intersystems/ipm/pull/485/commits/bf6e69973c179d3d55e59d9309c440f1510962a9) which attempts to perform data migration in all namespaces

Guess I shouldn't perform it in all namespaces according to this comment https://github.com/intersystems/ipm/blob/6a9dd0101cfadbb4d2a92bb660b0361ef32f12a9/src/cls/IPM/Utils/Migration.cls#L29 but I'm not sure why.

@isc-tleavitt Any insights?

src/cls/IPM/Main.cls
```diff
@@ -630,6 +635,21 @@ generate /my/path -export 00000,PacketName2,IgnorePacket2^00000,PacketName3,Igno
 		</example>
 	</command>
 
+<command name="unmap">
+	<summary>
+		Unmap %IPM package and rountines in specified namespaces. Will Skip non-mapped namespaces.
+	</summary>
+	<modifier name="namespaces" aliases="ns" value="true" description="Comma-separated namespaces in which IPM mapping needs to be deleted."/>
+	<modifier name="globally" aliases="g" description="Will unmap IPM in all explicit namespaces that currently do not have IPM installed. By default, this modifier is not set and will not install globally."/>
```
isc-tleavitt, line 643: Nitpick: "By default, this modifier is not set and will not install globally" is not appropriate here

isc-tleavitt:
> CIs are failing specifically in I4H because migration was not successful (see [log line 1362](https://github.com/intersystems/ipm/actions/runs/11040492268/job/30668625597#step:10:1363)). This failure was introduced by changes in [bf6e699](https://github.com/intersystems/ipm/pull/485/commits/bf6e69973c179d3d55e59d9309c440f1510962a9) which attempts to perform data migration in all namespaces
> 
> Guess I shouldn't perform it in all namespaces according to this comment
> 
> https://github.com/intersystems/ipm/blob/6a9dd0101cfadbb4d2a92bb660b0361ef32f12a9/src/cls/IPM/Utils/Migration.cls#L29
> 
> but I'm not sure why.
> @isc-tleavitt Any insights?

The issue here is:
```
Migrating ZPM data to IPM... in namespace HSCUSTOM
Migrating repo registry... 
IPM data migration was not fully successful. You may be able to reinstall the impacted packages to correct the issues.
ERROR #5809: Object to Load not found, class '%IPM.Storage.Module', ID '1'

Migrating ZPM data to IPM... in namespace HSLIB
IPM data migration was not fully successful. You may be able to reinstall the impacted packages to correct the issues.
ERROR #5002: ObjectScript error: <PROTECT>SetMapSelectability+17^%SYSTEM.SQL.Util.1 ^rINDEXEXT("MS","%IPM_Storage","ModuleItem","IDKEY"),/usr/irissys/mgr/hslib/

Migrating ZPM data to IPM... in namespace HSSYS
Migrating repo registry... 
IPM data migration was not fully successful. You may be able to reinstall the impacted packages to correct the issues.
ERROR #5809: Object to Load not found, class '%IPM.Storage.Module', ID '1'

Migrating ZPM data to IPM... in namespace HSSYSLOCALTEMP
Migrating repo registry... 
IPM data migration was not fully successful. You may be able to reinstall the impacted packages to correct the issues.
ERROR #5809: Object to Load not found, class '%IPM.Storage.Module', ID '1'
```

Two classes of issues here:
* HSLIB: here we can't migrate because the HSLIB database is read-only. In this case, we want to just skip. (Should be able to detect this.)
* Others: likely related to presence of IPM in HSLIB itself but kind of hard to pinpoint with available data.

Ultimately, I think the issue is going to be that we should only try to migrate a namespace's data after it's mapped.

Data migration post-installation should migrate everywhere that is already mapped (mostly will be useful for future automatic data migrations); mapping should migrate the newly-mapped namespace. The automatic mapping post-installation on an environment with older ZPM will end up migrating each namespace as it's mapped. That covers all scenarios.

src/cls/IPM/Utils/Migration.cls
```diff
@@ -5,19 +5,37 @@ Class %IPM.Utils.Migration
 
 ClassMethod RunAll(verbose As %Boolean = 1) As %Status
 {
+    New $Namespace
     Set sc = $$$OK
-    Try {
-        Do ..MigrateZPMToIPM(verbose)
-    } Catch e {
-        Set sc = e.AsStatus()
+    Do ##class(%IPM.Main).GetListNamespace(.list)
```
isc-tleavitt, line 10: Should run in namespaces with %IPM mapped from the current namespace's default routine database rather than in all namespaces.

isc-tleavitt (commented):
See also https://github.com/intersystems/ipm/pull/485#issuecomment-2377501538 - soooo close.

src/cls/IPM/Main.cls
```diff
@@ -2594,22 +2696,43 @@ ZPM(pArgs...)
  If $System.CLS.IsMthd("%IPM.Main", "Shell") { 
  	Do ##class(%IPM.Main).Shell(pArgs...) Quit
  } Else {
-	// TODO: Decide what to do here; definitely needs to support enabling IPM from here as well,
-	// but need to decide what level of customization to provide
+	// TODO: Needs to support enabling IPM from here, also need to decide what level of customization to provide
+	Write !, "IPM is not enabled in this namespace."
```
isc-tleavitt, line 2700: Recall: we need to add better error handling based on pArgs. This is used from two contexts, zpm "command":1:1 (say) and $zpm("command"). In the former, if the flag to terminate on error is set, we should terminate if IPM is not available. In the latter, if IPM is not available we want to return an error status.

isc-shuliu:
@isc-tleavitt I updated the migration logic to run only in namespaces where %IPM is mapped from the current namespace. This resolved the problem for HSLIB but still doesn't successfully migrate HSCUSTOM, HSSYS, and HSSYSLOCALTEMP.

In I4H, HSCUSTOM, HSSYS, and HSSYSLOCALTEMP did not have %IPM mapping initially. Upgrading from %ZPM to %IPM automatically adds %IPM mapping for them. However, data migration still fails in these 3 namespaces.

isc-shuliu:
@isc-tleavitt I think I figured out why:

In ##class(%IPM.Utils.Migration).MigrateOneRepo(), we should check for 
```objectscript
Set newObj = ##class(%IPM.Repo.Definition).%OpenId(newId,,.sc)
```
instead of
```objectscript
Set newObj = ##class(%IPM.Storage.Module).%OpenId(newId,,.sc)
```

This is likely an error due to copy-paste from `MigrateOnePackageToIPM()`

isc-tleavitt:
@isc-shuliu goooood catch! Thank you!

isc-shuliu (approved):

src/cls/IPM/Main.cls
```diff
@@ -2594,22 +2696,59 @@ ZPM(pArgs...)
...
  	Do ##class(%IPM.Main).Shell(pArgs...) Quit
  } Else {
-	// TODO: Decide what to do here; definitely needs to support enabling IPM from here as well,
-	// but need to decide what level of customization to provide
-	Quit ""
+	// TODO: Needs to support enabling IPM from here, also need to decide what level of customization to provide
+	Set tQuitOnError = $Get(pArgs(2))
+	Set tHaltOnComplete = $Get(pArgs(3))
+	Write !, "IPM is not enabled in this namespace."
+	&sql(SELECT %DLIST(Nsp) INTO :namespaces FROM %SYS.Namespace_List())
```
isc-kiyer, line 2703: In HS we've had issues in the past with embedded SQL and upgrades (with frozen query plans). Not sure if its still an issue but I prefer using dynamic SQL and just looping through the result set.

src/cls/IPM/Main.cls
```diff
@@ -2594,22 +2696,59 @@ ZPM(pArgs...)
...
+	If SQLCODE '= 0 {
+		Write !, "Error getting namespace name. SQLCODE =", SQLCODE
+	} Else {
+		New $Namespace
+		Set ptr = 0
+		Set found = 0	
+		While $ListNext(namespaces, ptr, ns) {
+			Set $Namespace = $Zstrip(ns, "<>WC")
+			If $System.CLS.IsMthd("%IPM.Main", "Shell") {
```
isc-kiyer, line 2712: I think we need helper method for IsIPMEnabled(). There is an IPMInstalled method below already so we have a single source of truth of such checks (also would rename it to IsIPMInstalled to indicate it returns a boolean).

src/cls/IPM/Main.cls
```diff
@@ -2594,22 +2696,59 @@ ZPM(pArgs...)
  If $System.CLS.IsMthd("%IPM.Main", "Shell") { 
  	Do ##class(%IPM.Main).Shell(pArgs...) Quit
  } Else {
-	// TODO: Decide what to do here; definitely needs to support enabling IPM from here as well,
-	// but need to decide what level of customization to provide
-	Quit ""
+	// TODO: Needs to support enabling IPM from here, also need to decide what level of customization to provide
+	Set tQuitOnError = $Get(pArgs(2))
```
isc-kiyer, line 2700: Nit: For net new code, don't use the legacy `t` prefix for variable names

src/cls/IPM/Main.cls
```diff
@@ -2594,22 +2696,59 @@ ZPM(pArgs...)
...
+	If tQuitOnError {
+		Write ! // Add a newline before quitting, otherwise the shell prompt will be on the same line as the error message
+		Do $System.Process.Terminate($Job, 1)
+	}
+	If tHaltOnComplete {
+		Write ! // Add a newline before halting
+		Halt
+	}
+	// The $$$ERROR macro is not available for use in language extension
+	Return $$Error^%apiOBJ(5001,"IPM is not enabled in this namespace.")
```
isc-kiyer, line 2734: Nit: Use the class method version instead of routine: $System.Status.Error().

src/cls/IPM/Main.cls
```diff
@@ -2594,22 +2696,59 @@ ZPM(pArgs...)
...
 
 ClassMethod EnableIPM(ByRef pCommandInfo)
 {
-	Write !, "Version of IPM in current namespace: "
-	Do ..GetListModules($Namespace,$$$IPMModuleName,.list)
-	If $Data(list($$$IPMModuleName)) {
-			Write !,($namespace)_"> "_$$$FormattedLine($$$Green,$$$IPMModuleName_" ")_$ListGet(list($$$IPMModuleName),1)
+	New $Namespace
+	Set initNamespace = $Namespace
+	If ##class(%IPM.Storage.Module).NameExists($$$IPMModuleName,.ipmModuleId) {
```
isc-kiyer, line 2742: Use IsIPMInstalled() helper method

isc-kiyer (changes requested):
@isc-shuliu a few comments. Can maybe skip the one about changing the structure of the subscripted array to JSON since that would be more invasive (just the current structure is quite hard to follow).

preload/cls/IPM/Installer.cls
```diff
@@ -35,6 +35,7 @@ XData PM [ XMLNamespace = INSTALLER ]
     <Arg Value="${MODDIR}"/>
   </Invoke>
   <Invoke Class="IPM.Installer" Method="ZPMCompile" CheckStatus="true" />
+  <Invoke Class="IPM.Installer" Method="MapIfLegacy" CheckStatus="true" />
```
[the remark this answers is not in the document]
isc-shuliu: My understanding is that `<Invoke>`s in module.xml should be a (not necessarily proper) subset of Installer.cls since a legacy user can always download a release artifact from GitHub or community PM registry. 

Not sure if there are `<Invoke>`s that we only want present in Installer.cls but not module.xml.

@isc-tleavitt Thoughts?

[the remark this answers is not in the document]
isc-tleavitt: Wait - I'm wrong, we shouldn't need this because the above (line 34) call to ZPMLoad will end up running all the `<Invoke>`s.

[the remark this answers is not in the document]
isc-tleavitt: I was confusing this with:
`<Invoke Class="IPM.Installer" Method="Map" Phase="Reload" When="Before" />`

src/cls/IPM/Main.cls
isc-tleavitt: @isc-kiyer the issue is that %IPM may not be mapped everywhere, so we need to duplicate code in the language extension (which is available instance-wide even in places where %IPM is not mapped)

preload/cls/IPM/Installer.cls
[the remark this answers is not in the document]
isc-kiyer: @isc-shuliu @isc-tleavitt so should the installer invokes be removed in favor of the Invoke's in module.xml? The installer then essentially installs the module using itself which would run invokes from the module.xml?

src/cls/IPM/Main.cls
```diff
@@ -1871,16 +1942,40 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
...
 		Set $ListBuild(tRegistry, tModuleName) = $ListFromString(tModuleName, "/")
 	}
-  If (tModuleName = "") {
-    Quit $$$OK
-  }
+	If (tModuleName = "") {
+		Quit $$$OK
+	}
+
+	If tModuleName = $$$IPMModuleName {
```
isc-kiyer, line 1976: I think this check should be moved to LoadNewModule instead to catch the case from the load command as well instead of just the install command. Before loading the file using $System.OBJ.Load(), can read the module name using xslts to confirm whether it can be loaded in the current namespace.

isc-kiyer (approved):

src/cls/IPM/Main.cls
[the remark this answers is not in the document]
isc-tleavitt: This is specific to the zpm command - should be covered in doc for %IPM.Main:Shell.

PR #488, opened 2024-05-30
Fix #487 Improve parsing of load path

### LINKED ISSUE #487: Improve parsing of load path
When running `zpm "load <path>"`, [the code](https://github.com/intersystems/ipm/blob/9978352e58b9926ce0ad0dd115f0edb16c99ed68/src/%25IPM/Main.cls#L1417) checks for whether `<path>` starts with http (a git repo) or not (local dir or tarball). 

While this is sufficient in almost all cases, a user might inadvertently parse in a relative path starting with http (w.r.t. `<IRISROOT>/mgr/`). A safer way is to check for prefix `http://` and prefix `https://`.

isc-shuliu:
Fixes #487

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -1857,12 +1857,13 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
 {
 	Set tDirectoryName = $Get(pCommandInfo("parameters","path"))
 	Merge tParams = pCommandInfo("data")
-  Set tParams("cmd") = "load"
-  if $Extract(tDirectoryName,1,4)="http" {
+	Set tParams("cmd") = "load"
+	Do ##class(%Net.URLParser).Decompose(tDirectoryName, .tComponents)
+	if $G(tComponents("scheme")) {
 		set tDirectoryName=..LoadFromRepo(tDirectoryName,.tParams)
 	}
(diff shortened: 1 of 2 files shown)


isc-kiyer (approved):

src/cls/IPM/Main.cls
```diff
@@ -1857,12 +1857,13 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
 {
 	Set tDirectoryName = $Get(pCommandInfo("parameters","path"))
 	Merge tParams = pCommandInfo("data")
-  Set tParams("cmd") = "load"
-  if $Extract(tDirectoryName,1,4)="http" {
+	Set tParams("cmd") = "load"
+	Do ##class(%Net.URLParser).Decompose(tDirectoryName, .tComponents)
+	if $G(tComponents("scheme")) {
```
isc-tleavitt, line 1862: This isn't quite right - strings aren't truthy. I would recommend:
if $Get(tComponents("scheme")) [ "http"

isc-tleavitt (changes requested):
As-is this breaks load over http/https

isc-tleavitt (changes requested):
This also needs a CHANGELOG.md update, and the issue description should include "Fixes #487" for a proper link in GitHub.

isc-tleavitt (approved):

PR #489, opened 2024-05-30
Reenable deployment of code without breaking embedded source control

### LINKED ISSUE #482: v1 - reenable deployed code export
Need to provide a proper fix for https://github.com/intersystems/ipm/blob/v1/src/cls/IPM/ResourceProcessor/Default/Document.cls#L191 to avoid export based on the module's DeveloperMode flag and/or active lifecycle status rather than the pParams array. (Needs to be tested with our internal source control extension.)

Should also add an integration test or two for deployed code export/load.

isc-tleavitt:
Fixes #482


isc-tleavitt:
@isc-dchui @isc-shuliu this is from our working session the other day.

isc-shuliu (approved):

isc-dchui (approved):

isc-kiyer (approved):

PR #491, opened 2024-05-31
v1 - reintroduce support for inc/rtn under PKG resource

### LINKED ISSUE #483: v1 - fix support for rtn/inc within PKG resource
From discussion in #428 - need to enable other document types in https://github.com/intersystems/ipm/blob/v1/src/cls/IPM/Storage/ResourceReference.cls#L129 same as https://github.com/intersystems/ipm/blob/master/src/%25ZPM/PackageManager/Developer/ResourceReference.cls#L125 and account for this so that only documents not explicitly listed in *any* module are exported.

isc-tleavitt:
Fixes #483

### DIFF
--- a/src/cls/IPM/Storage/ResourceReference.cls
+++ b/src/cls/IPM/Storage/ResourceReference.cls
@@ -126,7 +126,7 @@ ClassMethod GetChildren(pResourceName As %String, pModuleName As %String, pCheck
 			// PKG extension should only cover .CLS files
 			Set tFilesResult = ##class(%SQL.Statement).%ExecDirect(,
         		"select Name from  %Library.RoutineMgr_StudioOpenDialog(?,'',1,1,1,0,0)",
-				tPackage_"*.cls")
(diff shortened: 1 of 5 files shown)


isc-tleavitt:
@isc-kiyer @isc-bzhang I was unable to reproduce the issues related to this that were brought up in past conversation, though I did add unit tests around the case that we were worried about (that passed without needing any related code changes beyond reintroducing the line that was removed).

If you have or can find more information on why the other types were removed, that would be great. Is it possible that the source control extension could be misbehaving here?

isc-tleavitt:
@isc-bzhang (if you have bandwidth) @isc-kiyer (if not) bumping this back up. Would be good to know what the issues really were.

isc-tleavitt:
@isc-kiyer @isc-shuliu bumping this up again - trying to clean up old PRs and would like to get this merged.

isc-kiyer (approved):
@isc-tleavitt approving with the caveat that we may need some assistance with the source control hooks when adopting to make sure everything still works fine. I do like this change since it means we can reduce the number of .INC resources we have in our module.xml

PR #492, opened 2024-06-03
Attempt stripping % before keeping it when parsing resource path

### LINKED ISSUE #490: Default resource processor ignores percent (%) in path
The [default behaviour](https://github.com/intersystems/ipm/blob/b7ea8fa75f902fbfe7b2a20acc96321eba1e5020/src/cls/IPM/ResourceProcessor/Default/Document.cls#L32) of resource processor is to ignore percent (%) in resource paths.

It seems to be a problem for some legacy packages such as [zapm](https://github.com/SergeyMi37/zapm/blob/3b442780b65de5b7f837d28d472457b315df8446/module.xml#L16). 

@isc-tleavitt I wonder what is the rationale behind this convention to ignore "%" by default and if there is a way around it for package publishers (other than renaming their directory).

isc-shuliu:
Fix #490 . 
Attempt keeping % symbol in resource path before stripping it when parsing filesystem path.

@isc-tleavitt Does this count as a "Fix" in change log considering v0.7.0 doesn't seem to have this issue?

### DIFF
--- a/src/cls/IPM/ResourceProcessor/Default/Document.cls
+++ b/src/cls/IPM/ResourceProcessor/Default/Document.cls
@@ -128,7 +128,7 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
               Set identifiers = ..FilenameTranslateIdentifier
               Set associators = ..FilenameTranslateAssociator
               If allowPercentSymbol {
-                set pos = $Find(identifiers, "%")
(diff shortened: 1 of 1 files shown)


isc-shuliu:
@isc-tleavitt Does this count as a "Fix" in change log considering v0.7.0 doesn't seem to have this issue?

isc-tleavitt (approved):

Issue #475, opened 2024-04-30
Packages erroring on installation on v0.9 but not v0.7


isc-tleavitt:
Reviewing https://github.com/intersystems/ipm/actions/runs/8879408830/job/24377373103 vs. baseline https://github.com/intersystems/ipm/actions/runs/8883338832/job/24390310244 - install + test all packages with 4 or more released versions

*Many* packages are having issues with the FileCopy resource processor not picking up the proper root directory, trying to load from another directory instead. This is definitely an IPM regression.

Non-lifecycle dependencies - these are cases where the class rename has caused issues:
git-source-control (singleton, but I know there are more too; this one is mine)
zpm-generate-ui (%ZPM.PackageManager)
zpmhub (unknown)

Lifecycle parent class dependencies:
isc.ipm.js (this one is mine)
zapm
zapm-addcmd
zapm-editor

Other really random cases (possible IPM regressions):
interoperability-for-money fails with: ERROR! Missing role :%DB_IRISAPP
run-container-command-execution fails with: ERROR! Resource path '/usr/irissys/mgr/.modules/USER/run-container-command-execution/0.0.6/src/zrccx.xml' not found

isc-tleavitt:
Latest run with all the v1 fixes (https://github.com/intersystems/ipm/actions/runs/8899744512) shows all the FileCopy cases to be fixed, all the other cases still present.

shuheng-liu:
Update: I've figured out the reason for zpmhub to fail. This is also caused by %ZPM -> IPM package rename. See https://github.com/banksiaglobal/zpmhub/blob/00ba0598fffe7bdbe03277dbd90c8ea700cdbb09/src/ZPMHUB/Setup.cls#L25

isc-shuliu:
Additionally, interoperability-for-money fails because of this line https://github.com/oliverwilms/interoperability-for-money/blob/074bb6bca3712b3a0e0bc7eba379244bc77d4d16/module.xml#L20.

isc-shuliu:
As for appmsw-forbid-old-passwd, it's likely that the module.xml is not properly parsed to get the src/mac/PASSWORD.mac file. Instead, it's trying to find the src/PASSWORD.xml file.

See https://github.com/SergeyMi37/appmsw-forbid-old-passwd/blob/5c906f250b31306ddb3425c2abfac418c5d52546/module.xml#L27

isc-shuliu:
[...]

PR #495, opened 2024-06-05
Search legacy src/mac in addition to src/rtn when locating resources

isc-shuliu:
A previous commit (see [here](https://github.com/intersystems/ipm/blame/v1/src/cls/IPM/ResourceProcessor/Default/Routine.cls#L12)) changed the default subfolder containing routine files from `mac` to `rtn`. 

This PR ensures backward compatibility with legacy packages by checking for both `mac/` and `rtn/`. This fixes the error when installing certain packages (such as `run-container-command-execution` in #475 )


isc-tleavitt:
I don't think there's need to update CHANGELOG.md as this is a fix to an unreleased bug. @isc-kiyer is there a differing standard on this from an HS perspective?

isc-tleavitt (approved):
Good targeted solution.

isc-kiyer (approved):

isc-dchui (approved):

PR #507, opened 2024-06-18
CI: Allow testing multiple packages in a single workflow job

isc-shuliu:


isc-shuliu:
@isc-tleavitt Could you add an environment secret named "GITHUB_TOKEN" to the IPM repo, so I can use it to download and parse logs for workflow runs?

.github/workflows/packages-summary.yml
```diff
@@ -0,0 +1,34 @@
...
+
+jobs:
+  summarize:
+    runs-on: ubuntu-latest
+    env:
+      BAES_URL: https://api.github.com/repos/intersystems/ipm/actions/runs
+    steps:
+      - name: Donwload log zip
+        run: |
+          curl -L -o logs.zip "${{ env.BAES_URL }}/${{ inputs.runId }}/logs" -H 'Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}'
```
isc-tleavitt, line 24: @isc-shuliu I think this should be env.BASE_URL? This might be the issue rather than needing a GITHUB_TOKEN secret.

isc-tleavitt (changes requested):
I think there's a typo here. From my reading it seems that GITHUB_TOKEN is always/automatically available so shouldn't need to explicitly add it to secrets.

.github/workflows/packages-summary.yml
isc-shuliu: @isc-tleavitt Will definitely fix that. It is a typo indeed but I made the same typo on line 20 when defining the variable, so I think it shouldn't be a problem. I also ran `act` locally and verified that this workflow works.

If I try to run the curl without the authorization header, I got the following json response. According to [the documentation](https://docs.github.com/rest/actions/workflow-runs#download-workflow-run-logs) it appears the user must be authenticated to access this endpoint. Since this repo is public, any authenticated user will do
```json
{
  "message": "Must have admin rights to Repository.",
  "documentation_url": "https://docs.github.com/rest/actions/workflow-runs#download-workflow-run-logs",
  "status": "403"
}
```

isc-shuliu: FWIW, the `act` command I ran was this
```
act workflow_dispatch -W .github/workflows/packages-summary.yml --secret GITHUB_TOKEN --input runId=9570067277
```
It prompted me for value of `GITHUB_TOKEN` and ran successfully.

isc-tleavitt: We don't need to add GITHUB_TOKEN manually - see https://docs.github.com/en/actions/security-guides/automatic-token-authentication. I realized the typo was duplicated but didn't get to update my notes before then. :)

One thing that's a little weird, looking at e.g. https://github.com/intersystems/ipm/actions/runs/9570067277 it seems that we ended up perhaps DDOSing pm.community.intersystems.com! Going to follow up with that team to figure out what might have happened...

isc-tleavitt (approved):
Just drop the default # of workers down to 8 so we don't DDOS pm.community.intersystems.com (again!?)

evshvarov:
thank you!

PR #511, opened 2024-06-25
Fix deployed code publication for packages with unit tests

### LINKED ISSUE #505: Deployed code packaging fails with <ILLEGAL VALUE> in package with unit tests
See https://github.com/intersystems/ipm/blob/v1/src/cls/IPM/Lifecycle/Module.cls#L109 - need smarter handling for "no extension" and/or alternative extensions.

isc-tleavitt:
Fixes #505

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: '3'
 services:
   iris:
     build: .
@@ -38,3 +37,7 @@ services:
     volumes:
       - ./:/home/irisowner/zpm/
       - ./tests/sandbox/:/home/irisowner/sandbox/
+      - ./tests/sandbox/fhirpro/:/fhirpro/
+    command:
+      - -a 
+      - iris session iris -U%SYS '##class(Security.Users).UnExpireUserPasswords("*")'
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,6 @@ services:
     volumes:
       - ./:/home/irisowner/zpm/
       - ./tests/sandbox/:/home/irisowner/sandbox/
-      - ./tests/sandbox/fhirpro/:/fhirpro/
     command:
       - -a 
       - iris session iris -U%SYS '##class(Security.Users).UnExpireUserPasswords("*")'
--- a/src/cls/IPM/Lifecycle/Module.cls
+++ b/src/cls/IPM/Lifecycle/Module.cls
@@ -98,7 +98,9 @@ Method %PrepareDeploy(ByRef pParams) As %Status
 
 			Set tName = $Piece(tResource.Name, ".", 1, *-1)
 			Set tExtension = $$$ucase($Piece(tResource.Name, ".", *))
-			continue:tExtension=""
+			if (tExtension = "") || ($Extract(tExtension) = "/") {
+				continue
+			}
 
 			Set tSC = $Case(tExtension,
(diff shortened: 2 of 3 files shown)


tests/sandbox/Dockerfile
```diff
@@ -1,9 +1,15 @@
...
+ && apt-get install -y jq \
+ && apt-get install -y git
 
-USER irisowner
\ No newline at end of file
+USER irisowner
+
+RUN --mount=type=bind,src=.,dst=/home/irisowner/sandbox/ \
+  iris start iris && \
+  sh /home/irisowner/sandbox/install-artifact.sh zpm-0.7.1.xml \
```
isc-shuliu, line 14: Just an idea, how about we substitute `zpm-0.7.1.xml` with an ARG that defaults to `zpm-0.7.1.xml`? It's more handy for cases where we want to install betas of 0.9.0

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -98,17 +98,19 @@ Method %PrepareDeploy(ByRef pParams) As %Status
...
 
 			Set tSC = $Case(tExtension,
 				"GBL":##class(%IPM.Utils.Module).AddGlobalMapping(tPackageDeployNamespace, tName, tSource),
 				"INC":##class(%IPM.Utils.Module).AddRoutineMapping(tPackageDeployNamespace, tName, "INC", tSource),
 				"INT":##class(%IPM.Utils.Module).AddRoutineMapping(tPackageDeployNamespace, tName, "INT", tSource),
 				"MAC":##class(%IPM.Utils.Module).AddRoutineMapping(tPackageDeployNamespace, tName, "MAC", tSource),
 				"PKG":##class(%IPM.Utils.Module).AddPackageMapping(tPackageDeployNamespace, tName, tSource),
-				"":##class(%IPM.Utils.Module).AddRoutineMapping(tPackageDeployNamespace, tName, "ALL", tSource))
+				"":##class(%IPM.Utils.Module).AddRoutineMapping(tPackageDeployNamespace, tName, "ALL", tSource),
+				:$$$OK)
```
isc-shuliu, line 112: Question: is it our intention to allow other extension names to fail silently?

docker-compose.yml
```diff
@@ -37,4 +36,8 @@ services:
       - IRISNAMESPACE=USER
     volumes:
       - ./:/home/irisowner/zpm/
-      - ./tests/sandbox/:/home/irisowner/sandbox/
\ No newline at end of file
+      - ./tests/sandbox/:/home/irisowner/sandbox/
+      - ./tests/sandbox/fhirpro/:/fhirpro/
```
isc-shuliu, line 40: I'm not seeing a 1tests/sandbox/fhirpro/` folder. Is it supposed to be created on first run? If so, shall we add that to .gitignore?

isc-shuliu (changes requested):
A few questions and comments

src/cls/IPM/Lifecycle/Module.cls
isc-tleavitt: Yes.

docker-compose.yml
isc-tleavitt: This is debug code, should kill it.

isc-shuliu (approved):

isc-kiyer (approved):

PR #518, opened 2024-07-08
allow `-showsource` as a modifier to `list-installed`

isc-shuliu:


isc-tleavitt (approved):
Could add CHANGELOG.md update but otherwise looks good.

PR #519, opened 2024-07-11
feat(ResourceProcessor): support WSGIApplication as a resource

isc-shuliu:
Added a new WSGIApplication resource processor that handles WSGI applications like [this](https://github.com/isc-shuliu/flask-demo/). 

TODO:
- [x] Basic functionality.
- [ ] Need more work to make `FileCopy` a part of `<WSGIApplication>` automatically.
- [x] Add integration test
- [ ] A developer community article and a wiki page in IPM


isc-tleavitt:
Can this extend the base CSPApplication class rather than being a copy of it?

src/%ZPM/PackageManager/Developer/Lifecycle/Base.cls
```diff
@@ -540,6 +535,32 @@ Method InstallPythonRequirements(pRoot As %String = "", ByRef pParams)
   Quit tSC
 }
 
+ClassMethod RunPipCommand(cmd, verbose) As %Boolean [ Language = python ]
```
isc-tleavitt, line 538: This is our first Language = python method in IPM. I think appropriate for v1 but not master. We need to discuss.

isc-shuliu: As discussed in today's meeting, this change has been reverted

isc-shuliu:
> Can this extend the base CSPApplication class rather than being a copy of it?

@timleavitt Done. Will need future work to remove unnecessary properties.

src/%ZPM/PackageManager/Developer/Processor/WSGIApplication.cls
```diff
@@ -0,0 +1,46 @@
+Include (%sySecurity, %occErrors)
+
+Class %ZPM.PackageManager.Developer.Processor.WSGIApplication Extends %ZPM.PackageManager.Developer.Processor.CSPApplication
+{
+
+Parameter ATTRIBUTES = {""_
```
isc-kiyer, line 6: @isc-shuliu I would recommend changing this to be generic WebApplication rather than something specific for WSGI. Also, I created an initial version of such a generic one here: 
//projects/healthshare/appmodules/HSIEO-WIP/ZHSLIB/databases/hscommon/cls/ZHSLIB/PackageManager/Developer/Processor/WebApplication.cls

The approach in the file above, lets the processor class automatically include and support any new props of Security.Applications instead of a hardcoded list. Furthermore, it doesn't do automatic copying of files (which I think is good). Since the user can use the FileCopy resource processor if they want that. This limits side effects and allows them to just create a web app.

Apologies for the delay with the review. Wish I had gotten to it before you checked it in.

cc @isc-tleavitt

PR #528, opened 2024-07-29
fix: ignore case when checking for file existence

### LINKED ISSUE #527: Case sensitivity for resource files causing installation error on Linux only
## Bug
@isc-tleavitt @isc-kiyer I recently noticed that certain packages (such as `alwo-ompare`) can only be installed successfully on Mac/Windows, but not on Linux/containers. I've also found the underlying reason
https://github.com/intersystems/ipm/blob/64abbd3be43a5ae049c9e0c05922904a8e69a4a2/src/cls/IPM/ResourceProcessor/Default/Document.cls#L154

## Cause
This line checks for the existence of a `<IRIS-INSTANCE-ROOT>/mgr/.modules/USER/alwo-ompare/1.0.1/udl/ompare/Schedule.cls` file in the case of `alwo-ompare`. Although this file exists, it ends with capital `.CLS` instead of `.cls`.  This doesn't matter for Mac/Windows, but causes a problem for Linux.

## Solution & Discussion
I wonder if it makes sense to force filename checking to be case-insensitive in IPM. There is a utility function `##class(%IPM.Utils.File).Exists()` that does basically the same thing as `##class(%File).Exists()` except that the former ignores casing for all platforms where as the latter is platform dependent.

I think it might be potentially good, or even necessary, for us to enforce case-insensitivity for package publishers. The reason is that when developer from a case-sensitive platform publish packages on IPM, it should be installable by users on all platforms. If there are two files whose paths only differ in casing, it will cause file overwriting (or otherwise unpredictable behavior).

Additionally, it's uncommon for people to have 2 case-differing but otherwise identical files. For windows/mac users, they can't even do it by default. For Linux users, when they distribute a folder containing such files using git, it will also trigger a warning when others run `git clone` on windows/mac.

isc-shuliu:
Fixes #527

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - HSIEO-10274: Separate DependencyAnalyzer out from IPM
 - #261: IPM now truly supports using multiple registries for installation / discovery of packages (without needing to prefix the package with the registry name on "install", although it is still possible and now effective to use the prefix).
 - #454: IPM 0.9.x+ uses different globals for storage vs. 0.7.0 and previous. Installation will automatically migrate data from the old globals to the new ones. The old globals are left around in case the user decides to revert to an earlier version.
+- #527: IPM 0.9.x+ ignores the casing of resources when matching files on disk even on case-sensitive filesystems
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


isc-tleavitt (changes requested):
Should have CHANGELOG.md entry

isc-tleavitt (approved):

PR #538, opened 2024-07-30
enhance: allow specifying pip/pip3 path

isc-shuliu:
Allow specifying path to `pip` or `pip3` for installation of requirements.txt.

On many Unix systems, `pip` command is a symlink to `pip3` located somewhere on $PATH. In some cases, `pip` can be nonexistent and there's only `pip3`.

With this PR, users can specify path to `pip`/`pip3` by running `config set PipPath xxxx` in zpm. Otherwise, the default location is wherever `pip3` first appears on $PATH. This allows for maximum compatibility.

master equivalent: #543

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -31,7 +31,7 @@ jobs:
             do
               # Skip irishealth-community due to bad interaction with ZPM document type
               # Also skip 2023.2 because the license has expired
-              if [ "$n" = "irishealth-community" -a test "$tag" = "2023.3" -o "$tag" = "2023.2" ];
+              if [ "$n" = "irishealth-community" -a "$tag" = "2023.3" -o "$tag" = "2023.2" ];
                 then
                   continue
               fi
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -30,7 +30,8 @@ jobs:
             for tag in $tags
             do
               # Skip irishealth-community due to bad interaction with ZPM document type
-              if test "$n" = "irishealth-community" && test "$tag" = "2023.3"
+              # Also skip 2023.2 because the license has expired
+              if [ "$n" = "irishealth-community" -a test "$tag" = "2023.3" -o "$tag" = "2023.2" ];
                 then
                   continue
               fi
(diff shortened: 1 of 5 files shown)


isc-tleavitt:
CI failures are because the 2023.2 built-in license expired. We should cut that from CI.

isc-tleavitt (changes requested):
For 2024.2 / flexible python runtime I think the python executable should be located based on PythonRuntimeLibrary - gave this a try without it (and with WSGIApplication ported to v1, which we still need to do) and I'm running into issues because of different python versions being in play.

For <=2024.1 and Windows specifically, I think we need to stick with irispip to get things right.

isc-shuliu:
> For 2024.2 / flexible python runtime I think the python executable should be located based on PythonRuntimeLibrary - gave this a try without it (and with WSGIApplication ported to v1, which we still need to do) and I'm running into issues because of different python versions being in play.
> 
> For <=2024.1 and Windows specifically, I think we need to stick with irispip to get things right.

@isc-tleavitt What is `PythonRuntimeLibrary`? Is it a $System function or something else?

isc-shuliu:
@isc-tleavitt 
I made the changes to check for `<IRIS-ROOT>/bin/irispip.exe` first. If found, it will use that on windows. For 2024.2+, where `irispip.exe` doesn't exist, we proceed to check for `python3.exe -m pip` and `python.exe -m pip`.

For Unix, we still use `python3 -m pip` and `python -m pip`, whichever found first.

Does that make sense?

isc-tleavitt:
@isc-shuliu thanks - that's progress.

The problem is that we're finding python3/python based on the path, and it's possible if not likely that there would be multiple versions installed. It looks like, from some basic experimentation, if we pip install with the wrong version things end up broken. You know Python way better than I do - does this sound right? (looks like related to having the wrong cpython stuff)

Re: PythonRuntimeLibrary, see: https://docs.intersystems.com/iris20242/csp/docbook/Doc.View.cls?KEY=RACS_PythonRuntimeLibrary - this is new to 2024.2

isc-shuliu:
This should also fix #546

isc-tleavitt (approved):

isc-kiyer (approved):

Issue #514, opened 2024-07-05
Rename repo setting "DeploymentEnabled"


isc-dchui:
The repo setting "DeploymentEnabled" is rather misleading as a repo with "DeploymentEnabled"=0 can still be published it. Instead, it is a way of denoting the default repo that is published to and should be renamed to "DefaultDeploymentRepo".

isc-shuliu:
Fixed in #569

PR #541, opened 2024-08-01
Add support for OCI registries through ORAS

### LINKED ISSUE #581: Support "real" registries with ORAS
Just realizing we never had an issue for this (though we do have a PR); it relates to #281 but ultimately is a distinct task.

As part of this, we need to support deployed artifacts with the possibility of multiple artifacts of the same package version targeting different platform versions, and have the client be able to distinguish between them and select the right one.

isc-dchui:
Introduce support for OCI registries using ORAS. 

Mostly uses the ORAS Python client library via embedded Python, but augmented with ObjectScript and REST.

A docker container containing a simple OCI registry (ZOT) has been added to the docker-compose. To test out the containerized zot registry, in IPM configure an ORAS registry with http://oras:5000 as the URL.

Screenshot 2024-08-01 161527 50%

It can then be published to and installed from like any other remote registry. Visit [http://localhost:5001](http://localhost:5001) to see the registry's UI.

Fixes #581


requirements.txt
```diff
@@ -0,0 +1 @@
+oras == 0.1.30
```
isc-kiyer, line 1: If we are using embedded python in IPM, does this further limit the earliest IRIS version that IPM can be installed on?
I believe requirements.txt is currently always enforced. We should maybe make this optional so that IPM can still be installed on older versions of IRIS which don't have embedded python with just ORAS support disabled unless embedded python exists.

src/cls/IPM/Repo/Oras/Definition.cls
```diff
@@ -0,0 +1,183 @@
+Class %IPM.Repo.Oras.Definition Extends %IPM.Repo.Definition
+{
+
+Parameter DISPLAYNAME As STRING = "ORAS";
+
+Parameter MONIKER As STRING = "oras";
+
+Property URL As %String(MAXLEN = 2048) [ Required ];
```
isc-kiyer, line 8: Rather than repeating a bunch of properties, can this extend the remote repo definition or move the common properties to a common super class in the %IPM.General package?

src/cls/IPM/Repo/Oras/Definition.cls
```diff
@@ -0,0 +1,183 @@
...
+/// provide notification that the object is being saved. It is called before 
+/// any data is written to disk.
+/// 
+/// <P><VAR>insert</VAR> will be set to 1 if this object is being saved for the first time.
+/// 
+/// <P>If this method returns an error then the call to <METHOD>%Save</METHOD> will fail.
+Method %OnBeforeSave(insert As %Boolean) As %Status [ Private, ServerOnly = 1 ]
+{
+	// Leave DeploymentEnabled null to make unique index work right
+	// (i.e., deployment can only be enabled for one remote server.)
```
isc-kiyer, line 52: Why can deployment be enabled only for one remote server? If it is enabled for multiple, I think the user should be able to choose which one to deploy to at time of publishing

src/cls/IPM/Repo/Oras/Definition.cls
```diff
@@ -0,0 +1,183 @@
...
+	Set tClient.Namespace = ..Namespace
+	Set tClient.Username = ..Username
+	Set tClient.Password = ..Password
+    Set tClient.Token = ..Token
+	Quit tClient
+}
+
+/// Outputs information about this server to the current device.
+/// Subclasses may override to show additional information.
+Method Display()
```
isc-kiyer, line 83: A lot of the logic here is similar to that of the existing remote repo so could do some refactoring into helper methods. I think this would be true for not just ORAS but any other remote repositories we add support to in future

src/cls/IPM/Repo/Oras/Definition.cls
```diff
@@ -0,0 +1,183 @@
...
+	} Catch e {
+		Set tSC = e.AsStatus()
+	}
+	Quit tSC
+}
+
+Method GetSortOrder() As %Integer [ Private ]
+{
+	// TODO: Allow this to be changed dynamically.
+	Quit 2 + $Case(..%Id(),"":1,:..%Id())
```
isc-kiyer, line 153: What is the need for the "2 +"? If there is no Id then this hasn't yet been saved so the sort order wouldn't be relevant anyway right?

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -0,0 +1,463 @@
...
+	If name'="", ..Namespace'="" {
+		Set name = ..AppendURIs(..Namespace, name)
+	}
+
+	#; Get all modules
+	Set tList = ##class(%Library.ListOfObjects).%New()
+	Set request = ..GetHttpRequest()
+	
+	#; Make GET request
+	Set tSC=request.Get("/v2/_catalog")
```
isc-kiyer, line 71: Is this the base URI for all ORAS registries? If not, this should be something that the user can configure when creating the repo

src/cls/IPM/Repo/Oras/PublishService.cls
```diff
@@ -0,0 +1,168 @@
...
+    #; Empty temp directory first
+    If ##class(%File).DirectoryExists(tempDirectory) {
+        Set tSC = ##class(%IPM.Utils.File).RemoveDirectoryTree(tempDirectory)
+        $$$ThrowOnError(tSC)
+    }
+
+    #; Create temp directory
+    Set tCreated = ##class(%File).CreateDirectoryChain(tempDirectory,.tReturnValue)
+    If 'tCreated {
+		$$$ThrowStatus($$$ERROR($$$GeneralError,$$$FormatText("Error creating directory %1: %2",tempDirectory,tReturnValue)))
```
isc-kiyer, line 22: Nit: Fix spacing

src/cls/IPM/Repo/Oras/PublishService.cls
```diff
@@ -0,0 +1,168 @@
...
+    Set moduleXML = ""
+    While (pModule.Manifest.AtEnd = 0) {
+        Set moduleXML = moduleXML _ pModule.Manifest.ReadLine(,.status)
+        $$$ThrowOnError(status)
+    }
+
+	#; Set metadata
+	#; see https://github.com/opencontainers/image-spec/blob/main/annotations.md
+	Set metadata = {}
+	Do metadata.%Set("org.opencontainers.image.title", pModule.Name)
```
isc-kiyer, line 40: For this metadata used in both publishing and retrieval, would be good to have an ObjectScript class to represent it. You can define properties with non-alphanumeric characters using double quotes.
That class can extend %JSON.Adaptor.

src/cls/IPM/Repo/Oras/PublishService.cls
```diff
@@ -0,0 +1,168 @@
...
+		$$$ThrowStatus($$$ERROR($$$GeneralError,"Must specify version."))
+	}
+
+	#; Remove whitespace from tags
+	Set tags = $zstrip(tags,"*W")
+
+	Try {
+		Do ..PushPy(registry, repo, namespace, tags, path, metadata, username, password, token)
+	} Catch sc {
+		$$$ThrowStatus($$$ERROR($$$GeneralError,"Unable to push."))
```
isc-kiyer, line 95: Should print sc details

src/cls/IPM/Main.cls
```diff
@@ -2026,7 +2032,13 @@ ClassMethod Unpublish(ByRef pCommandInfo) [ Internal ]
 	Try {
 		Set tManager = ##class(%IPM.Repo.Remote.PublishManager).%Get(.tSC)
```
isc-kiyer, line 2033: To keep this consistent, I think the new repo should also have its own PublishManager and then only instantitate the publish manager based on the type value returned above (can instantiate it when setting tServer itself.)

isc-kiyer (changes requested):
@isc-dchui done an initial pass (not very thoroughly). Left some comments/questions

requirements.txt
isc-dchui: Thanks for the feedback! Just had a discussion with @isc-tleavitt and it seems one possible approach is allow IRIS versions without embedded python to tolerate a failure to compile requirements.txt and handle run time issues by preventing the creation of ORAS repository definitions. This would also mean avoiding [Language = python ] methods so that the classes compile.

src/cls/IPM/Repo/Oras/Definition.cls
isc-dchui: Yes, I'll go refactor by creating IPM.Repo.HTTP parent classes that both IPM.Repo.Remote and IPM.Repo.ORAS can inherit from.

src/cls/IPM/Repo/Oras/Definition.cls
isc-dchui: The field is misleadingly named (see [514](https://github.com/intersystems/ipm/issues/514)) and actually is specifying the default. The repo to be published to can be specified using the "-r" flag.

src/cls/IPM/Repo/Oras/Definition.cls
isc-dchui: Honestly not too sure, but it is from IPM.Repo.Remote.Definition. I'll take a closer look.

src/cls/IPM/Repo/Oras/PackageService.cls
isc-dchui: Yep that's the standard REST endpoint for OCI registries. See [here](https://distribution.github.io/distribution/spec/api/).

src/cls/IPM/Repo/Oras/Definition.cls
isc-dchui: Done.

src/cls/IPM/Repo/Oras/Definition.cls
isc-dchui: Refactored into superclass.

src/cls/IPM/Repo/Oras/Definition.cls
isc-dchui: Looks like the "2" is related to %IPM.Repo.FileSystem.Definition:
```
Method GetSortOrder() As %Integer [ Private ]
{
	Quit 2
}
```

src/cls/IPM/Repo/Oras/PublishService.cls
isc-dchui: Done.

src/cls/IPM/Repo/Oras/PublishService.cls
isc-dchui: Done.

src/cls/IPM/Main.cls
isc-dchui: Done.

src/cls/IPM/Repo/Oras/PublishService.cls
```diff
@@ -0,0 +1,164 @@
...
+{
+
+Method PublishModule(pModule As %IPM.Repo.Remote.ModuleInfo) As %Boolean
+{
+    Set repo = pModule.Name
+    Set tag = pModule.VersionString
+
+    #; Use a temp directory
+    Set tempFile = ""
+    Set tempDirectory = ##class(%File).NormalizeDirectory("/tmp/" _ repo _ "/" _ tag _ "/")
```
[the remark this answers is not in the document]
isc-dchui: Good catch. Fixed.

src/cls/IPM/Repo/Oras/PublishService.cls
isc-dchui: Done.

isc-dchui:
@isc-kiyer, @isc-shuliu fixed the majority of the requested changes. The only issue is the backwards compatibility with embedded Python. Unfortunately it looks like removing all `[ Language = python ]` methods has not proven straightforward. For example, take this snippet of Python code
```
# library function that returns a dictionary
manifest = client.get_manifest(target)
for key in manifest['annotations']:
	if key == "com.intersystems.ipm.module.v1+xml":
		moduleXML = manifest['annotations'][key]
		# remove whitespace
		return "".join(moduleXML.split())
```
Unless there's something I'm missing (which could be very possible), iterating through the Python dictionary in Objectscript is incredibly painful.
@isc-tleavitt Thoughts or suggestions?

isc-kiyer:
@isc-dchui Hmm I checked and embedded python was introduced in 21.2 and the first EM release to include it was 22.1. If we want to say we don't support older versions of IRIS for IPM v1, we could include this embedded python (I am fine with that but would want @isc-tleavitt to confirm). That being said, we need to make sure we don't hit any bugs embedded python may have had in those older versions so would need to test this so am a bit conflicted.

isc-tleavitt:
In weekly meeting today, https://openexchange.intersystems.com/package/PyHelper came up as an option - specifically ArrayFrompyDict looks useful.

isc-tleavitt:
Discussed with @isc-shuliu today. Plan is:
* I've polled on DC to see if anyone really wants earlier version compatibility: https://community.intersystems.com/post/anyone-using-ipm-earlier-20221
* @isc-shuliu is going to spend some time kicking the tires on this, then we'll merge to v1-next and *possibly* take removal [ Language = python ] methods as a follow-up action. This will get ORAS out of a feature branch and into a branch that will also be getting related fixes that HS/TC will need.

isc-tleavitt:
@isc-shuliu thought on a simple approach the migration failures: you could make the new intermediate HTTP repo class [NoExtent] and match the old storage definition in the subclass.

isc-tleavitt:
> @isc-shuliu thought on a simple approach the migration failures: you could make the new intermediate HTTP repo class [NoExtent] and match the old storage definition in the subclass.

This was silly; it's an intermediate class that already extends a class with an extent so it doesn't work. Alternative: just use the old storage def for the new intermediate class, which happens to work because all the properties moved to the intermediate class.

isc-tleavitt (approved):
I think we're good to go here.

src/cls/IPM/Repo/Http/PackageService.cls
```diff
@@ -0,0 +1,92 @@
...
+		Set proxy=$System.Util.GetEnviron("https_proxy")
+		If proxy'="" {
+			Do ##class(%Net.URLParser).Parse(proxy,.pr)
+			If (tRequest.Https) {
+				Set tRequest.ProxyHTTPS=1
+				Set tRequest.ProxyTunnel=1
+			}
+			Set tRequest.ProxyPort=pr("port")
+			Set tRequest.ProxyServer=pr("host")
+			If $Get(pr("username"))'="",$Get(pr("password"))'="" {
```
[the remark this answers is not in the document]
isc-shuliu: Since we are getting the shell environment variable `https_proxy`. it always takes the form `http://username:password@proxyhost:proxyport` which doesn't contain token. We're parsing this string using `%Net.URLParser:Parse` so I don't think we need to consider token authentication.

Actually I don't see why we need the whole proxy settings in the first place. Should have asked Dominic while he was here.

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -0,0 +1,412 @@
...
+	}
+	Set statusCode = request.HttpResponse.StatusCode
+	Quit (200 <= statusCode) && (statusCode < 300)
+}
+
+Method GetModule(pModuleReference As %IPM.Storage.ModuleInfo, Output AsArchive As %Boolean = 0) As %Stream.Object
+{
+	Set name = pModuleReference.Name _ ":" _ pModuleReference.VersionString
+	Set status = ..Pull(..Location, name, ..Namespace, ..Username, ..Password, ..Token, .stream)
+	If $$$ISERR(status) {
```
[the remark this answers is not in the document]
isc-shuliu: I'm unable to find this macro and it doesn't compile on my instance. Can you point me to the INC file where this macro is defined?

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -0,0 +1,412 @@
...
+	#; Handle results
+	Set json = response.Data.ReadLine()
+	Set data = ##class(%DynamicAbstractObject).%FromJSON(json)
+	Set iter = data.repositories.%GetIterator()
+	While iter.%GetNext(.key, .package, .type) {
+		If (package="") { 
+			Continue
+		}
+		#; filter by module name
+		If (name'="") && (package'=name) {
```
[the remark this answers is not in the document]
isc-shuliu: Just added comments for that

isc-tleavitt (approved):
Looks to me like this is in good shape!

src/cls/IPM/Repo/Oras/PublishService.cls
```diff
@@ -0,0 +1,170 @@
...
+
+		#; Push file
+		$$$ThrowOnError(..Push(..Location, repo, ..Namespace, tempDirectory, tag, metaDataString, ..Username, ..Password, ..Token, ..TokenBasicAuth))
+	} catch ex {
+		Set status = ex.AsStatus()
+	}
+	// Clean up temp directory whether successful or not
+    If ($Data(tempDirectory) # 2) && ##class(%File).DirectoryExists(tempDirectory) {
+        Set tSC = ##class(%IPM.Utils.File).RemoveDirectoryTree(tempDirectory)
+        $$$ThrowOnError(tSC)
```
isc-kiyer, line 57: Accumulate this status with the one above or else `status` will get lost if it fails to clean up the temp directory

isc-kiyer (changes requested):
@isc-shuliu final round of comments!

src/cls/IPM/Repo/Http/Definition.cls
```diff
@@ -0,0 +1,176 @@
...
+<commands>
+<command name="repo">
+<group name="HTTP(s) repository">
+
+<modifier name="url" value="true" description="For HTTP(s) repositories, specifies the URL for package retrieval web services." />
+<modifier name="publish-url" value="true" description="For HTTP(s) repositories, specifies the URL for package publishing web services."/>
+<modifier name="username" aliases="user" value="true" description="For HTTP(s) repositories, specifies the username to use when connecting." />
+<modifier name="password" aliases="pass" value="true" description="For HTTP(s) repositories, specifies the password to use when connecting." />
+<modifier name="token" value="true" description="For HTTP(s) repositories, specifies the key to use when connecting." />
+<modifier name="token-basic-auth" value="false" description="If set, token will be included as basic auth instead of in apikey header." />
```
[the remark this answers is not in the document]
isc-shuliu: What is the difference between bearer and basic?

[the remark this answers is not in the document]
isc-kiyer: Those are different options for authorizations headers. Based on the protocol, may use one vs the other. For OAuth for example, the token is passed using the bearer header and basic is used for username + password/sometimes api keys (api key could also be bearer based on the Identity provider).

src/cls/IPM/Repo/Http/PackageService.cls
```diff
@@ -0,0 +1,97 @@
+Include %IPM.Common
+
+Class %IPM.Repo.Http.PackageService Extends (%RegisteredObject, %IPM.Repo.IPackageService)
+{
+
+Parameter ALLOWPREFIXINLOCATION = 1;
+
+Property Location As %String;
```
isc-eneil, line 8: Nit: If I'm understanding correctly, this can be set to the %IPM.DataType.RepoLocation datatype like the URL IPM.Repo.Http.Definition.cls

isc-eneil (commented):
@isc-shuliu I left a few small comments/questions!

src/cls/IPM/Main.cls
```diff
@@ -1767,6 +1768,9 @@ ClassMethod Repository(ByRef pCommandInfo) [ Internal ]
 		}
 		While tRepoTypeResult.%Next(.tSC) {
 			$$$ThrowOnError(tSC)
+			If tRepoTypeResult.%Get("Moniker") = "http" {
```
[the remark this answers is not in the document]
isc-shuliu: It's because the class hierarchy. Both _Remote_ and _Oras_ repos are subclasses of _Http_ and _Http_ itself is not an abstract class. Normally, _Http_ repos will never be created because the zpm `repo` command doesn't allow for such an operation. However, if something goes wrong or if a user manually created a instance of this type, then we need to ignore it (or even raise a warning/error).

isc-eneil (approved):

isc-jili (approved):

src/cls/IPM/Repo/Oras/Definition.cls
```diff
@@ -0,0 +1,167 @@
+Class %IPM.Repo.Oras.Definition Extends %IPM.Repo.Http.Definition
+{
+
+Parameter DISPLAYNAME As STRING = "ORAS";
+
+Parameter MONIKER As STRING = "oras";
+
+Property Namespace As %String;
```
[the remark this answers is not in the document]
isc-shuliu: Created a new issue for adding the test case.

requirements.txt
isc-kiyer: @isc-shuliu @isc-tleavitt Not a blocker for this PR but what happens behind a firewall with installing python dependencies? Should IPM be packaged with a version of oras as a backup in case the internet is not available to install the latest version?
cc @isc-eneil @isc-jili @isc-jlechtne

isc-kiyer (approved):

PR #556, opened 2024-08-27
Fix hashtag

### LINKED ISSUE #498: v1 - #{objectscript-expression} syntax removed
v0.7.x has, in %ZPM.PackageManager.Developer.Module:%Evaluate,

```
  Set regex = ##class(%Regex.Matcher).%New("#\{([^}]+)\}", tAttrValue)
  While regex.Locate() {
    Set expr = regex.Group(1)
    Set value = ..%EvaluateExpression(expr)
    Set $Extract(tAttrValue, regex.Start, regex.End - 1) = value
    Set regex.Text = tAttrValue
  }
 
    Quit tAttrValue
}
 
Method %EvaluateExpression(pExpr) As %String [ Internal ]
{
  Try {
    return @pExpr
  } Catch ex {
  }
  Return ""
}
```

This has been removed in v0.9.x, perhaps intentionally. Need to figure out if we're documenting or restoring this feature.

isc-tleavitt:
Fixes #498 

Not bothering with CHANGELOG.md since this fixes an unreleased regression.

### DIFF
--- a/src/cls/IPM/Storage/Module.cls
+++ b/src/cls/IPM/Storage/Module.cls
@@ -1434,7 +1434,24 @@ Method %Evaluate(pAttrValue As %String, ByRef pParams) As %String [ Internal ]
 	Set name = ..Name
 	Set attrValue = ##class(%Regex.Matcher).%New("(?i)\{name\}", attrValue).ReplaceAll($Replace(name,"\","\\"))
 
+	Set regex = ##class(%Regex.Matcher).%New("#\{([^}]+)\}", tAttrValue)
+	While regex.Locate() {
+		Set expr = regex.Group(1)
+		Set value = ..%EvaluateExpression(expr)
+		Set $Extract(attrValue, regex.Start, regex.End - 1) = value
+		Set regex.Text = attrValue
+	}
+
 	Return attrValue
+}
+
+Method %EvaluateExpression(pExpr) As %String [ Internal ]
+{
+  Try {
+    return @pExpr
+  } Catch ex {
+  }
+  Return ""
 }
 
 Storage Default
@@ -1541,3 +1558,4 @@ Storage Default
 }
 
 }
+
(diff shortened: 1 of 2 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -1434,9 +1434,26 @@ Method %Evaluate(pAttrValue As %String, ByRef pParams) As %String [ Internal ]
...
+		Set expr = regex.Group(1)
+		Set value = ..%EvaluateExpression(expr)
+		Set $Extract(attrValue, regex.Start, regex.End - 1) = value
+		Set regex.Text = attrValue
+	}
+
 	Return attrValue
 }
 
+Method %EvaluateExpression(pExpr) As %String [ Internal ]
```
isc-kiyer, line 1448: Should we support evaluation of arbitrary expressions from a security perspective? I thought we just do string interpolation of the expressions

isc-tleavitt: From a security perspective, IPM already supports evaluation of arbitrary expressions in a much less transparent manner via Invoke elements. There are times where this kind of approach is necessary.

isc-tleavitt: Recent motivating example: needing to be able to reference ^%SYS("WebServer","URLPrefix") in a deployment URL for a React application, so that it'll work on any local IRIS instance using a real webserver. This is something that could ultimately have its own alias for string interpolation, but a package can't necessarily wait for an IPM enhancement to make that happen.

isc-tleavitt:
Oops - sorry @isc-pbarton didn't mean to tag you here.

isc-kiyer (approved):

src/cls/IPM/Storage/Module.cls
isc-kiyer: @isc-tleavitt  Note that using that value isn't helpful or newer versions of IRIS with NoPWS so should likely be provided when installing the module

PR #559, opened 2024-08-29
Semver 2.0 

### LINKED ISSUE #557: Exclude prereleases from SemVer range maximums
Per e.g. https://github.com/npm/node-semver?tab=readme-ov-file#x-ranges-12x-1x-12- / https://github.com/npm/node-semver/commit/059a5adec5aefaa764bf4fdb717ae8b42cbbefcd - we need to exclude prereleases from range maximums as a rule. The simple way to do this is representing a range maximum with a prerelease of 0 (e.g., 1.x becomes >=1.0.0 <2.0.0-0).

See discussion in https://github.com/intersystems-community/zpm-registry/issues/91

I'd recommend a test-driven development workflow for this, expanding on the existing SemVer unit tests. This is v0.9.0 critical.

isc-shuliu:
Fix #557

### DIFF
--- a/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
+++ b/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
@@ -93,7 +93,7 @@ Method Evaluate(pVersion As %IPM.General.SemanticVersion) As %Boolean
       Set tEquals = tEquals && (val1 = val2)
     }
   }
-	If (##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease") = 0) || ($Extract(..Operator,1) = "=")  {
+	If (+##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease") = 0) || ($Extract(..Operator,1) = "=")  {
 		// If SemVerPostRelease is disabled, then we should not consider post-release versions as equal to the base version
 		// ..Operator = "=" means direct equality and not <= or >= so compare pre-release as well
 		Set tEquals = tEquals && (pVersion.Prerelease = ..Prerelease)
--- a/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
+++ b/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
@@ -93,8 +93,9 @@ Method Evaluate(pVersion As %IPM.General.SemanticVersion) As %Boolean
       Set tEquals = tEquals && (val1 = val2)
     }
   }
-	If ($Extract(..Operator,1) = "=") {
-		// Means direct equality and not <= or >= so compare pre-release as well
+	If (##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease") = 0) || ($Extract(..Operator,1) = "=")  {
+		// If SemVerPostRelease is disabled, then we should not consider post-release versions as equal to the base version
+		// ..Operator = "=" means direct equality and not <= or >= so compare pre-release as well
 		Set tEquals = tEquals && (pVersion.Prerelease = ..Prerelease)
 	}
 	If tEquals || (..Operator = "=") {
--- a/tests/unit_tests/Test/PM/Unit/SemVer/Abstract.cls
+++ b/tests/unit_tests/Test/PM/Unit/SemVer/Abstract.cls
@@ -1,0 +1,20 @@
+Class Test.PM.Unit.SemVer.Abstract Extends %UnitTest.TestCase
+{
+
+/// Original value of the SemVerPostRelease setting
+Property OriginalSemVerPostRelease As %Boolean [ InitialExpression = {##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease")} ];
+
+/// Before each test, record the original value of the SemVerPostRelease setting and set it to 0 for a clean slate
+Method OnBeforeOneTest(testname As %String) As %Status
+{
+	Set ..OriginalSemVerPostRelease = ##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease")
+	Return ##class(%IPM.Repo.UniversalSettings).SetValue("SemVerPostRelease", 0)
+}
+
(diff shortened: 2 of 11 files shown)


isc-tleavitt (approved):

tests/unit_tests/Test/PM/Unit/SemVer/Expressions.cls
```diff
@@ -34,13 +34,25 @@ Method TestPrerelease()
 	Do ..AssertSatisfied(">=1.0.0-beta.1","1.0.0-beta.1")
 	Do ..AssertSatisfied(">=1.0.0-beta.1","1.0.0-beta.2")
 	Do ..AssertSatisfied(">=1.0.0-alpha.1","1.0.0-beta.1")
-	Do ..AssertSatisfied("^1.0.0","1.0.0-1.m1")
+	Do ..AssertNotSatisfied("^1.0.0","1.0.0-1.m1") // This shouldn't be satisfied because 1.0.0-1.m1 is a prerelease
```
isc-kiyer, line 37: @isc-tleavitt @isc-shuliu  This may cause us some problems with maintenance releases where we currently rely on this behavior. If we don't want to support this, our release tool will need to lock down versions in not just the base module but every dependency.
For example, HealthConnect 7.0.0 lists HS.REST 1.1.0-1.m1 and HealthConnect.Common 4.1.1-1.m1 as fixed version dependencies. However, HealthConnect.Common lists HS.REST ^1.0.0 as a dependency. This would now break unless HealthConnect.Common also lists a specific version as a dependency.
We need the ability to toggle this behavior on vs off I think for HS. Either that or come up with an alternate solution for maintenance releases.
cc @isc-eneil

isc-tleavitt: @isc-kiyer that's poor use of semantic versioning to say that a prerelease is sequentially after a main release. If there's an alternate solution for maintenance releases that would be preferable.

isc-tleavitt: Specifically, I think this comes down to discipline around saying what's really a bugfix vs. enhancement, and making decisions about releases of individual packages based on their contents. Under what circumstances would we not want a top-level release to include fixes to known bugs in a dependency under a given minor release?

isc-tleavitt: All that said, we could address this with a top-level command and configurable "include prerelease" switch similar to includePrerelease in node-semver.

isc-tleavitt: So question is, among the following options, where does an "include prerelease" switch belong?
* module.xml, as a subelement of the module (or could *perhaps* jump through hoops in XSLTs to support via an attribute of Dependencies) - use case: always make it explicit for top-level packages in our package-lock-ish use case for e.g. HealthConnect, HSBASE.
* option in "load" and "install" commands - use case: in a proper SemVer environment or possibly a CI workflow, pick up not-yet-released upcoming versions of dependencies for testing
* system-level setting - use case: development environments where we'll always want to be able to pick up prereleases

My gut feeling is the system level setting, because it'll cover all cases and I want to do as little work as possible to enable poor use of SemVer.

isc-kiyer: I think a system level setting is a good idea. That being said, I would like to perhaps revisit just taking the next patch version (even if including other patches along the way) with the HS team to see if we can change maintenance releases to do that instead. Copying @bourette

bourette: I raised this with Jaideep, perhaps we could do the same with I4H/HC.

src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
```diff
@@ -93,8 +93,9 @@ Method Evaluate(pVersion As %IPM.General.SemanticVersion) As %Boolean
       Set tEquals = tEquals && (val1 = val2)
     }
   }
-	If ($Extract(..Operator,1) = "=") {
-		// Means direct equality and not <= or >= so compare pre-release as well
+	If (##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease") = 0) || ($Extract(..Operator,1) = "=")  {
```
isc-tleavitt, line 96: I would recommend coercing this to be a number with:
`If (+##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease") = 0) || ($Extract(..Operator,1) = "=")  {`

That way null is 0 (the default)

tests/unit_tests/Test/PM/Unit/SemVer/Expressions.cls
bourette: Just for completeness, we moved from + to - when we realized that re-spinning a release caused it to pick up newer code, making it impossible to reproduce a kit.

tests/unit_tests/Test/PM/Unit/SemVer/Abstract.cls
```diff
@@ -0,0 +1,20 @@
+Class Test.PM.Unit.SemVer.Abstract Extends %UnitTest.TestCase
```
isc-tleavitt, line 1: Nicely done.

isc-tleavitt (changes requested):
One minor change requested (just in case the installer doesn't run).

isc-shuliu:
Actually, b9552ee should be part of #560 instead of this one. Will cherry pick there and revert on this branch.

isc-tleavitt:
> Actually, [b9552ee](https://github.com/intersystems/ipm/commit/b9552ee494415a3879236e9bcbed82ad58e3e0db) should be part of #560 instead of this one. Will cherry pick there and revert on this branch.

It really belongs on top of both of them; the two seem a bit tangled now. Will just wait until both are approved to merge.

isc-tleavitt (approved):

isc-kiyer (approved):

PR #560, opened 2024-08-30
Fix filter prerelease snapshots

### LINKED ISSUE #544: can we mark a version as a pre-release
It looks like there is some discussion about pre-releases but I don't have clarity on how to define this and or what its behavior does.
https://github.com/intersystems/ipm/issues/480

I would like to be able to mark a version as a pre-release.  I would hope that when a user is in ZPM and do

install -v _moduleName_

It would skip any of the pre-release versions.
I could still do

install -v _moduleName_ specificVersion

to install a specific version including a pre-release version.

If I do 
install -v _moduleName_ specificVersion and its a pre-released version and there are module dependencies it would pick any pre-release versions for the dependent modules.

The reason why I want to do this is I currently have
rclib-studio          1.5.0
                      Versions: 1.5.0, 1.4.0, 1.0.1
rclib-system          1.5.0
                      Versions: 1.5.0, 1.4.0, 1.0.2, 1.0.0
rclib-unittest        1.5.0
                      Versions: 1.5.0, 1.4.0, 1.0.2, 1.0.1
rclib-wellbase        1.5.0
                      Versions: 1.5.0, 1.4.0, 1.3.0, 1.2.0, 1.1.0, 1.0.1

1.4.0 is our GR Release and 1.5.0 is a module that is published but not expected to go to customer sites just yet.

if someone just does

install -v rclib-unittest 

I'd like it to be able to skip 1.5.0 and install the highest non pre-release version, that being 1.4.0

isc-shuliu:
Fix #544 (must be used together with https://github.com/intersystems-community/zpm-registry/pull/112)

### DIFF
--- a/src/cls/IPM/Repo/Manager.cls
+++ b/src/cls/IPM/Repo/Manager.cls
@@ -113,8 +113,8 @@ Method SearchRepositoriesForModule(pSearchCriteria As %IPM.Repo.SearchCriteria, 
             Set patch = $Order(versions(major, minor, patch), -1)
             Quit:patch=""
             Set sub = ""
+            Set subDirection = $SELECT(+##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease"): -1, 1: 1)
             For  {
-			  Set subDirection = $SELECT(+##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease"): -1, 1: 1)
               Set sub = $Order(versions(major, minor, patch, sub), subDirection)
               Quit:sub=""
               Set repoOrder = ""
--- a/src/cls/IPM/Repo/Manager.cls
+++ b/src/cls/IPM/Repo/Manager.cls
@@ -101,6 +101,8 @@ Method SearchRepositoriesForModule(pSearchCriteria As %IPM.Repo.SearchCriteria, 
 			}
 		}
 	Set major = ""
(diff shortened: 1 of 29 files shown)


isc-shuliu:
@isc-tleavitt 
It won't break. The older zpm registry server will simply ignore these two parameters

isc-shuliu:
CIs should pass once https://github.com/intersystems-community/zpm-registry/pull/115 gets merged and released. In order to treat `1.0.0-1.m1` as a post-release, use should do `config set SemVerPostRelease 1` and set `includePrerelease` to 1 at the repo level.

isc-tleavitt (approved):

src/cls/IPM/Repo/Manager.cls
```diff
@@ -114,7 +114,8 @@ Method SearchRepositoriesForModule(pSearchCriteria As %IPM.Repo.SearchCriteria,
             Quit:patch=""
             Set sub = ""
             For  {
-              Set sub = $Order(versions(major, minor, patch, sub), 1)
+			  Set subDirection = $SELECT(+##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease"): -1, 1: 1)
```
isc-tleavitt, line 117: Nitpick: should call this outside loop.

isc-tleavitt (changes requested):

isc-tleavitt (approved):

isc-kiyer (approved):

PR #561, opened 2024-09-04
bootstrap: make a 0.7 release that honors prerelease & snapshots

isc-shuliu:
CI in #559 is failing because a newer version of zpm-registry has been published. The new zpm-registry needs `IncludePrerelease` and `IncludeSnapshots` in the URL query string to return prereleases and snapshots. Since older version (0.7.x) of zpm doesn't yet properly set them in URL, it won't install prerelease `0.9.0-dev.559.1`, hence causing the migration test from 0.7.x to fail.

This PR addresses this problem by setting `IncludePrerelease` and `IncludeSnapshots`. Another commit to change the CI in v1 branch is also needed (after this PR is merged and released)


isc-shuliu:
With https://github.com/intersystems-community/zpm-registry/pull/115 merged and published, this PR is no longer critical to unblock CIs for #559, but it's still nice to have so ZPM 0.7.x users can properly filter by includePrerelease and includeSnapshots.

PR #562, opened 2024-09-05
feat: implement generic webapp processor

isc-shuliu:
This PR adds a generic resource processor for web applications. Future proof. No hardcoded list of properties required.


src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+    Set tOriginalNS = $namespace
+    Set dbDir = $$$defdir
+    New $namespace
+    Set $namespace = "%SYS"
+    
+    Merge properties = ..SecurityProperties
+    If $D(properties("MatchRoles"), tMatchRoles) # 2 {
+        Set properties("MatchRoles") = ..ReplaceMatchRoles(tMatchRoles, dbDir)
+    }
+    If $D(properties("NameSpace"), tNameSpace) {
```
isc-tleavitt, line 145: Nit: should also have `#2`

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+Parameter DESCRIPTION As STRING = "Facilitates Web Application create/update/delete.";
+
+/// Name of the web application.
+Property Name As %String [ Calculated ];
+
+Method NameGet() As %String
+{
+    Set tName = $Get(..SecurityProperties("Name"), $Get(..SecurityProperties("Url"), ""))
+    If tName = "" {
+        Throw $$$ERROR($$$GeneralError, "Name is required for web applications")
```
isc-kiyer, line 16: This needs to be $$$ThrowStatus

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+
+/// Description of resource processor class (shown in UI)
+Parameter DESCRIPTION As STRING = "Facilitates Web Application create/update/delete.";
+
+/// Name of the web application.
+Property Name As %String [ Calculated ];
+
+Method NameGet() As %String
+{
+    Set tName = $Get(..SecurityProperties("Name"), $Get(..SecurityProperties("Url"), ""))
```
isc-kiyer, line 14: Nit: for net new code, no need to use the legacy `t` prefix

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+        }
+    } Catch e {
+        Set sc = e.AsStatus()
+    }
+    Quit sc
+}
+
+Method CreateOrUpdateWebApp(pVerbose As %Boolean = 0) [ Internal ]
+{
+    // TODO: Make mirror-safe.
```
isc-tleavitt, line 135: @isc-kiyer do you think that making this mirror-safe needs to happen as part of initial implementation? (Same for other places with the comment.)

@isc-shuliu for context, this would mean using https://docs.intersystems.com/irisforhealthlatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=HSLIB&CLASSNAME=HS.HC.SystemConfig.API if available. (This was renamed in I4H 2023.1; previously it was https://docs.intersystems.com/irisforhealth20221/csp/documatic/%25CSP.Documatic.cls?LIBRARY=HSLIB&CLASSNAME=HS.Util.SystemAPI. Not sure how far back we'd need to go.)

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+    /*
+     * NOTE: Web app creation is done in OnBeforePhase so it happens before other
+     * resource processors such as FileCopy run.
+     */
+    Set sc = $$$OK
+    Try {
+        $$$ThrowOnError(##super(pPhase,.pParams))
+        
+        Set tVerbose = $Get(pParams("Verbose"))
+        #; Set tIsApplication = ..ResourceReference.Module.Lifecycle.%IsA(##class(%ZHSLIB.PackageManager.Developer.Lifecycle.Application).%ClassName(1))
```
isc-kiyer, line 115: Can remove the tIsApplication check here and below along with the note about making it mirror safe. Mirror safety is the responsibility of the platform and IRIS will get there eventually with the new security database (coming in IRIS 2024.3 and will be mirror aware in 25.1 hopefully)

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+    
+    Merge properties = ..SecurityProperties
+    If $D(properties("MatchRoles"), tMatchRoles) # 2 {
+        Set properties("MatchRoles") = ..ReplaceMatchRoles(tMatchRoles, dbDir)
+    }
+    If $D(properties("NameSpace"), tNameSpace) {
+        Set properties("NameSpace") = ..ReplaceNameSpace(tNameSpace, tOriginalNS)
+    } Else {
+        // Does WSGI application need a namespace?
+        Set properties("NameSpace") = tOriginalNS
```
isc-kiyer, line 149: Should only set a namespace is one is provided.

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+    Set dbDir = $$$defdir
+    New $namespace
+    Set $namespace = "%SYS"
+    
+    Merge properties = ..SecurityProperties
+    If $D(properties("MatchRoles"), tMatchRoles) # 2 {
+        Set properties("MatchRoles") = ..ReplaceMatchRoles(tMatchRoles, dbDir)
+    }
+    If $D(properties("NameSpace"), tNameSpace) {
+        Set properties("NameSpace") = ..ReplaceNameSpace(tNameSpace, tOriginalNS)
```
isc-kiyer, line 146: No need for this since %Evaluate() handles it.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-tleavitt, line 115: Any idea what's up with this? Vestigial code?

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+        Set eval = ..%Evaluate(value)
+        Set ..SecurityProperties(key) = eval
+    }
+}
+
+Method ReplaceMatchRoles(matchRoles As %String, dbDir As %String) As %String
+{
+    New $NAMESPACE 
+    Set $NAMESPACE = "%SYS"
+    Set templates = $Listbuild("{$dbrole}", "${dbrole}")
```
isc-kiyer, line 46: Would add this to the generic %EvaluateSystemExpression() in %IPM.Utils.Module instead of being resource specific.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-tleavitt: (if so should just delete)

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+Method CreateOrUpdateWebApp(pVerbose As %Boolean = 0) [ Internal ]
+{
+    // TODO: Make mirror-safe.
+    Set tOriginalNS = $namespace
+    Set dbDir = $$$defdir
+    New $namespace
+    Set $namespace = "%SYS"
+    
+    Merge properties = ..SecurityProperties
+    If $D(properties("MatchRoles"), tMatchRoles) # 2 {
```
isc-kiyer, line 142: Nit: Use full name for functions/expressions

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,263 @@
...
+
+/// This removes an existing CSP application
+Method DeleteWebApp(pVerbose As %Boolean = 0) [ Internal ]
+{
+    // TODO: Make mirror-safe.
+    
+    // Only try to purge files if the directory exists.
+    Set sc = ##class(%Library.EnsembleMgr).deletePortal(..Name,"",pVerbose)
+    Set errorCodes = $System.Status.GetErrorCodes(sc)
+    If (errorCodes [ $$$ApplicationDoesNotExist || errorCodes [ $$$DeleteObjectNotFound) {
```
isc-tleavitt, line 210: ObjectScript operator precedence: this doesn't do what you think it does. Change to:
`If ((errorCodes [ $$$ApplicationDoesNotExist) || (errorCodes [ $$$DeleteObjectNotFound))`

isc-tleavitt (changes requested):

tests/integration_tests/Test/PM/Integration/_data/rest-app/module.xml
```diff
@@ -0,0 +1,35 @@
...
+      <Packaging>module</Packaging>
+      <SourcesRoot>src</SourcesRoot>
+
+      <WebApplication 
+        Name="/restdemo"
+        Path="/src"
+        Recurse="1"
+        Directory="{$cspdir}/restdemo" 
+        MatchRoles=":{$dbrole}"
+        PasswordAuthEnabled="1"
```
isc-kiyer, line 25: Wouldn't this fail? We only have AutheEnabled and provide it values. I would recommend adding a few more options for evaluating each of the $$$Authe* macros. e.g. ${AutheCache}, ${AutheDelegated} that get mapped to the corresponding macro before evaluating

isc-kiyer (changes requested):
@isc-shuliu thanks for getting this in! My comments are up

tests/integration_tests/Test/PM/Integration/_data/rest-app/module.xml
isc-shuliu: Just fixed it by setting `"AutheEnabled` dynamically based on `PasswordAuthEnabled`, `UnauthenticatedEnabled`, etc.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Fixed.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Deleted.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Done.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Done.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Removed.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Removed the whole if-clause and delegate the processing it to %Evaluate()

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Done!

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Done!

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-kiyer: @isc-tleavitt @isc-shuliu  Nope do not bother making it mirror safe. The IRIS security database will eventually support mirroring (hopefully by next year)

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,230 @@
...
+Method CreateOrUpdateWebApp(pVerbose As %Boolean = 0) [ Internal ]
+{
+    // TODO: Make mirror-safe.
+    New $namespace
+    Set $namespace = "%SYS"
+
+    Merge properties = ..SecurityProperties
+    If $Get(properties("AutheEnabled")) = "" {
+        Set properties("AutheEnabled") = (
+            ($Get(properties("PasswordAuthEnabled")) * $$$AutheCache) +
```
isc-kiyer, line 107: I don't think we should have these sort of flags at all since it will go stale if a new auth type is introduced (for example, oauth will be a new type in IRIS 2025.1). Best to just pass through to Security.Applications APIs but allow for expressions to make it less abstract/simpler instead of having the numbers directly.

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,230 @@
...
+        }
+        $$$ThrowOnError(sc)
+        Write:pVerbose !,"Done."
+    }
+}
+
+/// This removes an existing CSP application
+Method DeleteWebApp(pVerbose As %Boolean = 0) [ Internal ]
+{
+    // TODO: Make mirror-safe.
```
isc-kiyer, line 172: Nit: remove comment about mirror safety.

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,230 @@
...
+    
+    // Only try to purge files if the directory exists.
+    Set sc = ##class(%Library.EnsembleMgr).deletePortal(..Name,"",pVerbose)
```
isc-kiyer, line 175: I think best to use the Security.Applications APIs directly instead.

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,230 @@
...
+
+Method GetSourceControlInfo(Output pInfo As %IPM.ExtensionBase.SourceControl.ResourceInfo) As %Status
+{
+    Set pInfo = ##class(%IPM.ExtensionBase.SourceControl.ResourceInfo).%New()
+    If (..Name '= "") {
+        Set pInfo.SourceControlAware = 1
+        Set pInfo.ResourceType = "/CSP/"
+        Set pInfo.Prefix = ..Name
+        Set pInfo.RelativePath = ..ResourceReference.Name
+        Set pInfo.IsDirectory = ($Piece(..ResourceReference.Name,"/",*) '[ ".")
```
isc-kiyer, line 202: IsDirectory can be set to 1 since a web app is always a directory

src/cls/IPM/Utils/Module.cls
```diff
@@ -2297,6 +2301,17 @@ ClassMethod GetDatabaseInfoForNamespace(pNamespace As %String, Output pPropertie
...
 
+ClassMethod GetDatabaseRole()
+{
+	Set dbDir = $$$defdir
+
+	New $NAMESPACE
+	Set $NAMESPACE = "%SYS"
+
+	Set db = ##class(SYS.Database).%OpenId(dbDir)
+	Return $SELECT($IsObject(db): db.ResourceName, 1: "%DB_DEFAULT")
```
isc-kiyer, line 2312: Can get db role using ##class(%SYS.Namespace).GetNSInfo() instead which won't need a namespace switch.

tests/integration_tests/Test/PM/Integration/_data/rest-app/module.xml
```diff
@@ -0,0 +1,37 @@
...
+      <SourcesRoot>src</SourcesRoot>
+
+      <Resource Name="Test.Rest.Demo.CLS"/>
+      <WebApplication 
+        Name="/restdemo"
+        NameSpace="{$namespace}"
+        Path="/src"
+        Recurse="1"
+        Directory="{$cspdir}/restdemo" 
+        MatchRoles=":{$dbrole}"
```
isc-kiyer, line 26: Would include some more roles in the test

isc-kiyer (changes requested):
@isc-shuliu few more small comments

isc-shuliu:
The CIs are failing with a `<protect>` error except for I4H 2022.1, probably because of `{$dbrole}` is evaluated to `:%DB_USER-VERIFY-DATA` rather than `:%DB_USER-VERIFY`. 

Only I4H 2022.1 evaluates this to `:%DB_USER-VERIFY` hence it passes.

I tried both
```objectscript
Do ##class(%SYS.Namespace).GetNSInfo($NAMESPACE, .nsInfo)
Return $Get(nsInfo("Resource"), "%DB_DEFAULT")
```
and 
```objectscript
Set db = ##class(SYS.Database).%OpenId(dbDir)
Return $SELECT($IsObject(db): db.ResourceName, 1: "%DB_DEFAULT")
```
but neither seems to be giving the correct result, except on I4H 2022.1.

@isc-kiyer @isc-tleavitt Any clue why this is the case?

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-tleavitt: Oooooh beautiful! That's news to me but great to hear.

isc-tleavitt:
> The CIs are failing with a `<protect>` error except for I4H 2022.1, probably because of `{$dbrole}` is evaluated to `:%DB_USER-VERIFY-DATA` rather than `:%DB_USER-VERIFY`.

The error is:
`ERROR #5002: ObjectScript error: <PROTECT>^%CSP.REST.1 ^|^^\/usr\/irissys\/mgr\/USER-VERIFY-CODE\/|Test.Rest.Demo.1`

That is, there is a lack of read privilege on the *routine* database. Our CI setup for verify creates separate routine and data DBs for the temporary namespace used.

This is actually a common/annoying problem - see e.g. https://github.com/intersystems/isc-perf-ui/blob/main/cls/pkg/isc/perf/ui/Installer.cls#L11 which is called from an invoke at https://github.com/intersystems/isc-perf-ui/blob/main/module.xml#L68. From a defense in depth perspective, we generally would only want to grant read privilege on the routine database, but such a role does not inherently exist (we need to create it as this automation does).

IMO the right solution is to bake the automation from this installer into IPM itself and make it easier to use/manage.

isc-shuliu:
@isc-tleavitt Point of clarification: do we want to call this method on all web applications after they created, or only on those that allow unauthenticated access?

isc-tleavitt:
> @isc-tleavitt Point of clarification: do we want to call this method on all web applications after they created, or only on those that allow unauthenticated access?

Great question - I think we should make it easy to opt in to it. Let's say that similar to ${dbRole} we have ${codeReadRole} and we automatically manage an appropriate role (derived from the package name and namespace name to ensure uniqueness system-wide) - that is, create if referenced, and on uninstallation delete.

Given that we need to do this with duplicated boilerplate code for every web application using isc.ipm.js or similar less-advanced constructs, I'm a big fan of automating it.

@isc-kiyer any concerns/objections here?

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: I am okay with using comma-separated indices.

isc-tleavitt: @isc-shuliu I agree with @isc-kiyer that we should add expression support for this. To be more specific, this might look like:
${authPassword} -> $$$AutheCache
${authUnauthenticated} -> $$$AutheUnauthenticated
${authDelegated} -> $$$AutheDelegated
etc.

That works well if you support a single mode; not sure if this would work, but would be interesting to see if it does (and perhaps we should make it so):
`AutheEnabled="#{${authDelegated} + ${authPassword}}"`

If a new mode is released that IPM doesn't yet support, a user could add the relevant integer without the expression and add a comment about it.

isc-shuliu: @isc-tleavitt How does this commit https://github.com/intersystems/ipm/pull/562/commits/1c3892153759780fde94b62b64724e6a327a3a51 look to you? 

Basically we substitute ${authPassword} in the same way we handle ${namespace}, ${installDir}, etc., and separate the string by "+", and finally add the numbers up. 

Comments are also made possible because "32 abc" is treated as 32 during addition in ObjectScript

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -88,32 +88,46 @@ Method OnBeforePhase(pPhase As %String, ByRef pParams) As %Status
         } ElseIf ((pPhase = "Unconfigure") || (pPhase = "Clean")) {
             // Remove web application
             Do ..DeleteWebApp(tVerbose)
+            Do ..DeleteCodeReadRole($Get(..SecurityProperties("NameSpace")), tVerbose)
```
isc-tleavitt, line 91: Possible problem with this:
There could be multiple packages in the same namespace that all create this role.

Two possible solutions:
* Change the role name to incorporate the package name
* Check that there are no remaining web applications (users, etc.) referencing the role prior to deleting it

If somehow Security.Roles has built-in checks to avoid deletion of roles that are referenced (would be great but I doubt it), then the code is fine as-is.

tests/integration_tests/Test/PM/Integration/_data/rest-app/module.xml
```diff
@@ -23,9 +23,8 @@
         Path="/src"
         Recurse="1"
         Directory="{$cspdir}/restdemo" 
-        MatchRoles=":{$dbrole}"
-        PasswordAuthEnabled="1"
-        UnauthenticatedEnabled="1"
+        MatchRoles=":${dbrole}:%SQL:${codeReadRole}:%Developer,%All:%Manager"
```
isc-tleavitt, line 26: Clarifying question: what is the format of MatchRoles? What does this end up configuring in the web application?
I'd expect it to be <user role>:<grant role>,<user role>:<grant role>,... but maybe I'm wrong here.

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -227,4 +236,105 @@ ClassMethod DoPropertiesExist(pPropList As %Library.List) As %Library.List [ Int
...
+/// Grant read permissions to the routine database. This is necessary because 
+/// Adapted from https://github.com/intersystems/isc-perf-ui/blob/main/cls/pkg/isc/perf/ui/Installer.cls
+ClassMethod CreateCodeReadRole(pNamespace As %String, pVerbose As %Boolean) As %String [ Internal, Private ]
+{
+    New $Namespace
+    Set $Namespace = "%SYS"
+
+    set dbResource = ..GetDbResource(pNamespace)
+    If pVerbose {
+        Write !, "DB Resoure: ", dbResource
```
isc-tleavitt, line 280: typo: resource

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -227,4 +236,105 @@ ClassMethod DoPropertiesExist(pPropList As %Library.List) As %Library.List [ Int
...
+            Write !, "Public Read permission already exists on ", dbResource, !, "Skipping"
+        }
+        Return ""
+    }
+
+    // Create role granting R permission on namespace-default routine DB resource
+    // This is the least-access means to be able to use the REST handler class
+    // in unauthenticated mode
+    Set roleName = ..ConstructRoleName(dbResource)
+    Set roleProps("Description") = "Role created by %IPM that grants only R permission on "_dbResource
```
isc-tleavitt, line 297: Say "Role created by IPM" instead of "%IPM"

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -227,4 +236,105 @@ ClassMethod DoPropertiesExist(pPropList As %Library.List) As %Library.List [ Int
...
+}
+
+ClassMethod DeleteCodeReadRole(pNamespace As %String, pVerbose As %Boolean) [ Internal, Private ]
+{
+    New $Namespace
+    Set $Namespace = "%SYS"
+
+    Set dbResource = ..GetDbResource(pNamespace)
+    If pVerbose {
+        Write !, "DB Resoure: ", dbResource
```
isc-tleavitt, line 323: Typo (again): resource

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -88,32 +88,46 @@ Method OnBeforePhase(pPhase As %String, ByRef pParams) As %Status
...
             Do ..DeleteWebApp(tVerbose)
+            Do ..DeleteCodeReadRole($Get(..SecurityProperties("NameSpace")), tVerbose)
         }
     } Catch e {
         Set sc = e.AsStatus()
     }
     Quit sc
 }
 
+Method ComputeAutheNumber(pAuth As %String) As %String
```
isc-tleavitt, line 99: This shouldn't be necessary with the #{} syntax, but I think it is because of a bug in %IPM.Storage.Module that prevents nesting of system expressions within an ObjectScript expression:

```
	Set tAttrValue = ##class(%IPM.Storage.ModuleSetting.Default).EvaluateAttribute(tAttrValue,.customParams)
	Set attrValue = ##class(%IPM.Utils.Module).%EvaluateSystemExpression(tAttrValue)
	
	// {name} expression does not contain dollar sign; handle separately
	Set name = ..Name
	Set attrValue = ##class(%Regex.Matcher).%New("(?i)\{name\}", attrValue).ReplaceAll($Replace(name,"\","\\"))

	Set regex = ##class(%Regex.Matcher).%New("#\{([^}]+)\}", tAttrValue)
	While regex.Locate() {
		Set expr = regex.Group(1)
		Set value = ..%EvaluateExpression(expr)
		Set $Extract(attrValue, regex.Start, regex.End - 1) = value
		Set regex.Text = attrValue
	}
```

It should be instead:
```
Set regex = ##class(%Regex.Matcher).%New("#\{([^}]+)\}", attrValue)
```

Then I think this method would be unneeded.

isc-tleavitt: (this is valid to fix in your PR)

isc-tleavitt (changes requested):
Some changes needed + some questions.

tests/integration_tests/Test/PM/Integration/_data/rest-app/module.xml
isc-shuliu: So it's comma separated pieces, each piece is colon separated in the form of `MatchRole:TargetRole1:TargetRole2:TargetRole3`. 

- If MatchRole is nonempty, the target roles will always be added as long as `MatchRole` is detected. This configures the `Match Roles` tab in web portal
- If MatchRole is empty, the target roles will always be assigned. This configures the `Application Roles` tab in web portal.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: It appears that delete a role will strip that role from all applications. Will fix that by incorporating the package name. After all, this role should be created by the package.

isc-kiyer:
> > The CIs are failing with a `<protect>` error except for I4H 2022.1, probably because of `{$dbrole}` is evaluated to `:%DB_USER-VERIFY-DATA` rather than `:%DB_USER-VERIFY`.
> 
> The error is: `ERROR #5002: ObjectScript error: <PROTECT>^%CSP.REST.1 ^|^^\/usr\/irissys\/mgr\/USER-VERIFY-CODE\/|Test.Rest.Demo.1`
> 
> That is, there is a lack of read privilege on the _routine_ database. Our CI setup for verify creates separate routine and data DBs for the temporary namespace used.
> 
> This is actually a common/annoying problem - see e.g. https://github.com/intersystems/isc-perf-ui/blob/main/cls/pkg/isc/perf/ui/Installer.cls#L11 which is called from an invoke at https://github.com/intersystems/isc-perf-ui/blob/main/module.xml#L68. From a defense in depth perspective, we generally would only want to grant read privilege on the routine database, but such a role does not inherently exist (we need to create it as this automation does).
> 
> IMO the right solution is to bake the automation from this installer into IPM itself and make it easier to use/manage.

@isc-tleavitt @isc-shuliu for this, why not look up the role for the routine database? It does mean we can no longer use the %SYS.Namespace call and would need to use Config.Databases but that's fine. I recommend two expressions: $routinesDbRole and $globalsDbRole to make available for evaluation. For web apps, would need routines at a minimum but may also want to grant access to globals (so the user role doesn't need access to the database. This is what we do in HS for example).

isc-tleavitt:
> > > The CIs are failing with a `<protect>` error except for I4H 2022.1, probably because of `{$dbrole}` is evaluated to `:%DB_USER-VERIFY-DATA` rather than `:%DB_USER-VERIFY`.
> > 
> > 
> > The error is: `ERROR #5002: ObjectScript error: <PROTECT>^%CSP.REST.1 ^|^^\/usr\/irissys\/mgr\/USER-VERIFY-CODE\/|Test.Rest.Demo.1`
> > That is, there is a lack of read privilege on the _routine_ database. Our CI setup for verify creates separate routine and data DBs for the temporary namespace used.
> > This is actually a common/annoying problem - see e.g. https://github.com/intersystems/isc-perf-ui/blob/main/cls/pkg/isc/perf/ui/Installer.cls#L11 which is called from an invoke at https://github.com/intersystems/isc-perf-ui/blob/main/module.xml#L68. From a defense in depth perspective, we generally would only want to grant read privilege on the routine database, but such a role does not inherently exist (we need to create it as this automation does).
> > IMO the right solution is to bake the automation from this installer into IPM itself and make it easier to use/manage.
> 
> @isc-tleavitt @isc-shuliu for this, why not look up the role for the routine database? It does mean we can no longer use the %SYS.Namespace call and would need to use Config.Databases but that's fine. I recommend two expressions: $routinesDbRole and $globalsDbRole to make available for evaluation. For web apps, would need routines at a minimum but may also want to grant access to globals (so the user role doesn't need access to the database. This is what we do in HS for example).

The role for the routine database would include write permission. Principle of least privilege means that we shouldn't grant the ability to modify code as an application role; we need a separate role that grants only read permission. That's what Shuheng has implemented based on the sample code I've copied to half a dozen projects that need to do this.

isc-kiyer:
> > > > The CIs are failing with a `<protect>` error except for I4H 2022.1, probably because of `{$dbrole}` is evaluated to `:%DB_USER-VERIFY-DATA` rather than `:%DB_USER-VERIFY`.
> > > 
> > > 
> > > The error is: `ERROR #5002: ObjectScript error: <PROTECT>^%CSP.REST.1 ^|^^\/usr\/irissys\/mgr\/USER-VERIFY-CODE\/|Test.Rest.Demo.1`
> > > That is, there is a lack of read privilege on the _routine_ database. Our CI setup for verify creates separate routine and data DBs for the temporary namespace used.
> > > This is actually a common/annoying problem - see e.g. https://github.com/intersystems/isc-perf-ui/blob/main/cls/pkg/isc/perf/ui/Installer.cls#L11 which is called from an invoke at https://github.com/intersystems/isc-perf-ui/blob/main/module.xml#L68. From a defense in depth perspective, we generally would only want to grant read privilege on the routine database, but such a role does not inherently exist (we need to create it as this automation does).
> > > IMO the right solution is to bake the automation from this installer into IPM itself and make it easier to use/manage.
> > 
> > 
> > @isc-tleavitt @isc-shuliu for this, why not look up the role for the routine database? It does mean we can no longer use the %SYS.Namespace call and would need to use Config.Databases but that's fine. I recommend two expressions: $routinesDbRole and $globalsDbRole to make available for evaluation. For web apps, would need routines at a minimum but may also want to grant access to globals (so the user role doesn't need access to the database. This is what we do in HS for example).
> 
> @isc-kiyer the role for the routine database would include write permission. Principle of least privilege means that we shouldn't grant the ability to modify code as an application role; we need a separate role that grants only read permission. That's what Shuheng has implemented based on the sample code I've copied to half a dozen projects that need to do this.

Ah ok that makes sense. For HS, our code databases are RO so although not ideal, granting write permission is fine.

isc-tleavitt (approved):

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,320 @@
...
+    // Find the namespace-default routine DB
+    $$$ThrowOnError(##class(Config.Namespaces).Get(pNamespace,.nsProps))
+    Set defaultDB = nsProps("Routines")
+    $$$ThrowOnError(##class(Config.Databases).Get(defaultDB,.dbProps))
+    Set db=##Class(SYS.Database).%OpenId(dbProps("Directory"),,.sc)
+    $$$ThrowOnError(sc)
+    Return db.ResourceName
+}
+
+/// Grant read permissions to the routine database. This is necessary because
```
isc-kiyer, line 251: Nit: Comment seems incomplete

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,320 @@
...
+}
+
+Method ConstructRoleName(pDbResource As %String) As %String [ Internal, Private ]
+{
+    // Include the module name in the role name to avoid conflicts
+    // This role should only be managed by IPM itself
+    Return $ListToString($lb("IPM", "readonly", ..ResourceReference.Module.Name, pDbResource), ".")
+}
+
+Method ReplaceCodeReadRole(pMatchRoles As %String, pNamespace As %String, pVerbose As %Boolean) [ Internal, Private ]
```
isc-kiyer, line 224: Nit: Add return type of %String

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,320 @@
...
+        }
+    }
+    Return badList
+}
+
+Method ConstructRoleName(pDbResource As %String) As %String [ Internal, Private ]
+{
+    // Include the module name in the role name to avoid conflicts
+    // This role should only be managed by IPM itself
+    Return $ListToString($lb("IPM", "readonly", ..ResourceReference.Module.Name, pDbResource), ".")
```
isc-kiyer, line 221: Max length of a role name is 64 characters so this could definitely go over the max length with long module names. For example, in HS we have a module named HealthShare.HP.OAuth2.ResourceServer so the role name "IPM.readonly.HealthShare.HP.OAuth2.ResourceServer.%DB_HSREGISTRY" has length 64 already which is the maxlen so a db/namespace name that is slightly longer will push it over the edge

isc-kiyer (commented):
@isc-shuliu Sorry couldn't get to another round of review earlier in the week but there are some actionable items I would like to see resolved before 0.9.0 is published if you can tackle them in a follow up PR
cc @isc-tleavitt

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -0,0 +1,320 @@
...
+Method CreateOrUpdateWebApp(pVerbose As %Boolean = 0) [ Internal ]
+{
+    New $namespace
+    Set $namespace = "%SYS"
+
+    Merge properties = ..SecurityProperties
+
+    // Replace ${codeReadRole} with a role (potentially to be created) with read permission to routine database
+    If ($Data(properties("NameSpace"), ns) # 2) && ($Data(properties("MatchRoles"), mr) # 2) {
+        Set properties("MatchRoles") = ..ReplaceCodeReadRole(mr, ns, pVerbose)
```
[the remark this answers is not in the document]
isc-shuliu: @isc-kiyer 
I was thinking this isn't a simple string substitution because of the side-effect of creating a role. We also need to remove the role later if the package is uninstalled.

Issue #530, opened 2024-07-30
Permit phases for <Invoke> not implemented in the lifecycle class


isc-tleavitt:
Use case:
I want to add a LoadData phase that I can call via IPM - basically the equivalent of an npm script - that just does something via an `<Invoke>`. I shouldn't need to create a whole lifecycle class for this, I should just be able to say, for example:
`<Invoke Phase="LoadData" Class="LSHPP.Installer" Method="LoadData" />`

Right now this fails with <METHOD DOES NOT EXIST> because %Loaddata isn't defined in the Module lifecycle class. We should allow it instead and run the invoke.

@isc-kiyer thoughts on this?

isc-kiyer:
@isc-tleavitt  So when would such an invoke run if it is not during a defined phase? What benefit does this give the user vs running before/after the Activate phase?

isc-tleavitt:
@isc-kiyer it would be run manually - e..g:
zpm "isc.ls.personalization loadData"

In this case it's something that will need to be done on a recurring basis but not necessarily automated as part of (re)installation.

isc-tleavitt:
This is sort of similar to https://docs.npmjs.com/cli/v6/using-npm/scripts

isc-kiyer:
@isc-tleavitt ah ok yes I agree this would be useful then. Perhaps though as a different attribute (maybe CustomPhase) to clearly indicate it is not an OOTB phase?

isc-shuliu:
Fixed by #570

PR #570, opened 2024-09-17
V1next feat permit non lifecycle phase

isc-shuliu:
Implement #530

Added a `CustomPhase` attribute to `<Invoke>` that doesn't require a corresponding %method in lifecycle class.

### DIFF
--- a/src/cls/IPM/Storage/InvokeReference.cls
+++ b/src/cls/IPM/Storage/InvokeReference.cls
@@ -11,6 +11,9 @@ Property Method As %String(MAXLEN = 255, XMLPROJECTION = "ATTRIBUTE") [ Required
 
 Property Phase As %String(MAXLEN = 255, XMLPROJECTION = "ATTRIBUTE") [ InitialExpression = "Configure" ];
 
+/// If provided, overrides the Phase property will be ignored. This CustomPhase will be used and no cooresponding lifecycle is required.
+Property CustomPhase As %String(MAXLEN = 255, XMLPROJECTION = "ATTRIBUTE");
+
 Property When As %String(MAXLEN = 255, VALUELIST = ",Before,After", XMLPROJECTION = "ATTRIBUTE") [ InitialExpression = "After", SqlFieldName = _WHEN ];
 
 Property CheckStatus As %Boolean(XMLPROJECTION = "ATTRIBUTE") [ InitialExpression = 0 ];
@@ -44,7 +47,8 @@ Method GetArgsArray(pParams, Output args) As %Status
 
 Method OnBeforePhase(pPhase As %String, ByRef pParams) As %Status
 {
-  If (pPhase '= ..Phase) || ("Before" '= ..When) {
+  Set tPhase = $Select(..CustomPhase'="":..CustomPhase,1:..Phase)
+  If (pPhase '= tPhase) || ("Before" '= ..When) {
 	  Quit $$$OK
   }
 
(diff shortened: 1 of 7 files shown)


isc-tleavitt (approved):

src/cls/IPM/Storage/InvokeReference.cls
```diff
@@ -11,6 +11,9 @@ Property Method As %String(MAXLEN = 255, XMLPROJECTION = "ATTRIBUTE") [ Required
 
 Property Phase As %String(MAXLEN = 255, XMLPROJECTION = "ATTRIBUTE") [ InitialExpression = "Configure" ];
 
+/// If provided, overrides the Phase property will be ignored. This CustomPhase will be used and no cooresponding lifecycle is required.
```
isc-kiyer, line 14: Nit: Change comment to remove the word "override" and typo for corresponding

src/cls/IPM/Storage/Module.cls
```diff
@@ -273,7 +300,9 @@ ClassMethod ExecutePhases(pModuleName As %String, pPhases As %List, pIsComplete
 		Set tLifecycle.PhaseList = tPhases
 		Set tPointer = 0
 		Set tFullStart = $zh
+        Set tCustomPhases = tModule.GetCustomPhases()
```
isc-kiyer, line 303: Nit: Fix spacing

tests/integration_tests/Test/PM/Integration/CustomPhase.cls
```diff
@@ -0,0 +1,17 @@
...
+{
+
+Parameter TargetModuleName As STRING = "custom-phase-without-lifecycle";
+
+Method TestCustomPhaseWithoutLifecycle()
+{
+    Set tModuleDir = ..GetModuleDir(..#TargetModuleName)
+    Set tSC = ##class(%IPM.Main).Shell("load -verbose " _ tModuleDir)
+    Do $$$AssertStatusOK(tSC,"Loaded module successfully")
+    Set tSC = ##class(%IPM.Main).Shell(..#TargetModuleName _ " greeting -only")
```
isc-kiyer, line 11: Nit: Would be good to capture the output of the call to confirm it wrote the correct message i.e. the phase actually ran.
Also, can add tests for before, after and during for the custom phase

tests/integration_tests/Test/PM/Integration/_data/custom-phase-without-lifecycle/src/CustomPhaseWithoutLifecycle/Manager.cls
```diff
@@ -0,0 +1,15 @@
+Class CustomPhaseWithoutLifecycle.Manager
+{
+
+ClassMethod SayHello() As %Status
+{
+    Write !, "Hello from CustomPhaseWithoutLifecycle.Manager"
+    Quit $$$OK
```
isc-kiyer, line 7: Either use CheckStatus true in the module or don't return status here

src/cls/IPM/Storage/Module.cls
```diff
@@ -273,7 +300,9 @@ ClassMethod ExecutePhases(pModuleName As %String, pPhases As %List, pIsComplete
...
 		While $ListNext(tPhases,tPointer,tOnePhase) {
+            Set tIsCustomPhase = $ListFind(tCustomPhases, tOnePhase)
```
isc-kiyer, line 305: Nit: Rather than converting to a list and doing $ListFind, can leave custom phases in a subscripted array. This list will never be long but seems odd to convert to a list just to use O(n) operation to find entries when a subscript check has better performance anyway.

isc-kiyer (changes requested):
@isc-shuliu few small comments but looks good

isc-kiyer (approved):

PR #572, opened 2024-09-18
V1 enhance webapp processor

### LINKED ISSUE #571: Enhance generic web application processor
Fix remaining comments in #562

isc-shuliu:
Fix #571 

Remaining items for discussion
- [x] Is there a way to dynamically convert the string "$$$Authe*" to its macro to avoid compilation error on legacy systems where some macros aren't available? 
  **Idea**: Support macro expansion as described in #575. 
  On systems where a macro is not defined (say $$$AuthePassword is unavailable before 2024), will be substituted as an empty string.
- [x] ${codeDbReadRole} is removed. If a custom `<Role>` processor is needed, will do it in another PR


src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -216,15 +216,23 @@ ClassMethod DoPropertiesExist(pPropList As %Library.List) As %Library.List [ Int
...
 Method ConstructRoleName(pDbResource As %String) As %String [ Internal, Private ]
 {
-    // Include the module name in the role name to avoid conflicts
+    // Include the module name in the role name to avoid conflicts, using hash because role name can't be longer than 64 characters
     // This role should only be managed by IPM itself
-    Return $ListToString($lb("IPM", "readonly", ..ResourceReference.Module.Name, pDbResource), ".")
+    Set hashBytes = $System.Encryption.MD5Hash(..ResourceReference.Module.Name) // 128 bits
+    Set hashHexString = $System.Encryption.Base32HexEncode(hashBytes, 1) // 26 characters, without '=' padding
+    // As long as pDbResource is shorter than 33 characters, the role name will be shorter than 64 characters
+    Set roleName = $ListToString($ListBuild("IPM", pDbResource, hashHexString), ".")
```
isc-kiyer, line 224: Nit: The resource could begin with %DB_ so this could look a bit odd. Also in general resource names are `-` or `_` separated so I think the following would be more appropriate:
```
$ListToString($ListBuild(pDbResource, "IPM", hashHexString), "_")
```

src/cls/IPM/ResourceProcessor/WebApplication.cls
```diff
@@ -216,15 +216,23 @@ ClassMethod DoPropertiesExist(pPropList As %Library.List) As %Library.List [ Int
 
 Method ConstructRoleName(pDbResource As %String) As %String [ Internal, Private ]
 {
-    // Include the module name in the role name to avoid conflicts
+    // Include the module name in the role name to avoid conflicts, using hash because role name can't be longer than 64 characters
     // This role should only be managed by IPM itself
-    Return $ListToString($lb("IPM", "readonly", ..ResourceReference.Module.Name, pDbResource), ".")
+    Set hashBytes = $System.Encryption.MD5Hash(..ResourceReference.Module.Name) // 128 bits
```
isc-kiyer, line 221: MD5 is not available when using FIPS mode so maybe not the best function to use? I recall a customer reporting this as being problematic in the past.
Is there a reason to not just create `%DB_<database>_READ` role? I think that's simpler and cleaner. Description can indicate it was created by IPM. Creating many duplicate roles doing the same thing for each module seems like overkill. 
The downside is the role can't be cleaned up but one additional (fairly useful) role is worth the simplicity and not creating excess config to manage IMO

@isc-shuliu @isc-tleavitt thoughts?

src/cls/IPM/Utils/Module.cls
```diff
@@ -2245,7 +2245,16 @@ ClassMethod GetFileLines(pFileName As %String, Output pOutput) As %Status [ Inte
 /// 	<li>{$binDir} - the instance's bin directory.</li>
 /// 	<li>{$libDir} - the instance's lib directory.</li>
 /// 	<li>{$webroot} - the instance's constructed url with host and port (e.g. http://123.45.678.90:52773/) </li>
-/// 	<li>{$dbrole} - the current namespace's database role</li>
+/// 	<li>{$dbrole} - (deprecated in favor of {$globalsDbRole}) the current namespace's globals database role</li>
+/// 	<li>{$globalsDbRole} - the current namespace's globals database role, functionally equivalent to {$dbrole} in legacy packages.</li>
+/// 	<li>{$codeDbReadRole} - [In WebApplication only] the current namespace's routine database role (will be created if needed)</li>
```
isc-kiyer, line 2250: I know it has side effects but still worth moving the role creation here. I don't think it should be for web apps only since a customer may use it for the web app and then want to use it in an invoke as well

src/cls/IPM/Utils/Module.cls
```diff
@@ -2288,15 +2297,17 @@ ClassMethod %EvaluateSystemExpression(pString As %String) As %String [ Internal
...
-	Set result = ..%RegExReplace(result, "authUnauthenticated", $$$AutheUnauthenticated)
-	Set result = ..%RegExReplace(result, "authLDAP", $$$AutheLDAP)
-	Set result = ..%RegExReplace(result, "authDelegated", $$$AutheDelegated)
-	Set result = ..%RegExReplace(result, "authLoginToken", $$$AutheLoginToken)
-	Set result = ..%RegExReplace(result, "authTwoFactorSMS", $$$AutheTwoFactorSMS)
-	Set result = ..%RegExReplace(result, "authTwoFactorPW", $$$AutheTwoFactorPW)
+	Set result = ..%RegExReplace(result, "globalsDbRole", dbRole)
+
+	// Authentication bits used by Security.Applications
+	Set result = ..%RegExReplace(result, "authK5API", 2**2)
```
isc-kiyer, line 2303: Open question: for new auth modes, will need to manually add new expressions. Is there a way to not have to update this each time a new auth mode is added?

isc-kiyer (changes requested):
@isc-shuliu few more comments and one question

src/cls/IPM/Utils/Module.cls
```diff
@@ -2288,15 +2297,17 @@ ClassMethod %EvaluateSystemExpression(pString As %String) As %String [ Internal
 
 	Set dbRole = ..GetDatabaseRole()
 	Set result = ..%RegExReplace(result, "dbrole", dbRole)
```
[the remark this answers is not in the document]
isc-kiyer: @isc-shuliu Ah ok. In that case, leave it for now but I think we should remove it in v1 to remove ambiguity (can remain in 0.9.0).

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-shuliu: Now that I think about it, if we want to automatically delete the role, the role should be listed as a standalone resource. So I agree we can simplify it and move the handling to %IPM.Utils.Module::%Evaluate() and create the role there.

@isc-tleavitt Does that make sense?

src/cls/IPM/Utils/Module.cls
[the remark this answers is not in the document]
isc-tleavitt: If it's only been used in the context of the old web application processor (not this newly-added one), which I believe is the case, then it wouldn't be a compatibility issue to change the expression name for the new one.

src/cls/IPM/ResourceProcessor/WebApplication.cls
isc-tleavitt: The reason for a package-specific role is so that a package can clean up after itself. But fair enough to create a reusable READ role by convention.

src/cls/IPM/Utils/Module.cls
isc-tleavitt: Could code-generate based on contents of %sySecurity.inc but that feels kind of gross.

isc-shuliu: @isc-tleavitt @isc-kiyer 
I'm proposing this:

We can do regex search for patterns like `$$$Authe*` in attributes, and try to evaluate them using the following method. This removes the need to hard-code a list of supported `${auth*}` templates. If the user's system supports the macro, we return its value; otherwise, return 0 (or maybe raise a warning/error)

```objectscript
ClassMethod EvaluateMacro(pMacro As %String) As %String
{
    Set stream = ##class(%Stream.TmpCharacter).%New()
    Do stream.WriteLine("Class Shuheng.Generated Extends %RegisteredObject")
    Do stream.WriteLine("{")
    Do stream.WriteLine("ClassMethod EvalMacro() [ CodeMode = objectgenerator ]")
    Do stream.WriteLine("{")
    Do stream.WriteLine("  Do %code.WriteLine(""#IfDef "_ pMacro _""")")
    Do stream.WriteLine("  Do %code.WriteLine("" Quit $$$"_pMacro_""")")
    Do stream.WriteLine("  Do %code.WriteLine(""#EndIf "")")
    Do stream.WriteLine("  Do %code.WriteLine("" Quit 0 "")")
    Do stream.WriteLine("  Quit $$$OK") 
    Do stream.WriteLine("}")
    Do stream.WriteLine("}")

    Do $system.OBJ.LoadStream(stream, "ck-d")
    Quit $ClassMethod("Shuheng.Generated", "EvalMacro")
    // Maybe delete the helper class after the macro is evaluated
}
```

isc-shuliu: The module.xml will look like
```xml
<WebApplication 
    AutheEnabled="#{$$$AutheCache + $$$AutheUnauthenticated}"
    ... 
/>
```

src/cls/IPM/Utils/Module.cls
```diff
@@ -2288,19 +2290,47 @@ ClassMethod %EvaluateSystemExpression(pString As %String) As %String [ Internal
...
-	Set result = ..%RegExReplace(result, "authTwoFactorSMS", $$$AutheTwoFactorSMS)
-	Set result = ..%RegExReplace(result, "authTwoFactorPW", $$$AutheTwoFactorPW)
+	Set result = ..%RegExReplace(result, "globalsDbRole", dbRole)
 
 	Return result
 }
 
+ClassMethod %EvaluateMacro(pString As %String) As %String [ Internal ]
+{
+	Set tString = pString
```
isc-kiyer, line 2300: For better performance, only do this if the string contains `$$$`. If it doesn't, just return pString as is

src/cls/IPM/Utils/Module.cls
```diff
@@ -2288,19 +2290,47 @@ ClassMethod %EvaluateSystemExpression(pString As %String) As %String [ Internal
...
+    Do stream.WriteLine("  Do %code.WriteLine(""#IfDef "_ pMacro _""")")
+    Do stream.WriteLine("  Do %code.WriteLine("" Quit $$$"_pMacro_""")")
+    Do stream.WriteLine("  Do %code.WriteLine(""#Else "")")
+    Do stream.WriteLine("  Do %code.WriteLine("" Quit """""""" "")")
+    Do stream.WriteLine("  Do %code.WriteLine(""#EndIf "")")
+    Do stream.WriteLine("  Quit $$$OK") 
+    Do stream.WriteLine("}")
+    Do stream.WriteLine("}")
+
+    Do $system.OBJ.LoadStream(stream, "ck-d")
```
isc-kiyer, line 2328: Check if there was a compilation error by checking the status here and if so, don't call the class method below

isc-kiyer (approved):
@isc-shuliu small comment but this looks good so approving (assuming it will be addressed)

isc-kiyer (approved):

Issue #501, opened 2024-06-10
Preserve module.xml format on disk upon import to .ZPM document stream


isc-tleavitt:
We need to preserve the format of module.xml on disk when it is imported to the .ZPM document stream, such that the .ZPM document can be edited server-side effectively with embedded source control.

PR #576, opened 2024-10-02
V1next internal xslt

isc-shuliu:
Implement #501

### DIFF
--- a/src/cls/IPM/StudioDocument/Module.cls
+++ b/src/cls/IPM/StudioDocument/Module.cls
@@ -55,7 +55,7 @@ ClassMethod UpdatePersistentFromStream(pStream As %IPM.StudioDocument.ModuleStre
 				Quit
 			}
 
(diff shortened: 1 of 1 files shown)


isc-shuliu:
Here's the current comparison between module.xml and <module-name>.ZPM.

image

isc-shuliu:
@isc-tleavitt Can you confirm this is what we want? Or do we need some more wrapper around `<module>...<module/>` in modulename.ZPM?

isc-tleavitt:
> @isc-tleavitt Can you confirm this is what we want? Or do we need some more wrapper around `<module>...<module/>` in modulename.ZPM?

This looks right to me appearance-wise!!

isc-tleavitt (approved):
I think this looks good but I'd like @isc-kiyer's eyes on it too.

isc-kiyer (approved):

Issue #366, opened 2023-01-23
Log local package installation history


isc-tleavitt:
I just advised an ipm user on how to track a *local* history of installations of their package through `<Invoke>` elements, but really this is something the package manager could/should just do itself.

History should be logged for "load", "install" (including as a dependency), and "uninstall".

History records should include:
* Package name
* Version
* Source registry information (for install only)
  * Name (uniquely identifies registry)
  * Moniker (type of registry in friendlier terms than a classname)
  * Details (would capture e.g. the URL or filesystem path)
* Timestamp
* Namespace (set to $Namespace via InitialExpression)
* Username (set to $Username via InitialExpression)
* Success (%Status)
* Committed (e.g., 0 in non-developer mode if installation fails; 1 otherwise)
* IPM command string (e.g., if I zpm "load https://github.com/intersystems/git-source-control -b hcc" would be good to be able to see that)

Note that we still want to log failed installations - this is a minor wrinkle from a transaction handling perspective; if install rolls back we still need the history.

The command zpm "history" should display package installation history. We also need a documented API for retrieving it (even just saying "you can run SQL queries against this table" would be fine); we should also add a cross-namespace class query wrapper around that API to get all history at the instance level.

Should also investigate feasibility of user audit event types in the IRIS Audit database with corresponding data.

The user persona here is a user of IPM looking for history on their own instance - not a publisher of packages looking for analytics on usage.

isc-tleavitt:
Clarifying: this is a *local* history on the instance where the packages are installed.

evshvarov:
Got you. Such a functionality is already implemented in zpmhub: https://openexchange.intersystems.com/package/zpmhub

isc-tleavitt:
Would be great to expose this via zpm "history" or something like that, and also to provide a public API for programmatic access.

isc-tleavitt:
@isc-kiyer @isc-rkuszews @isc-eneil - could you chime in if you think anything is missing from the spec? This is very important for v1 / inclusion in IRIS.

isc-kiyer:
@isc-tleavitt I don't know what Committed is but the rest makes sense to me. I like the idea of using the IRIS Audit log since we can then leverage its existing UI and capability to turn auditing on/off for specific Audit events (for example, a user may only care about failed installs).
We can still have zpm "history" to wrap around the IRIS audit log but have that as the common source of info.

PR #578, opened 2024-10-03
V1next history log

isc-shuliu:
Implement #366

### DIFF
--- a/src/cls/IPM/General/History.cls
+++ b/src/cls/IPM/General/History.cls
@@ -1,0 +1,79 @@
+Class IPM.General.History Extends %Persistent
+{
+
+Property Action As %String(VALUELIST = ",load,install,uninstall") [ Required ];
+
+Property Package As %IPM.DataType.ModuleName [ Required ];
+
+Property Version As %IPM.General.SemanticVersion [ Required ];
+
+Property SourceName As %String(MAXLEN = 100);
+
+Property SourceMoniker As %String(MAXLEN = 100);
+
+Property SourceDetails As %String(MAXLEN = "");
+
+Property TimeStamp As %TimeStamp [ Required ];
+
+Property NameSpace As %String [ InitialExpression = {$NAMESPACE}, Required ];
+
+Property UserName As %String [ InitialExpression = {$USERNAME}, Required ];
+
+Property Success As %Status [ Required ];
+
+Property Committed As %Boolean [ Required ];
+
+Property CommandString As %String(MAXLEN = 8192) [ Required ];
+
+Storage Default
+{
+<Data name="HistoryDefaultData">
+<Value name="1">
+<Value>%%CLASSNAME</Value>
+</Value>
+<Value name="2">
+<Value>Package</Value>
+</Value>
+<Value name="3">
+<Value>Action</Value>
+</Value>
+<Value name="4">
+<Value>Version</Value>
+</Value>
+<Value name="5">
+<Value>SourceName</Value>
+</Value>
+<Value name="6">
+<Value>SourceMoniker</Value>
+</Value>
+<Value name="7">
+<Value>SourceDetails</Value>
+</Value>
+<Value name="8">
+<Value>TimeStamp</Value>
+</Value>
+<Value name="9">
+<Value>NameSpace</Value>
+</Value>
+<Value name="10">
+<Value>UserName</Value>
+</Value>
+<Value name="11">
+<Value>Success</Value>
+</Value>
+<Value name="12">
+<Value>Committed</Value>
+</Value>
+<Value name="13">
+<Value>CommandString</Value>
+</Value>
+</Data>
+<DataLocation>^IPM.General.HistoryD</DataLocation>
+<DefaultData>HistoryDefaultData</DefaultData>
+<IdLocation>^IPM.General.HistoryD</IdLocation>
+<IndexLocation>^IPM.General.HistoryI</IndexLocation>
+<StreamLocation>^IPM.General.HistoryS</StreamLocation>
+<Type>%Storage.Persistent</Type>
+}
+
+}
--- a/src/cls/IPM/General/History.cls
+++ b/src/cls/IPM/General/History.cls
@@ -1,4 +1,4 @@
-Class IPM.General.History Extends %Persistent
+Class %IPM.General.History Extends %Persistent
 {
 
 Property Action As %String(VALUELIST = ",load,install,uninstall") [ Required ];
@@ -13,17 +13,80 @@ Property SourceMoniker As %String(MAXLEN = 100);
 
 Property SourceDetails As %String(MAXLEN = "");
 
-Property TimeStamp As %TimeStamp [ Required ];
+Property TimeStart As %TimeStamp [ Required ];
+
+Property TimeEnd As %TimeStamp [ Required ];
 
 Property NameSpace As %String [ InitialExpression = {$NAMESPACE}, Required ];
 
 Property UserName As %String [ InitialExpression = {$USERNAME}, Required ];
 
-Property Success As %Status [ Required ];
+Property Success As %Status [ InitialExpression = 0, Required ];
 
-Property Committed As %Boolean [ Required ];
+Property Committed As %Boolean [ InitialExpression = 0, Required ];
 
 Property CommandString As %String(MAXLEN = 8192) [ Required ];
+
+ClassMethod Init(Action As %String, Package As %IPM.DataType.ModuleName, Command As %String) As %IPM.General.History [ Private ]
+{
+    Set log = ..%New()
+    Set log.TimeStart = $ZDateTime($Now(), 3)
+    Set log.TimeEnd = $ZDateTime($Now(), 3) // Default to now so can call %Save(), will be updated in Finalize
+    Set log.Action = Action
+    Set log.Package = Package
+    Set log.CommandString = Command
+    Set log.NameSpace = $Namespace
+    Set log.UserName = $Username
+    $$$ThrowOnError(log.SetVersion()) // Set the version to placeholder 0.0.0-0, Will be updated before finalizing
+    $$$ThrowOnError(log.%Save()) // Save it now in case Finalize isn't called
+    Quit log
+}
+
+ClassMethod InstallInit(Package As %IPM.DataType.ModuleName, Command As %String) As %IPM.General.History
+{
+    Quit ..Init("install", Package, Command)
+}
+
+ClassMethod LoadInit(Package As %IPM.DataType.ModuleName, Command As %String) As %IPM.General.History
+{
+    Quit ..Init("load", Package, Command)
+}
+
+ClassMethod UninstallInit(Package As %IPM.DataType.ModuleName, Command As %String) As %IPM.General.History
+{
+    Quit ..Init("uninstall", Package, Command)
+}
+
+Method SetSource(pInfo As %IPM.Storage.QualifiedModuleInfo) As %Status
+{
+    $$$ThrowOnError(..%Reload())
+    Set ..SourceName = pInfo.ServerName
+    Set ..SourceMoniker = pInfo.ServerMoniker
+    Set ..SourceDetails = pInfo.ServerDetails
+    Quit ..%Save()
+}
+
+Method SetVersion(pVersion As %IPM.General.SemanticVersion) As %Status
+{
+    If ..%Id() '= "" {
+        $$$ThrowOnError(..%Reload())
+    }
+    If $Get(pVersion) = "" {
+        Set pVersion = "0.0.0-0"
+    }
+    Set ..Version = $Select($IsObject(pVersion): pVersion, 1: ##class(%IPM.General.SemanticVersion).FromString(pVersion))
+    Quit ..%Save()
+}
+
+Method Finalize(status As %Status) As %Status
+{
+    $$$ThrowOnError(..%Reload())
+    Set ..Success = status
+    // TODO what does committed mean?
+    Set ..Committed = 1
+    Set ..TimeEnd = $ZDateTime($Now(), 3)
+    Return ..%Save()
+}
 
 Storage Default
 {
@@ -32,10 +95,10 @@ Storage Default
 <Value>%%CLASSNAME</Value>
 </Value>
 <Value name="2">
-<Value>Package</Value>
+<Value>Action</Value>
 </Value>
 <Value name="3">
-<Value>Action</Value>
+<Value>Package</Value>
 </Value>
 <Value name="4">
 <Value>Version</Value>
@@ -50,29 +113,32 @@ Storage Default
 <Value>SourceDetails</Value>
 </Value>
 <Value name="8">
-<Value>TimeStamp</Value>
+<Value>TimeStart</Value>
 </Value>
 <Value name="9">
-<Value>NameSpace</Value>
+<Value>TimeEnd</Value>
 </Value>
 <Value name="10">
-<Value>UserName</Value>
+<Value>NameSpace</Value>
 </Value>
 <Value name="11">
-<Value>Success</Value>
+<Value>UserName</Value>
 </Value>
 <Value name="12">
-<Value>Committed</Value>
+<Value>Success</Value>
 </Value>
 <Value name="13">
+<Value>Committed</Value>
+</Value>
+<Value name="14">
 <Value>CommandString</Value>
 </Value>
 </Data>
-<DataLocation>^IPM.General.HistoryD</DataLocation>
+<DataLocation>^%IPM.General.HistoryD</DataLocation>
 <DefaultData>HistoryDefaultData</DefaultData>
-<IdLocation>^IPM.General.HistoryD</IdLocation>
-<IndexLocation>^IPM.General.HistoryI</IndexLocation>
-<StreamLocation>^IPM.General.HistoryS</StreamLocation>
+<IdLocation>^%IPM.General.HistoryD</IdLocation>
+<IndexLocation>^%IPM.General.HistoryI</IndexLocation>
+<StreamLocation>^%IPM.General.HistoryS</StreamLocation>
 <Type>%Storage.Persistent</Type>
 }
 
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -754,6 +754,7 @@ ClassMethod ShellInternal(pCommand As %String, Output pException As %Exception.A
 			Do ..GetNamespaceDefaultModifiers(.tDefaultArray)
 			Merge tCommandInfo("data") = tDefaultArray
 			Merge tCommandInfo = tParsedCommandInfo
+			Set tCommandInfo("CommandString") = tCommand
 			
 			If (tCommandInfo = "quit") {
 				Return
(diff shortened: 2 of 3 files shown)


isc-tleavitt:
Re: "committed" - this would track, in non-dev mode, if installation was attempted but failed in some way. The reason for the failure (probably a %Status) would be worth recording as well for a failed installation.

isc-kiyer:
> Re: "committed" - this would track, in non-dev mode, if installation was attempted but failed in some way - that is, it would be set to 0 in this case and 1 if installation succeeded. The reason for the failure (probably a %Status) would be worth recording as well for a failed installation.

This comes into the overall lifecycle rework where we flag the phase in which a module reached during install and track the error that occurred as well. For modules in dev mode, can permit resuming install from the failed phase onwards and for non-dev mode, always start from scratch for the module.

isc-shuliu:
Not sure why [2c240de](https://github.com/intersystems/ipm/pull/578/commits/2c240dee4afe73ab2ea0c154ee1309c6fb774575) doesn't work. 

I have a local registry with `modA` and `modB`, where `modB` depends on `modA`. When installing `modB`, IPM will call `%IPM.Utils.Module:LoadDependencies()`, which spawned a new process that runs `..LoadModuleReference()`. 

The changes made by commit [2c240de](https://github.com/intersystems/ipm/pull/578/commits/2c240dee4afe73ab2ea0c154ee1309c6fb774575) is to simply create a %Persistent history log object, save it in database, and pass the ID to the work unit. 

However, when calling %Save() from the spawned process, it raises the error `Failed to acquire exclusive lock on instance of '%IPM.General.History`

isc-shuliu:
It appears the locks is created because we're in a transaction. (Tried stepping through the code while monitoring management portal, lock disappears after manually running TCOMMIT)

isc-shuliu:
Here's how it currently looks like
image

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,394 @@
...
+    }
+    Return ..%Save()
+}
+
+/// @API.Method
+/// Get the history of all installations, uninstalls, and loads in given namespace
+/// The filter argument is a multidimensional array with structure 
+/// filter(columnName) = value
+/// Where value can optionally start with >, >=, <, <=, =, <> or contain * to indicate a wildcard
+ClassMethod GetHistory(ByRef filter, ascend As %Boolean = 0, limit As %Integer = 0, namespace As %String) As %SQL.StatementResult
```
isc-tleavitt, line 112: If we're going to make an API method that returns a %SQL.StatementResult, I'd rather expose it as a class query.

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,394 @@
...
+    Set duration = $Select(diffDays: diffDays _ "D ", 1:"") 
+        _$Select(diffHours: diffHours _ "h ", 1:"")
+        _$Select(diffMinutes: diffMinutes _ "m ", 1:"")
+        _$Select(diffSeconds: diffSeconds _ "s ", 1:"")
+
+    If duration = "" {
+        Set duration = "0s"
+    }
+
+    Quit rs.%Get("TimeStart") _ " (" _ $Zstrip(duration, "<>W") _ ")"
```
isc-tleavitt, line 275: I'd wonder about using $System.SQL.DATEDIFF and just always using seconds as the unit. Ideally this will always be a manageably small number.

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,394 @@
...
+<Value>Success</Value>
+</Value>
+<Value name="13">
+<Value>Committed</Value>
+</Value>
+<Value name="14">
+<Value>CommandString</Value>
+</Value>
+</Data>
+<DataLocation>^%IPM.General.HistoryD</DataLocation>
```
isc-tleavitt, line 386: How is this mapped?

isc-tleavitt (changes requested):
A few really minor comments, but I think this is great overall.

src/cls/IPM/General/History.cls
isc-shuliu: The data is mapped to each namespace where %IPM package is mapped. Do we want to separate them?

isc-shuliu:
@isc-tleavitt 
Based on your feedback, I changed the time diff display to always show seconds and exposed API queries instead of API methods. 

Currently, the table data is shared among all namespaces where %IPM is mapped. Do we want to separate them?

isc-tleavitt (dismissed):

isc-tleavitt:
I think this is good now, just need @isc-kiyer's review

src/cls/IPM/Main.cls
```diff
@@ -2867,6 +2916,44 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
+	Set action = $Get(pCommandInfo("parameters", "action"))
+
+	If (action = "find") || (action = "") {
+		Set globally = $$$HasModifier(pCommandInfo, "globally")
+		Set ascend = $$$HasModifier(pCommandInfo, "ascend")
+		Set limit = $$$GetModifier(pCommandInfo, "limit")
+
+		Merge filter = pCommandInfo("data", "zpm")
+		Set mthd = $Select(globally: "GetHistoryGlobally", 1: "GetHistory")
+		Set rs = $Classmethod("%IPM.General.History", mthd, .filter, ascend, limit)
```
isc-kiyer, line 2930: Nit: Rather than using dynamic method dispatch which loses intellisense, could have GetHistory() call GetHistoryGlobally() if it is a global search based on a method arg. Same with below

src/cls/IPM/Main.cls
```diff
@@ -2867,6 +2916,44 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
+			If idx = "" {
+				Quit
+			}
+			If $Extract(col, 1, 3) = "x__" {
+				Continue
+			}
+			Write !, "- ", col
+		}
+	} ElseIf action = "delete" {
+		Do ##class(%IPM.General.History).%KillExtent()
```
isc-kiyer, line 2951: For delete, should be able to filter what deletes. Such as by time or at a minimum by specific namespace. Should require passing another parameter/modifier to indicate what the scope of delete is

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,479 @@
...
+ClassMethod RecordsClose(ByRef qHandle As %Binary) As %Status [ Internal, PlaceAfter = RecordsFetch ]
+{
+    Set qHandle = ""
+    Quit $$$OK
+}
+
+/// Get the history of all installations, uninstalls, and loads in given namespace
+/// The filter argument is a multidimensional array with structure 
+/// filter(columnName) = value
+/// Where value can optionally start with >, >=, <, <=, =, <> or contain * to indicate a wildcard
```
isc-kiyer, line 167: Rather than value starting with such things, perhaps make value a $ListBuild? The possible issue with it starting with that is that the property's value may actually be using one of these symbols.

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,479 @@
...
+    If '$$$comClassDefined("%IPM.General.History") {
+        Quit $$$NULLOREF
+    }
+    Kill varargs
+    Set limit = +limit // Coerce to number to prevent SQL injection
+    Set query = "SELECT " _ $SELECT(limit>0: "TOP " _limit, 1: "") _  " * FROM %IPM_General.History WHERE 1 = 1 "
+    Set col = ""
+    For {
+        Set col = $Order(filter(col), 1, value)
+        If (col = "") || ($Data(value) # 2 = 0) || (col '? 1A.AN){
```
isc-kiyer, line 183: Nit: Use $Match instead of pattern match

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,479 @@
...
+
+/// Get the history of all installations, uninstalls, and loads in given namespace
+/// The filter argument is a multidimensional array with structure 
+/// filter(columnName) = value
+/// Where value can optionally start with >, >=, <, <=, =, <> or contain * to indicate a wildcard
+ClassMethod GetHistory(ByRef filter, ascend As %Boolean = 0, limit As %Integer = 0, namespace As %String) As %SQL.StatementResult
+{
+    New $namespace
+    If $Data(namespace) # 2 {
+        Set $namespace = $ZStrip(namespace, "<>WC")
```
isc-kiyer, line 172: So over here, no namespace switching is needed. You can simply choose to include namespace filtering as an arg in the query to distinguish global vs local.
This is because the records are stored in a % global in %IPM.Storage.History so are in %SYS and available everywhere.
@isc-shuliu @isc-tleavitt is this the intention or should the history table be namespace specific, in which case this would need to switch to each namespace to run the query in?

src/cls/IPM/Main.cls
```diff
@@ -1863,6 +1888,19 @@ ClassMethod LoadFromRepo(tDirectoryName, ByRef tParams) [ Internal ]
 }
 
 ClassMethod Load(ByRef pCommandInfo) [ Internal ]
+{
+	Set devMode = $Get(pCommandInfo("data", "DeveloperMode"), 1)
+	Set log = ##class(%IPM.General.History).LoadInit()
+	Try {
+		Do ..LoadInternal(.pCommandInfo, log)
+	} Catch Ex {
+		$$$ThrowOnError(log.Finalize(Ex.AsStatus(), devMode))
```
isc-kiyer, line 1897: General question: if the module install fails and transactions are enabled, does rolling back of the transaction result in the history data getting rolled back as well?

isc-kiyer (changes requested):
@isc-shuliu This is a great and exhaustive feature! My comments and some questions are up

src/cls/IPM/General/History.cls
isc-shuliu: That's a good idea, my concern is that It can be cumbersome for users to construct a command string that contains `$ListBuild()`s. They would need to do something like
```objectscript
zpm "history find -DTimeStart=" _ $lb(">", "2000-01-01 00:00:00")
```
And I can't think of a way for it to work in interactive mode:
```
USER>zpm

=============================================================================
|| Welcome to the Package Manager Shell (ZPM). Version:                    ||
|| Enter q/quit to exit the shell. Enter ?/help to view available commands ||
|| Current registry https://pm.community.intersystems.com                  ||
=============================================================================
zpm:USER>history find -DTimeStart=???
```

src/cls/IPM/General/History.cls
isc-shuliu: Just fixed it. We don't need the table to be namespace specific. All we care about is being able to filter by namespace.

isc-tleavitt:
This should account for #704 (the pitfalls of being stuck here for a while, while we've been focused on more pressing things...)

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+Property TimeStart As %TimeStamp [ Required ];
+
+/// Time when the action ended. This will be empty if the action is still in progress or ended abnormally.
+Property TimeEnd As %TimeStamp;
+
+/// Namespace where the action is being performed
+Property NameSpace As %String [ InitialExpression = {$NAMESPACE}, Required ];
+
+/// User who initiated the action
+Property UserName As %String [ InitialExpression = {$USERNAME}, Required ];
```
isc-kiyer, line 35: Nit: Use %Library.Username

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+
+/// Status of the action. If the action is still in progress, this will be 0. Otherwise, it will be the status code of the action.
+Property Success As %Status [ InitialExpression = 0, Required ];
+
+/// Whether the action is committed. In non developer mode, this indicates whether the action was successful. In developer mode, this is always true.
+Property Committed As %Boolean [ InitialExpression = 0, Required ];
+
+/// The command string that triggered the action
+Property CommandString As %String(MAXLEN = 8192) [ Required ];
```
isc-kiyer, line 44: Does command string need a maxlen?

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+
```
isc-kiyer, line 45: Currently there are no indices on this. I would guess it won't get too large but worthwhile adding some?

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+    Set ..SourceName = ServerName
+    Set ..SourceMoniker = server.#MONIKER
+    Set ..SourceDetails = server.Details
+    Quit ..%Save()
+}
+
+Method SetVersion(pVersion As %IPM.General.SemanticVersion) As %Status
+{
+    If ..%Id() '= "" {
+        $$$ThrowOnError(..%Reload())
```
isc-kiyer, line 92: For my understanding, why is an ID check needed here but not in SetName? If no particular reason, can call Reload() without the check in both. If there is a reason, please add a comment

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+/// Namespace where the action is being performed
+Property NameSpace As %String [ InitialExpression = {$NAMESPACE}, Required ];
+
+/// User who initiated the action
+Property UserName As %String [ InitialExpression = {$USERNAME}, Required ];
+
+/// Status of the action. If the action is still in progress, this will be 0. Otherwise, it will be the status code of the action.
+Property Success As %Status [ InitialExpression = 0, Required ];
+
+/// Whether the action is committed. In non developer mode, this indicates whether the action was successful. In developer mode, this is always true.
```
isc-kiyer, line 40: For my understanding: what does it mean for an action to be "committed"?

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+{
+    $$$ThrowOnError(..%Reload())
+    Set ..Success = status
+    Set ..Committed = (dev = 1) || $$$ISOK(status)
+    Set ..TimeEnd = $ZDateTime($Now(), 3)
+    Return ..%Save()
+}
+
+/// @API.Query
+/// Get load/install/uninstall records in the specified namespace
```
isc-kiyer, line 118: Nit: Can replace load/install/uninstall with history

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+    Return $$$OK
+}
+
+/// Get the history of all installations, uninstalls, and loads in given namespace
+/// The filter argument is a multidimensional array with structure 
+/// filter(columnName) = value
+/// Where value can optionally start with >, >=, <, <=, =, <> or contain * to indicate a wildcard
+ClassMethod GetHistory(ByRef filter, ascend As %Boolean = 0, limit As %Integer = 0, namespace As %String) As %SQL.StatementResult
+{
+    Set limit = +limit // Coerce to number to prevent SQL injection
```
isc-kiyer, line 222: You can use `?` like in where clauses as `TOP ?` to prevent SQL injection

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+        Do ..ShowColumns(rs, 1)
+    } Else {
+        $$$ThrowStatus($$$ERROR($$$GeneralError, "No record found with ID "_id_". Use `history find` command to show all entries. "))
+    }
+}
+
+/// If detail is 1, will display full details each property on a separate line
+/// If detail is "header", will display the column names using the same style as the data
+/// Otherwise, will display on a single line, with some details omitted
+ClassMethod ShowColumns(rs As %SQL.StatementResult, detail As %Boolean = 0) As %Status [ Internal, Private ]
```
isc-kiyer, line 332: Rather than having detail be a fake boolean, perhaps have it be an enum? So valid options could be "header" and "full" perhaps?

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
...
+    Set qHandle = ""
+    Quit $$$OK
+}
+
+/// Get an SQL "where" clause based on the filter argument
+/// See GetHistory() for the `filter` argument
+/// Output "clause" is the SQL WHERE clause as a string, container zero or more ? placeholders
+/// The "varargs" is an array of values to be substituted in the ? placeholders. 
+/// The "varargs" may already be populated with values, and new values will be appended to it.
+ClassMethod ConstructSQLWhere(filter As %String, namespace As %String, Output clause As %String, ByRef varargs As %String) As %Status [ Internal ]
```
isc-kiyer, line 179: Nit: here filter should be ByRef since it a multi-dim array

src/cls/IPM/Main.cls
```diff
@@ -1889,6 +1915,19 @@ ClassMethod LoadFromRepo(tDirectoryName, ByRef tParams) [ Internal ]
 }
 
 ClassMethod Load(ByRef pCommandInfo) [ Internal ]
+{
+	Set devMode = $Get(pCommandInfo("data", "DeveloperMode"), 1)
```
isc-kiyer, line 1919: Should devMode default be 0? Without -dev flag, it is 0 right?

src/cls/IPM/Main.cls
```diff
@@ -1939,73 +1978,84 @@ ClassMethod Install(ByRef pCommandInfo) [ Internal ]
...
+	If (tModuleName = "") {
+		Quit $$$OK
+	}
 
 	Set tVersion = $Get(pCommandInfo("parameters","version"))
 	Set tKeywords = $$$GetModifier(pCommandInfo,"keywords")
+
+	#dim log As %IPM.General.History
+	Set log = ##class(%IPM.General.History).InstallInit(tModuleName)
+	Set devMode = $Get(tParams("DeveloperMode"), 0)
```
isc-kiyer, line 1990: There is no tParams here. Did you mean to write `pCommandInfo("data","DeveloperMode")` ?

isc-kiyer (changes requested):
@isc-shuliu another round of comments. Also, worthwhile adding unit tests since this is a brand new feature. Some test cases I can think about:
- simple commands with no modifiers
- commands with valid modifiers
- commands with invalid modifiers
- modules in dev and non-dev mode
- testing all commands (load, install, uninstall)
- filtering
- altering number of history results to keep
- searching per namespace vs globally
- error handling for invalid filtering columns

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,524 @@
+Include (%IPM.Common, %IPM.Formatting)
+
+/// Persistent class for history of all installations, uninstalls, and loads in the system. This class is shared among namespaces that share %IPM mappings.
+Class %IPM.General.History Extends %Persistent
+{
+
+/// Action of this history record. Can be load, install, or uninstall
+Property Action As %String(VALUELIST = ",load,install,uninstall") [ Required ];
```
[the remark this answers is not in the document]
isc-dchui: I'm not sure we want to log every command as some are pretty uninteresting (e.g. zpm "list"). The core list of package affecting commands of install, load, uninstall, and update seem most important. Also added some lifecycle phase logging.

isc-kiyer:
> Sorry if this is already addressed, but I am wondering whether there is a housekeeping default of keeping only history within x number of days? In case over time the logs grow to be too much storage burden

Yes this is set to a default of 20 records I believe

src/cls/IPM/General/AbstractHistory.cls
```diff
@@ -0,0 +1,170 @@
...
+{
+
+/// Action of this history record. Can be load, install, uninstall or update
+Property Action As %String(VALUELIST = ",load,install,uninstall,update") [ Required ];
+
+/// Name of the package being logged. This is not necessarily required, e.g. when loading a nonexistent directory.
+Property Package As %IPM.DataType.ModuleName;
+
+/// Version of the package being logged. This is not necessarily required, e.g. when loading a nonexistent directory.
+Property Version As %IPM.General.SemanticVersion [ Required ];
```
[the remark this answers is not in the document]
isc-dchui: Good catch; I've improved the wording to make it clearer.

isc-kiyer (changes requested):
@isc-dchui looks mostly good! Some comments

isc-kiyer:
> @isc-jili, @isc-kiyer no actually the default records of 20 is just how many records will be shown at a given time when running `history`. There's no current mechanism to purge old records automatically. I can see how that might be helpful but I also wonder if that would defeat the point of logging all actions for an audit trail.

So generally with auditing, there is often a purge "by time" mechanism since only audit history for a certain time period matters. Lets spawn off purging into its own issue to not add scope creep to this one.

src/cls/IPM/General/AbstractHistory.cls
```diff
@@ -0,0 +1,170 @@
...
+Property TimeStart As %TimeStamp [ Required ];
+
+/// Time when the action ended. This will be empty if the action is still in progress or ended abnormally.
+Property TimeEnd As %TimeStamp;
+
+/// User who initiated the action
+Property UserName As %Library.Username [ InitialExpression = {$username}, Required ];
+
+/// Status of the action. If the action is still in progress, this will be 0. Otherwise, it will be the status code of the action.
+Property Success As %Status [ InitialExpression = 0, Required ];
```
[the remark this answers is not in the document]
isc-dchui: The %Status result does get logged here! See screenshot for example. It's just set to 0 initially.

Screenshot 2025-12-03 093102

src/cls/IPM/General/History.cls
```diff
@@ -0,0 +1,532 @@
+Include (%IPM.Common, %IPM.Formatting)
+
+/// Persistent class for history of all installations, uninstalls, and loads in the system. This class is scoped by namespace.
+Class %IPM.General.History Extends %IPM.General.AbstractHistory
```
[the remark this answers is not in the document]
isc-dchui: @isc-kiyer I've added an %OnBeforeSave method to prevent editing of a finalized history log entry and a trigger to prevent all SQL updates

isc-tleavitt (dismissed):
This has lingered long enough and had enough eyes on it. I've given it a final shallow review. It's well-tested and keeping it unmerged presents risk. Good to go IMO.

isc-kiyer (dismissed):

Issue #486, opened 2024-05-23
show install date with list action


SCanzano:
when I use the list command to show the modules that have been installed , which shows this

`
zpm:HSCUSTOM>list
rclib-rcipm          1.0.6
rclib-unittest       1.0.2
rclib-rcreportsbasic 1.0.2
rclib-rdm            1.0.1
rclib-system         1.0.2
rclib-rcdlf          1.0.1
rclib-ens            1.0.4
rclib-enslib         1.0.3
rclib-fhir           1.0.1

zpm:HSCUSTOM>
`
I would like some way to either show the install date via a flag or always show the date upon which the module was installed

isc-tleavitt:
This relates to #366 - we don't have the data to back this now, but #366 would provide it and exposing it in "list" with a flag would make a lot of sense.

PR #582, opened 2024-10-11
feat: log and display time of last updated & registry version of each module

isc-shuliu:
Implement #486

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -469,6 +469,7 @@ run C:\Temp\MyCommands.json, where contents of the file are as follows:
 <modifier name="description" aliases="d" dataAlias="Desc" dataValue="1" description="Additional information is displayed for each module." />
 <modifier name="showsource" aliases="ss" description="If specified, show dependency list with local source for installed modules." />
 <modifier name="showtime" aliases="st" description="If specified, show the time of last update for each module" />
+<modifier name="showupstream" aliases="su" value="true" description="If specified, show the lastest registry version for each module" />
 <modifier name="repository" aliases="repo" value="true" description="If specified, only show modules installed that belong to the provided repository." />
 
 </command>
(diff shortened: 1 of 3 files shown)


src/cls/IPM/Main.cls
```diff
@@ -2951,22 +2953,36 @@ ClassMethod DisplayModules(ByRef pList, pNumbered As %Boolean = 0, pWithNamespac
...
   For i=1:1:pList {
     Set info = pList(i)
     Set developerMode = 0
     Set root = ""
-    Set $ListBuild(name, version, externalName, developerMode, root) = info
+    Set $ListBuild(name, version, externalName, developerMode, root, lastUpdated) = info
     Write:i>1 !,?pIndent
     Write $$$FormattedLinePadRight($$$Green, name, width), $$$FormattedLine($$$Blue, version)
+    If $Data(pModifiers("showupstream")) && ($Data(serverVersions(name), version) # 2) {
+      Write " ", $$$FormattedLine($$$Blue, "(ServerSide: "_version_")")
```
isc-tleavitt, line 2975: I'd consider saying "latest available" here rather than "ServerSide" which is confusing.

src/cls/IPM/Main.cls
```diff
@@ -468,6 +468,8 @@ run C:\Temp\MyCommands.json, where contents of the file are as follows:
 <modifier name="tree" aliases="t" description="If specified, show dependency tree for installed modules." />
 <modifier name="description" aliases="d" dataAlias="Desc" dataValue="1" description="Additional information is displayed for each module." />
 <modifier name="showsource" aliases="ss" description="If specified, show dependency list with local source for installed modules." />
+<modifier name="showtime" aliases="st" description="If specified, show the time of last update for each module" />
+<modifier name="showupstream" aliases="su" value="true" description="If specified, show the lastest registry version for each module" />
```
isc-tleavitt, line 472: Typo: lastest -> latest

src/cls/IPM/Main.cls
isc-tleavitt: Opinion: should only show this if there's a newer version available upstream.

src/cls/IPM/Main.cls
```diff
@@ -2951,22 +2953,36 @@ ClassMethod DisplayModules(ByRef pList, pNumbered As %Boolean = 0, pWithNamespac
     $ListBuild("Repository", $$$Blue)
   )
 
+  Kill serverVersions
+  If $Data(pModifiers("showupstream"), repoName) && ##class(%IPM.Repo.Definition).ServerDefinitionKeyExists(repoName) {
```
isc-tleavitt, line 2957: Can we show latest available across all configured repos? Or with a default value if there's no value?

isc-tleavitt (changes requested):

src/cls/IPM/Main.cls
isc-shuliu: I was think package developers may be interested in knowing if the package version published on the registry is older than their current dev version.

isc-shuliu:
@isc-tleavitt 
Just updated the PR to show upstream version from all configured repos where the latest version is different from the local version.

isc-kiyer (approved):

isc-tleavitt (approved):

Issue #427, opened 2023-08-15
Module Dependency Version =* does not seem to work


SCanzano-ready:
I have 2 modules

Module A 
was verison 1.0.1
and then was republished as version 1.0.2

Module B has a dependency on Module A with the version being *
which according to https://github.com/intersystems/ipm/wiki/03.-IPM-Manifest-(Module.xml) 

`Module dependencies, specify each dependent module in <ModuleReference> element.
In dependant module Version element you can specify an exact version (1.2.36) or use asterisk.
<Version>2.2.*</Version> - means the latest version grater than 2.2.0 and less then 2.3.0
<Version>*</Version> - means the latest version`

On my target environment I have Module A installed with version 1.0.1.

When I install Module B which reports to have a dependency on *, the latest version I would expect the process of installing Module B would cause Module A to be upgraded to 1.0.2 but this does not occur.

when BuildDependencyGraph is called in %ZPM.PackageManager.Developer.Module we have

`			// Test whether dependency version is a valid semantic version. These are handled in the first pass only.
			// Other cases are handled in the second pass only.
			
			Set tIsExactVersion = +##class(%ZPM.PackageManager.Core.SemanticVersion).IsValid(tDep.VersionString) kill %objlasterror
			If ((pPass = $$$EXACT) && 'tIsExactVersion) || ((pPass = $$$FUZZY) && tIsExactVersion) {
				Continue
			}
`

and because tIsExactVersion is 0 this method is Continuing to the next Dependecy.

Issue #550, opened 2024-08-22
Package manager should have an upgrade framework


isc-kiyer:
If a package is already installed and needs to be upgraded, the package should be able to indicate upgrade steps to be run between any source and target versions.
The package manager should have separate commands for upgrading a package vs fresh install with the ability to intentionally override and do a fresh install (bypassing upgrade steps) if they want (for example in a dev environment).

isc-shuliu:
@isc-kiyer What are some of the differences to consider between upgrades and fresh new installs? 

I can think of only 1 case where the package developer renames some classes and we need to migrate the associated data (like with %ZPM -> %IPM). 

Are there any more things to look out for?

isc-tleavitt:
@isc-shuliu the primary motivation is data migration/schema change steps, which don't necessarily correspond to a class rename. New indices would be another use case. Really, it's all around data.

HealthShare has some more specific use cases / constraints (along with plenty of examples) and I'd recommend that you and @isc-kiyer or @isc-eneil meet to discuss them.

isc-shuliu:
@isc-tleavitt 
After discussion with the HSOPS team, IPM will provide an abstract class for upgrades. Package developers are responsible for subclassing it to implement their own upgrade logic for every earlier version. 

Each upgrade could contain many methods (tasks), even if just bumping up one version number, so that users can see which tasks are successful. 

IPM will figure out the correct order to run the methods as part of the installation process. We still need to figure out which conventions to use. This GitHub issue page can be used for discussion.

isc-shuliu:
Internal reference: HS.Util.Installer.Upgrade.Loader & HS.Util.Installer.Upgrade.ModuleAbstract

isc-dchui:
Closing this out as MVP 1 is done. Planned future work can be turned into their own issues.

PR #586, opened 2024-10-16
Add command for "update" and "outdated"

isc-shuliu:
Implement the solution mentioned in [#427](https://github.com/intersystems/ipm/issues/427#issuecomment-1999919343)

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -469,6 +469,7 @@ run C:\Temp\MyCommands.json, where contents of the file are as follows:
 <modifier name="description" aliases="d" dataAlias="Desc" dataValue="1" description="Additional information is displayed for each module." />
 <modifier name="showsource" aliases="ss" description="If specified, show dependency list with local source for installed modules." />
 <modifier name="showtime" aliases="st" description="If specified, show the time of last update for each module" />
+<modifier name="showupstream" aliases="su" value="true" description="If specified, show the lastest registry version for each module" />
 <modifier name="repository" aliases="repo" value="true" description="If specified, only show modules installed that belong to the provided repository." />
 
 </command>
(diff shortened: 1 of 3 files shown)


isc-kiyer:
@isc-shuliu this PR is related to #550 where part of running "update" should _upgrade_ a module and run its upgrade steps (if any).

isc-shuliu:
> @isc-shuliu this PR is related to #550 where part of running "update" should _upgrade_ a module and run its upgrade steps (if any).

Thanks for the heads-up! I'm aware of that hence posting it as a draft.

src/cls/IPM/Main.cls
```diff
@@ -2865,6 +2878,62 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
 	}
 }
 
+ClassMethod CanUpdate(modName As %String, semver As %IPM.General.SemanticVersion) As %Boolean
```
isc-kiyer, line 2881: @isc-shuliu  Doesn't look like this method is used?

isc-shuliu: Yes, you're right. I haven't gone to that point yet. My plan is to show if an outdated package can be updated to a particular version without violating dependency constraint. This PR is still a Work-In-Progress

AshokThangavel:
Hi @isc-dchui 
Thanks for the feedback, I'll implement it as a `-outdated (and -o)` flag within the existing list command. I'll also ensure the logic aligns with the newer `update` command infrastructure to keep things consistent.

AshokThangavel:
Hi @isc-dchui 
I've noticed in the module search results between two different API calls. When I use `##class(%IPM.Repo.Utils).SearchRepositoriesForModule(tSearchCriteria,.results)` , it only returns 109 entries and is missing several expected modules.
However, when I execute `##class(%IPM.Main).GetUpstreamPackageVersions(.modules)` it returns over 500 modules.

PR #587, opened 2024-10-22
fix: aggregate all versions of module when searching FS repo

### LINKED ISSUE #583: Filesystem Repo doesn't list modules in descending order of versions
`%IPM.Repo.Filesystem.PackageService:ListModules()` doesn't return the latest version of each module by default, which is different from `%IPM.Repo.Remote.PackageService:ListModules()`. 

This affects #582 and ongoing work of #586

We should enforce a protocol such that any subclass of `%IPM.Repo.IPackageService` implements the `ListModules()` method in a manner that returns the latest version first.

isc-shuliu:
Fix #583

Make sure that 
* `%IPM.Repo.Filesystem.PackageService:ListModules()` returns a list of modules with their **latest** versions; and
* If all versions need to be returned, they appear in the `AllVersions` column, separated by `", "`, instead of creating a new row.

### DIFF
--- a/src/cls/IPM/Repo/Filesystem/Cache.cls
+++ b/src/cls/IPM/Repo/Filesystem/Cache.cls
@@ -37,7 +37,7 @@ ClassMethod %OnBeforeBuildIndices(ByRef indexlist As %String(MAXLEN="") = "") As
 	Quit $$$OK
 }
 
-Query OrderedMatches(pRoot As %String = "", pName As %String = "", pVersionExpression As %String = "*", pParameters As %String = "") As %Query(ROWSPEC = "Name:%String,VersionString:%String") [ SqlProc ]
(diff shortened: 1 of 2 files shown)


src/cls/IPM/Repo/Filesystem/Cache.cls
```diff
@@ -62,7 +62,7 @@ ClassMethod OrderedMatchesFetch(ByRef qHandle As %Binary, ByRef Row As %List, By
 		}
 		Set tRoot = ..RootGetStored($ListGet(tRow))
 		If (tRoot = qHandle("root")) {
-			Set Row = $ListBuild(..NameGetStored($ListGet(tRow)),..VersionStringGetStored($ListGet(tRow)))
+			Set Row = $ListBuild(..NameGetStored($ListGet(tRow)),..VersionStringGetStored($ListGet(tRow)))_..VersionGetStored($ListGet(tRow))
```
isc-tleavitt, line 65: This is neat and made me do a double take.

isc-tleavitt (approved):

PR #589, opened 2024-10-22
fix: move namespace filtering from SQL to ObjectScript for IRIS compat.

### LINKED ISSUE #579: Can't install packages via Lite Terminal
This is an upstream issue (ref: DP-435099), but I think we can provide a workaround as it'll be a while before this makes it into IRIS kits.

For example, on an environment with just git-source-control installed:

```
USER>zpm "list"
git-source-control 2.5.0
USER>w $stack
3
USER>d $System.SQL.Purge()
 
USER>w $stack
3
USER>zpm "list"
 
USER>w $stack
40
```

Issue happens in SQL compilation of:
`SELECT Nsp FROM %SYS.Namespace_List() WHERE Nsp = ?`

Workaround would be to implement the WHERE clause in ObjectScript outside of this query.

Should be fixed in PRs targeting both master and v1. We'll want a new 0.7 release.

isc-shuliu:
Fix #579 

Move `WHERE` clause in `SELECT nsp FROM %SYS.Namespace_List()` outside of SQL into ObjectScript code, to avoid the weird Data Platform bug related to Lite terminal in latest vscode extension, as described in #579.

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -902,7 +902,7 @@ ClassMethod GetListNamespace(Output list, pSearch As %String = "")
   Set transformedPieces = ""
   Set ptr = 0
   While $ListNext(pieces, ptr, val) {
-    If val '? .(1AN,1"%") { // namespace
+    If val '? .(1AN,1"%") { // namespace contain only numbers, letters and %, otherwise there's guaranteed no match
       Return
     }
     Set transformedPieces = transformedPieces _ $ListBuild(1_$$$QUOTE(val))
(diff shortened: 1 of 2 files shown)


src/cls/IPM/Main.cls
```diff
@@ -889,20 +889,39 @@ ClassMethod Namespace(ByRef pCommandInfo) [ Internal ]
...
+  Set pieces = $ListFromString(pSearch, "*")
+  Set transformedPieces = ""
+  Set ptr = 0
+  While $ListNext(pieces, ptr, val) {
+    If val '? .(1AN,1"%") { // namespace contain only numbers, letters and %, otherwise there's guaranteed no match
+      Return
+    }
+    Set transformedPieces = transformedPieces _ $ListBuild(1_$$$QUOTE(val))
+  }
+  Set pattern = $ListToString(transformedPieces, ".(1AN,1""%"")")
```
isc-kiyer, line 910: Nit: Better to pattern match using $Match with regex as it is more performant (also regex is easier for anyone to understand instead of proprietary matching rules in objectscript)

isc-tleavitt (approved):

Issue #526, opened 2024-07-26
Make pip install process faster/smarter


isc-tleavitt:
I'm working on a project with a large number of large Python dependencies in requirements.txt and installation takes forever.

It would be nice if there was a way to bypass processing requirements.txt OR a way to say "only process if it's changed since last time" to avoid this overhead every time zpm "load" is called.

isc-tleavitt:
My quick solution for this in my local base lifecycle class:
```
Method InstallPythonRequirements(pRoot As %String = "", ByRef pParams)
{
  Set cmd = $Get(pParams("cmd"))
  If '$ListFind($ListBuild("install", "load"), cmd) {
    Quit $$$OK
  }
  Set tVerbose = $Get(pParams("Verbose"))
  Set pythonRequirements = ##class(%File).NormalizeFilename("requirements.txt", pRoot)
  If '##class(%File).Exists(pythonRequirements) {
    Quit $$$OK
  }
  If $Data(pParams("BypassPython"),bypass)#2 && bypass {
    Do ..Log("requirements.txt SKIPPED")
	Quit $$$OK
  }
  ```
  
  Then run load with -DBypassPython=1

isc-shuliu:
It looks great and I suggest using `-DBypassPython` to `-DBypassPythonDeps` or `-DBypassPythonReqs`. After all, who knows where else we need to use Python in the future 😄

PR #595, opened 2024-10-24
feat: allow bypass installation of python dependencies

isc-shuliu:
Implement #526 

Allow bypass installation of python dependencies

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -251,7 +251,7 @@ load https://github.com/user/repository.git
 load https://github.com/user/repository.git -branch feature-1
 </example>
 <example description="Loads the module described in C:\module\root\path\module.xml without installing python dependencies">
-	load -DBypassPyDeps=1 C:\module\root\path\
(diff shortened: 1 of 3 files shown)


src/cls/IPM/Main.cls
```diff
@@ -250,6 +250,9 @@ load -dev -verbose C:\module\root\path\module-0.0.1.tgz
 load https://github.com/user/repository.git
 load https://github.com/user/repository.git -branch feature-1
 </example>
+<example description="Loads the module described in C:\module\root\path\module.xml without installing python dependencies">
+	load -DBypassPyDeps=1 C:\module\root\path\
```
isc-kiyer, line 254: Instead of using a -D modifer which doesn't get documented in a method's regular commands, can you explicitly add the modifier to the load and install commands?
Long term, would be nice to get rid of -D modifiers and only deal with those that are registered. Custom lifecycle/processor classes can register similar to how the repo command is added to from separate XData blocks.

isc-kiyer (approved):

isc-tleavitt (approved):

PR #600, opened 2024-10-25
V1 integration test cleanup

### LINKED ISSUE #563: Integration tests does not delete dummy modules
So far the integration tests has only been running on GitHub CI servers so this hasn't been an issue, but we should clean up the dummy modules installed in each test case. This also tests the uninstall functionality.

isc-shuliu:
When finished, will fix #563 
Trying to record all packages during setup and remove new packages in teardown.
Somehow this is not working yet. All modules are uninstalled (even the preexisting ones)


isc-dchui:
Closing because #563 has already been addressed.

PR #609, opened 2024-10-30
feat(package): allow passing -export-deps to override lifecycle behavior

### LINKED ISSUE #608: Packaging with dependencies should not require a custom lifecycle class
Currently to package with dependencies you'd need a custom lifecycle class that overrides the EXPORTDEPENDENCIES parameter to set it to 1. This should instead be a command-line argument such as `-export-deps`

isc-shuliu:
Fix #608 
Allow users to pass in a `-export-deps` flag to override lifecycle behavior and export dependencies of a module.

### DIFF
--- a/src/cls/IPM/Lifecycle/Base.cls
+++ b/src/cls/IPM/Lifecycle/Base.cls
@@ -864,7 +864,7 @@ Method %Export(ByRef pParams, ByRef pTargetDirectory As %String, Output pDepende
 	Set tSC = $$$OK
 	Try {
 		Set tVerbose = $Get(pParams("Verbose"))
-    Merge tParams = pParams
+		Merge tParams = pParams
 
 		Kill pDependencyGraph
 		If ($Get(pTargetDirectory) = "") {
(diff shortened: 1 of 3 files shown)


src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -881,15 +881,16 @@ Method %Export(ByRef pParams, ByRef pTargetDirectory As %String, Output pDepende
 				Quit
 			}
 		}
-    Set tParams("ModuleExportPath") = pTargetDirectory
+		Set tParams("ModuleExportPath") = pTargetDirectory
 		
-		Set tSC = ..Module.GetResolvedReferences(.tResourceArray,..#EXPORTDEPENDENCIES,..PhaseList,1,.pDependencyGraph)
+		Set tExportDependencies = $Get(pParams("ExportDependencies"), ..#EXPORTDEPENDENCIES)
```
isc-eneil, line 886: Out of curiosity: why not choose to update the EXPORTDEPENDENCIES parameter continue to use that in the condition below? And if not, is there any reason to keep the EXPORTDEPENDENCIES parameter in the class?
@isc-shuliu

isc-shuliu: I was thinking that we need to maintain backward compatibility. There may be existing published packages that make use of `#EXPORTDEPENDENCIES` parameter already, so we should default back to the value of `#EXPORTDEPENDENCIES`

src/cls/IPM/Main.cls
```diff
@@ -70,6 +70,7 @@ Currently, there may only be one of these per namespace.
 <modifier name="dev" dataAlias="DeveloperMode" dataValue="1" description="Sets the DeveloperMode flag for the module's lifecycle. Key consequences of this are that ^Sources will be configured for resources in the module, and installer methods will be called with the dev mode flag set." />
 <modifier name="quiet" aliases="q" dataAlias="Verbose" dataValue="0" description="Produces minimal output from the command." />
 <modifier name="verbose" aliases="v" dataAlias="Verbose" dataValue="1" description="Produces verbose output from the command." />
+<modifier name="export-deps" dataAlias="ExportDependencies" dataValue="1" description="If specified, will export dependencies of the module to the filesystem. If omitted, defaults to the value of the #EXPORTDEPENDENCIES in lifecycle class. Can be manually set to 0 with -DExportDependencies=0. This modifier is only used in &quot;Package&quot; lifecycle." />
```
isc-kiyer, line 73: Do not recommend usage of -D modifiers as we should plan to remove support for them eventually. They are great for initial dev but not good with documentation and discovery and just appear as magical modifiers.

isc-eneil (approved):

src/cls/IPM/Main.cls
isc-shuliu: Shall we add a `-no-export-deps` modifier to account then? If a legacy package has #EXPORTDEPENDENCIES set to 1 and we want to explicitly exclude dependencies during packing, we need a way to override it.

isc-shuliu: Never mind, I can require arg for -export-deps

isc-tleavitt (approved):

isc-kiyer (approved):

PR #614, opened 2024-10-31
Fix: add makedeployed lifecycle method (HSIEO-11006)

isc-eneil:
Overview of changes:
1) Removing the call to the MakeDeployed lifecycle method from the Activate lifecycle method (in IPM.Lifecycle.Module.cls). In HS methods that build our kits and images, we now independently call MakeDeployed via the IPM shell command.
2) Adding a "makedeployed" shell command and a "recurse" flag option that recursively deploys all those resources in the module and its dependencies that are marked as deployed.
3) Updating some methods that match an inputted phase to the correct lifecycle method to account for CamelCased lifecycle phases (e.g. ExportData, MakeDeployed).

Tested makedeployed -only, activate -only, and package commands before and after updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #454: IPM 0.9.x+ uses different globals for storage vs. 0.7.0 and previous. Installation will automatically migrate data from the old globals to the new ones. The old globals are left around in case the user decides to revert to an earlier version.
 
 ### Fixed
+- HSIEO-11006: Fix conditions for marking code as deployed
 - HSIEO-9269, HSIEO-9402: % percent perforce directories are no longer necessary
 - HSIEO-9269, HSIEO-9404: Repo check should happen in the order to repo creation, not by repo name
 - HSIEO-9269, HSIEO-9411: Make sure can load and export xml Package-type resource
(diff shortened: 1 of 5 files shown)


isc-eneil:
Also, note that this PR represents the changes in HSIEO-10801, which is the only outstanding HS Jira with IPM changes that had not already been merged into this GitHub IPM repo.

More details on my process of verifying outstanding differences between the two IPM versions:
- I performed a diff on the two repositories (//projects/ipm-modules/sysmodules/IPM/latest/src/ and the src/ folder of the v1 branch of this repo) using Compare Folders, a useful VSCode extension
- This revealed the following:
-- HS IPM does not contain any files that are not in the community IPM version
-- IPM on GitHub contains some new files:
--- ipm/src/IPM/Repo/Remote/Info.cls
--- ipm/src/IPM/ResourceProcessor/WebApplication.cls
--- ipm/src/IPM/Utils/LegacyCompat.cls
--- ipm/src/IPM/Migration.cls
-- There are a number of small changes to many other files 
- For each common file that had differences, I examined the HS IPM file in Perforce to see if any changes had been made since a commit to reconcile versions. If yes, I then checked whether this change was present in the GitHub IPM v1 code.

isc-kiyer (approved):

isc-jlechtne (approved):

isc-shuliu (approved):

isc-tleavitt (approved):

Issue #649, opened 2024-12-10
Authentication error when switching between username/password and token


nsolov:
When I first specify a username and password for a registry (e.g., repo -n registry -r -url https://test.pm.community.intersystems.com/registry/ -user test -pass PassWord42), and then attempt to use a different registry that requires a token (e.g., repo -r -n registry -url https://pm.intersystems.com -token <mykey>), an error occurs. The problem is that the username and password are not cleared and remain set.

I suggest that the credentials (username/password) should be cleared automatically when the registry URL changes.

PR #650, opened 2024-12-10
clear all existing credentials

nsolov:
fix https://github.com/intersystems/ipm/issues/649

when registry URL changing - clear all existing credentials


isc-tleavitt:
@nsolov could you retarget this against the v1 branch? We're very close to merging / releasing v0.9 and I suspect this is still an issue there.

cc @isc-shuliu

isc-tleavitt:
Actually, this is good to merge to master; we'll just duplicate against v1

isc-kiyer (approved):

isc-tleavitt:
Going to address this in v1 along with #787 (which seems related) today.

PR #657, opened 2024-12-11
HSIEO-11006: Fix conditions for marking code as deployed

isc-eneil:
Overview of changes:

1. Removing the call to the MakeDeployed lifecycle method from the Activate lifecycle method (in IPM.Lifecycle.Module.cls). In HS methods that build our kits and images, we now independently call MakeDeployed via the IPM shell command.
2. Adding a "makedeployed" shell command and a "recurse" flag option that recursively deploys all those resources in the module and its dependencies that are marked as deployed.
3. Updating some methods that match an inputted phase to the correct lifecycle method to account for CamelCased lifecycle phases (e.g. ExportData, MakeDeployed).


isc-kiyer (approved):

isc-tleavitt:
Good to merge after CI passes

PR #660, opened 2024-12-13
Updating FileCopy to check for $ variables in path

isc-jlechtne:


isc-tleavitt (changes requested):
Needs changelog.md; also, it's valid to use ${ or {$ so should probably support both.

src/cls/IPM/ResourceProcessor/FileCopy.cls
```diff
@@ -39,7 +39,12 @@ Method OnBeforePhase(pPhase As %String, ByRef pParams) As %Status
 		}
 
 		If (pPhase = "Activate") && (..InstallDirectory '= "") && ('..Defer) {
```
isc-eneil, line 41: @isc-jlechtne, it seems like everything in this if statement and the one below is identical. Perhaps that can be separated out into a helper method that cleans up the source directory, sets the target, and runs DoCopy(...) iff the phase is Activate and the Installer directory is not empty. 

Or at least, break out the logic to clean up the source directory?

isc-eneil (changes requested):

isc-shuliu (approved):

src/cls/IPM/ResourceProcessor/FileCopy.cls
```diff
@@ -121,7 +132,21 @@ Method NormalizeNames(ByRef pSource As %String, ByRef pTarget As %String, Output
   }
 }
 
-Method DoCopy(tSource, tTarget, pParams)
+Method GetSource()
```
isc-eneil, line 135: @isc-jlechtne 
Nit: I'd name this GetNormalizedSourcePath.
Also, it's worth adding one line of method documentation to explain that this prefixes the path to the module root if no $ variables are used.

isc-eneil (changes requested):

src/cls/IPM/ResourceProcessor/FileCopy.cls
isc-jlechtne: @isc-eneil I don't want to name it Normalized since the normalization actually happens in the DoCopy method. My comment inside the method is kind of the inverse of what you're saying but changing it to a method description instead as you suggested does make sense.

isc-eneil: That's fair. I'm good with GetSource then.

isc-eneil (approved):

PR #666, opened 2024-12-17
V0.10.x update to date

isc-shuliu:
Merging changes from main (v0.9.x) branch


isc-shuliu:
The CIs are failing on GitHub but passing for me locally. Specifically, when migrating from ZPM 0.7, the `zpm "list"` command doesn't list `zpm` itself on GitHub while it does for me locally.

isc-tleavitt:
> The CIs are failing on GitHub but passing for me locally. Specifically, when migrating from ZPM 0.7, the `zpm "list"` command doesn't list `zpm` itself on GitHub while it does for me locally.

@isc-shuliu it looks like we install `zpm` in `%SYS`. In that context, it shouldn't exist in `USER`.

isc-shuliu:
It seems the legacy ZPM 0.7.x is always installed in %SYS but CIs were passing fine. https://github.com/intersystems/ipm/actions/runs/12380237397/job/34556166023#step:11:778

I'm guessing something in the migration logic changed this behavior.

isc-shuliu:
@isc-tleavitt 
What is the expected behavior when a user migrate from ZPM 0.7 (which is always installed in %SYS) to the latest IPM from another namespace, say USER? I know we should map IPM globally and migrate repo & module definitions. 

But should we list the new 0.10.x `zpm` module in USER or %SYS?

isc-eneil (approved):

isc-tleavitt:
> @isc-tleavitt What is the expected behavior when a user migrate from ZPM 0.7 (which is always installed in %SYS) to the latest IPM from another namespace, say USER? I know we should map IPM globally and migrate repo & module definitions.
> 
> But should we list the new 0.10.x `zpm` module in USER or %SYS?

We should list it wherever the installer was run.

PR #678, opened 2025-01-03
fix: preserve user-side changes to lang. ext. outside of comment tags

### LINKED ISSUE #674: Installing IPM
Installing IPM doesn't consider existing %ZLANGC00 and %ZLANGF00 implementations.
Thanks for removing my own Zxx implementations.
Please parse any existing implementations and overwrite only the IPM part.

isc-shuliu:
Fix #674 
Make sure we don't overwrite user custom code in language extensions.


isc-tleavitt (changes requested):
Needs CHANGELOG.md entry and a unit test would be nice.

isc-kiyer (approved):

PR #682, opened 2025-01-06
Fix miscellaneous problems in `enable`

isc-shuliu:
Fix ##676
This will fix miscellaneous problems in `enable`


src/cls/IPM/Main.cls
```diff
@@ -2995,23 +2997,25 @@ ClassMethod EnableIPM(ByRef pCommandInfo)
...
-	Kill menuList				// menu to let user choose which version of local ipm to install (if not in quiet mode and no version is specified)
-	While resultSet.%Next(.sc) {
-		$$$ThrowOnError(sc)
-		Set fileName = resultSet.%Get("Name")
-		If (resultSet.%Get("Type") = "F") {
-			// get the exact version string from filename
-			Set exactIPMVer = $Piece($Piece(fileName, "ipm-", 2, *), ".xml", 1, *-1)
-			Set ipmLocalArray(exactIPMVer) = fileName
-			Set menuList($Increment(menuList)) = exactIPMVer_" (local version)"
+	If ##class(%File).DirectoryExists(XMLDir) {
```
isc-kiyer, line 3000: Is this XMLDir something we have decided as the kitted location of IPM when shipping with IRIS?

isc-kiyer (approved):

src/cls/IPM/Main.cls
isc-shuliu: @isc-tleavitt Can you confirm that? (I think you wrote that directory)

isc-tleavitt: I think this goes back to before my time. This is (per earlier context) /<instance install dir>/lib/ipm/ which seems reasonable to me.

isc-tleavitt (approved):

PR #690, opened 2025-01-07
ci: debug changelog check with rev-parse

isc-shuliu:


isc-shuliu:
The issue seems to be actions/checkout only preserves one commit of the repo (the one resulting from merging feature branch into target branch), hence `git diff` doesn't work in this case.

PR #695, opened 2025-01-08
fix(ci): download legacy artifact from comm. registry instead of github

### LINKED ISSUE #694: Investigate why GitHub removes older releases (say v0.7.4)
This has been happening for a while. The main CI relies on the existence of v0.7 release artifacts to perform a migration test. However, the v0.7.x artifacts keep disappearing. First v0.7.2 disappeared and then v0.7.3, now v0.7.4 have disappeared and we have no stable artifacts to use.

https://api.github.com/repos/intersystems/ipm/releases

isc-shuliu:
Download legacy artifact from comm. registry instead of github
Partially fix #694 . We still need to figure out why zpm-0.7.4.xml and other versions were deleted from GitHub

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -174,15 +174,11 @@ jobs:
         run: |
           curl http://localhost:52773/registry/packages/-/all | jq
           curl http://localhost:52773/registry/packages/zpm/ | jq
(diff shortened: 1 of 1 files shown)


isc-shuliu:
The CHANGELOG check is supposed to fail as this is purely a CI change.

PR #700, opened 2025-01-10
V0.10.x fix translate semver tags

### LINKED ISSUE #699: SemVer Incompatible with OCI tag spec
Eric noted the docker image tag spec requires

>  The tag must be valid ASCII and can contain lowercase and uppercase letters, digits, underscores, periods, and hyphens.

Our semantic versioning currently uses `major.minor.patch-prerelease+build`. We may need to change the `+` to `_` in order to comply with the tag spec. 

CC @isc-tleavitt @isc-kiyer @isc-eneil @isc-jili @isc-jlechtne

isc-shuliu:
Fix #699 

Auto convert "+" to "_" when publishing to an ORAS repo. Convert them back when displaying SemVer.

### DIFF
--- a/src/inc/IPM/Common.inc
+++ b/src/inc/IPM/Common.inc
@@ -67,3 +67,7 @@ ROUTINE %IPM.Common [Type=INC]
 
 #; Variable to mark deprecation warning has been shown in the current process
 #define DeprecationWarned %IPMModuleDeprecatedResource
+
+#; Convert module version from/to OCI tag
+#def1arg Semver2Tag(%semver) $Replace(%semver,"+","_")
+#def1arg Tag2Semver(%tag) $Replace(%tag,"_","+")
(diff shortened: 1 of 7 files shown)


isc-jili (approved):

src/inc/IPM/Common.inc
```diff
@@ -66,4 +66,8 @@ ROUTINE %IPM.Common [Type=INC]
 #def1arg ENDTAG ##Expression($$$ENDTAGQ)
 
 #; Variable to mark deprecation warning has been shown in the current process
-#define DeprecationWarned %IPMModuleDeprecatedResource
\ No newline at end of file
+#define DeprecationWarned %IPMModuleDeprecatedResource
+
+#; Convert module version from/to OCI tag
+#def1arg Semver2Tag(%semver) $Replace(%semver,"+","_")
+#def1arg Tag2Semver(%tag) $Replace(%tag,"_","+")
```
isc-eneil, line 73: So this is never used because the module.xml version is used for display once the package is pulled and loaded. Is that right? @isc-shuliu

isc-eneil (changes requested):

src/inc/IPM/Common.inc
isc-shuliu: Yes. I only convert tags back to semver in `GetPackageVersionPy`, which is done using equivalent Python code. 

The macro `$$$Tag2Semver` itself isn't used. I'm still keeping it in case we need it in the future.

isc-eneil (approved):

isc-kiyer (approved):

Issue #631, opened 2024-11-25
Add support for performing module-specified namespace or system configuration at installation/load time


isc-eneil:
Per conversations with HSOPS, we'd like to ask for support of pre-packaged installation-time configuration of a module as part of the install/load command (or another command that also encompasses installation/loading). In particular, we imagine modules that contain a CPF file that is identified either by its name or file path would be iris merged programmatically before the Reload lifecycle phase.

This feature would not include user-specific installation-time configuration. That is a separate issue.

isc-tleavitt:
Spec:
You can put a CPF file path in module.xml as a resource (you can have more than one)
We'll add a pre-reload lifecycle phase (named "Initialize") and hook into that from the CPF file's resource processor class - default behavior is to do the CPF merge during Initialize, but a resource attribute can be overridden to do this later instead (in a named phase)

Sets of configuration:
Creating namespace + database(s) into which module will be installed (covered in #629)
Module-specific configuration (mappings, creating roles/resources, granting SQL privileges - covered here)
User-specified configuration (OAuth config, UM integration config - equivalent to an "env" file for docker run; this is covered in #632)

Orchestrate this across multiple namespaces: #630

Issue #633, opened 2024-11-26
Support different modes for build and deployment


isc-eneil:
Per conversations with HS, we'd like a way to differentiate between installation during build-time (i.e. loading a module in order to package and maybe publish it) vs. deployment-time (i.e. installing a pre-packaged module on an instance). 

Example use case: 
Certain module configuration should only be performed at deployment time. It may be specified in CPF files or in a class that is in an <Invoke> element in the module.xml. This could include ObjectScript code that should be run before/after an existing lifecycle phase. Many of the same lifecycle phases are run as part of publication and straight installation. 

Some ideas from @isc-kiyer: IPM could add this functionality to the Invoke argument, Scope (currently can be empty, test or verify), and change it to a comma-separated list which could include the mode and perhaps custom values. Alternatively, IPM could create a different Invoke argument, CustomScope, paralleling the CustomPhase argument.

isc-tleavitt:
Ultimate decision on this was a "-build" flag on install/load to differentiate build-time from deployment-time (where the default assumption is deployment-time)

PR #703, opened 2025-01-13
Feat: CPF Merge

isc-shuliu:
Implement #631


src/cls/IPM/ResourceProcessor/CPF.cls
```diff
@@ -0,0 +1,51 @@
...
+    Try {
+        If pPhase '= "Initialize" {
+            Quit
+        }
+
+        If '$Data(pParams("RootDirectory"), root) {
+            $$$ThrowStatus($$$ERROR($$$GeneralError, "RootDirectory unspecified"))
+        }
+        // Use Construct first, rather than NormalizeFilename, so we don't have to deal with leading/trailing slashes
+        Set filename = ##class(%File).Construct(root, ..Path)
```
isc-kiyer, line 29: @isc-shuliu  I would recommend perhaps having all cpf files under a top level `/cpf` folder by default (similar to how classes under /cls, inc files under /inc etc.). Then `Directory` can optionally override it if needed. The resource name should be the file name. 
cc @isc-eneil @isc-jlechtne @isc-jili @isc-tleavitt thoughts?

src/cls/IPM/ResourceProcessor/CPF.cls
```diff
@@ -0,0 +1,51 @@
...
+        }
+        // Use Construct first, rather than NormalizeFilename, so we don't have to deal with leading/trailing slashes
+        Set filename = ##class(%File).Construct(root, ..Path)  
+        Set filename = ##class(%File).NormalizeFilename(filename)
+
+        Set stream = ##class(%Stream.FileCharacter).%New()  
+        $$$ThrowOnError(stream.LinkToFile(filename))
+        If verbose {
+            Write !, "Merging CPF file: ", filename, !
+            While 'stream.AtEnd {
```
isc-kiyer, line 36: Nit: For this, can instead call OutputToDevice()

tests/integration_tests/Test/PM/Integration/CPFMerge.cls
```diff
@@ -0,0 +1,16 @@
+Class Test.PM.Integration.CPFMerge Extends Test.PM.Integration.Base
+{
+
+Parameter TargetModuleName As STRING = "cpf-merge";
+
+Method TestCPFMerge()
```
isc-kiyer, line 6: I like this test driven development!

tests/integration_tests/Test/PM/Integration/_data/cpf-merge/Merge.cpf
```diff
@@ -0,0 +1,3 @@
+[config]
```
isc-kiyer, line 1: Would recommend testing things like package mappings, roles, resources, web apps etc.

src/cls/IPM/ResourceProcessor/CPF.cls
```diff
@@ -0,0 +1,51 @@
...
+                Write stream.Read()
+            }
+        }
+        // TODO actually merge the CPF file
+        // ...
+
+        Set pResourceHandled = 1
+    } Catch ex {
+        Set pResourceHandled = 0
```
isc-kiyer, line 45: For error handling, should see if the error is helpful if the user doesn't have the appropriate permissions and/or doesn't have %System_Callout privileges. Not sure how easy it is to have tests as different users with different privileges but worth adding if possible

isc-kiyer (commented):
@isc-shuliu I know this is still a draft but was curious so took a look and left a few comments

isc-tleavitt (changes requested):

tests/integration_tests/Test/PM/Integration/CPFMerge.cls
```diff
@@ -0,0 +1,16 @@
...
+{
+    Set tModuleDir = ..GetModuleDir(..#TargetModuleName)
+
+    Set tSC = ##class(%IPM.Main).Shell("load -v "_tModuleDir)
+    Do $$$AssertStatusOK(tSC, "Loaded module successfully")
+
+    // TODO - Add more tests here
```
isc-tleavitt, line 13: And assert that the contents of the CPF merge file were actually applied using other Config.* APIs in %SYS

src/cls/IPM/ResourceProcessor/CPF.cls
```diff
@@ -0,0 +1,51 @@
...
+
+        Set stream = ##class(%Stream.FileCharacter).%New()  
+        $$$ThrowOnError(stream.LinkToFile(filename))
+        If verbose {
+            Write !, "Merging CPF file: ", filename, !
+            While 'stream.AtEnd {
+                Write stream.Read()
+            }
+        }
+        // TODO actually merge the CPF file
```
isc-tleavitt, line 40: Found this - Config.CPF:Merge

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -501,6 +503,24 @@ Method %Unconfigure(ByRef pParams) As %Status
...
 
+Method %Initialize(ByRef pParams) As %Status
+{
+	Set tSC = $$$OK
+	Try {
+		Set tKey = ""
+		For {
+			Set tResource = ..Module.Resources.GetNext(.tKey)
+			Quit:tKey=""
+			If $IsObject(tResource.Processor) && ($CLASSNAME(tResource.Processor) = "%IPM.ResourceProcessor.CPF") {
```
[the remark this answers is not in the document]
isc-shuliu: It turns out removing the preloading step in `LoadNewModule()` caused `IPM.Installer` to not be loaded in `verify` phase of CI. Will investigate further.

tests/integration_tests/Test/PM/Integration/_data/cpf-merge/Merge.cpf
isc-shuliu: I tried both `##class(Config.CPF).Merge()` and `$zf(-100)`. Neither worked for the sample 
```ini
[Actions]
CreateRole:Name=Shuheng_Role
```
I can only get it working by running `iris merge IRIS /home/irisowner/iris-merge.cpf` outside of the IRIS shell. 

When calling `##class(Config.CPF).Merge()`, `messages.log` shows
```
01/15/25-20:21:48:401 (646) 0 [Utility.Event] Merging /home/irisowner/iris-merge.cpf into /usr/irissys/iris.cpf
01/15/25-20:21:48:406 (646) 0 [Utility.Event] Validating CPF file /usr/irissys/iris.cpf_Merge
01/15/25-20:21:48:588 (646) 0 [Utility.Event] No errors found in CPF file /usr/irissys/iris.cpf_Merge
01/15/25-20:21:48:588 (646) 0 [Utility.Event] Reading [Actions] section
01/15/25-20:21:48:588 (646) 0 [Utility.Event] Validating [Actions] section
```

When running `iris merge IRIS /home/irisowner/iris-merge.cpf` outside of the IRIS shell, `messages.log` shows more content
```
01/15/25-20:23:58:116 (663) 0 [Utility.Event] Merging /home/irisowner/iris-merge.cpf into /usr/irissys/iris.cpf
01/15/25-20:23:58:122 (663) 0 [Utility.Event] Validating CPF file /usr/irissys/iris.cpf_Merge
01/15/25-20:23:58:299 (663) 0 [Utility.Event] No errors found in CPF file /usr/irissys/iris.cpf_Merge
01/15/25-20:23:58:299 (663) 0 [Utility.Event] Reading [Actions] section
01/15/25-20:23:58:299 (663) 0 [Utility.Event] Validating [Actions] section
01/15/25-20:23:58:474 (663) 0 [Utility.Event] Reading [Actions] section
01/15/25-20:23:58:475 (663) 0 [Utility.Event] Validating [Actions] section
01/15/25-20:23:58:475 (663) 0 [Utility.Event] Executing [Actions] section
01/15/25-20:23:58:497 (664) 0 [Utility.Event] Purging old application errors
01/15/25-20:23:58:497 (663) 0 [Utility.Event] Activating Network
01/15/25-20:23:58:561 (663) 0 [Utility.Event] Activating new namespace map
01/15/25-20:23:58:589 (663) 0 [Utility.Event] Namespace changes have been activated
01/15/25-20:23:58:609 (666) 0 [Utility.Event] Purging old application errors
01/15/25-20:23:58:609 (663) 0 [Utility.Event] Activating Network
01/15/25-20:23:58:671 (663) 0 [Utility.Event] Activating new namespace map
01/15/25-20:23:58:685 (663) 0 [Utility.Event] Namespace changes have been activated
01/15/25-20:23:58:705 (668) 0 [Utility.Event] Purging old application errors
01/15/25-20:23:58:705 (663) 0 [Utility.Event] Activating Network
01/15/25-20:23:58:767 (663) 0 [Utility.Event] Activating new namespace map
01/15/25-20:23:58:795 (663) 0 [Utility.Event] Namespace changes have been activated
```
It appears something resulted in an silent error after `Validating [Actions] section`

isc-shuliu: Worked out how to use $zf(-100) with @isc-eneil. Somehow, the STDOUT must be set in order for $zf(-100) to work. 

As a future task, we need to investigate why `##class(Config.CPF).Merge()` doesn't work.  (Probably need to dig into the Config.CPF source code)

isc-eneil:
HSDevOps was discussing today and along the lines of https://github.com/intersystems/ipm/issues/633, we want to make sure that it is possible for HS to merge CPF resources only during deployment (and not during build-time). In both cases, we call zpm "load" or zpm "install", which means in both cases the initialize phase would run and merge the file.

We had spoken about a CustomScope attribute in #633 but another option is to support just running the merge during a CustomPhase. @isc-shuliu @isc-tleavitt Thoughts on supporting this instead of adding the initialize lifecycle phase? And my apologies for not thinking of this before you implemented it, Shuheng!
cc @isc-kiyer @isc-jlechtne @isc-jili

isc-shuliu:
@isc-eneil 
In that case, shall we do this
* `<CPF Name="merge.cpf">` will by default merge the CPF during the new `Initialize` phase, 
* `<CPF Name="merge.cpf" Phase="MakeDeployed" When="Before/After">` will merge the CPF before/after the `MakeDeployed` phase.
* `<CPF Name="merge.cpf" CustomPhase="MyPhaseName">` will merge the CPF during the `MyPhaseName` custom phase

isc-tleavitt (approved):
Move of loading /preload in initialize instead of reload should be noted as a possible compatibility-breaking change. But I think this looks good otherwise.

isc-shuliu:
> Move of loading /preload in initialize instead of reload should be noted as a possible compatibility-breaking change (really would be an edge case). But I think this looks good otherwise.

@isc-tleavitt When would it break compatibility? Is it only when running `<package> reload -only`?

src/cls/IPM/DataType/ResourceDirectory.cls
```diff
@@ -0,0 +1,24 @@
...
+
+Parameter MAXLEN = 255;
+
+/// Tests if the logical value <var>%val</var>, which is a string, is valid.
+/// The validation is based on the class parameter settings used for the class attribute this data type is associated with.
+/// In this case, <a href="#MINLEN">MINLEN</a>, <a href="#MAXLEN">MAXLEN</a>, <a href="#VALUELIST">VALUELIST</a>, and <a href="#PATTERN">PATTERN</a>.
+ClassMethod IsValid(%val As %RawString) As %Status [ ServerOnly = 0 ]
+{
+    If $Extract(%val) = "/" {
+        Return $$$ERROR($$$GeneralError, "Resource directory cannot start with a slash.")
```
isc-kiyer, line 12: Why can't it start with a slash? This would be the case for an unix platforms right? I think just check for existence of the directory. Regarding checking `..`, I think that check is fine.

src/cls/IPM/ResourceProcessor/CPF.cls
```diff
@@ -0,0 +1,90 @@
...
+        Set root = ..ResourceReference.Module.Root
+        Set sourcesRoot = ..ResourceReference.Module.SourcesRoot
+        // Use Construct first, rather than NormalizeFilename, so we don't have to deal with leading/trailing slashes
+        Set dir = $SELECT($$$isWINDOWS: $REPLACE(..Directory, "/", "\"), 1: ..Directory)
+        Set dir = ##class(%File).Construct(root, sourcesRoot, dir)
+        Set filename = ##class(%File).NormalizeFilename(..Name, dir)
+        If filename = "" {
+            $$$ThrowStatus($$$ERROR($$$GeneralError, $$$FormatText("CPF file '%1' not found in directory '%2'", ..Name, dir)))
+        } ElseIf '##class(%File).Exists(filename) {
+            $$$ThrowStatus($$$ERROR($$$GeneralError, $$$FormatText("CPF file '%1' not found", filename)))
```
isc-kiyer, line 56: Probably worthwhile having the same error message to indicate which directory the code looked at for the cpf file

tests/integration_tests/Test/PM/Integration/_data/cpf-merge/module.xml
```diff
@@ -0,0 +1,24 @@
...
+      <Name>cpf-merge</Name>
+      <Version>0.0.1</Version>
+      <Packaging>module</Packaging>
+      <SourcesRoot>src</SourcesRoot>
+
+      <!-- By default, the CPF merge happens right before Initialize -->
+      <CPF Name="config-globals.cpf"/>
+
+      <!-- Optionally, make it happen after MakeDeployed -->
+      <CPF Name="actions.cpf" Directory="custom-cpf-directory" Phase="MakeDeployed" When="After"/>
```
isc-kiyer, line 14: @isc-shuliu  Does mapping of this structure to the structure below happen automatically using XSLTs?
```
<CPF Name="actions.cpf">
<Attribute Name="Directory">custom-cpf-directory</Attribute>
<Attribute Name="Phase">MakeDeployed</Attribute>
<Attribute Name="When">After</Attribute>
</CPF>
```

I ask because its a bit confusing distinguishing between what is a valid property on ResourceReference vs in the ResourceProcessor class

cc @isc-tleavitt

isc-kiyer (changes requested):
@isc-shuliu Looks great! Few questions/comments

isc-eneil (changes requested):

isc-eneil:
@isc-shuliu I left just a couple of comments but it looks good to me too!

src/cls/IPM/Storage/Module.cls
```diff
@@ -177,11 +177,23 @@ Method GetCustomPhases(Output pPhases)
...
             Set pPhases($$$lcase(tInvoke.CustomPhase)) = tInvoke.CustomPhase
         }
     }
+	For {
+		Set tResource = ..Resources.GetNext(.key)
+		If key = "" {
+			Quit
+		}
+		Set tProcessor = tResource.Processor
+		If ($ClassName(tProcessor) = "%IPM.ResourceProcessor.CPF") && (tProcessor.CustomPhase '= "") {
```
isc-kiyer, line 193: Do we need to specifically do this just for CPF? I think perhaps check if the processor class has CustomPhase and if so, check its value

src/cls/IPM/ResourceProcessor/Abstract.cls
```diff
@@ -192,6 +192,13 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
 	Quit $$$OK
 }
 
+/// Some processors may need to do something custom when a custom phase is executed.
+/// Such as the CPF processor, which optionally merges a CPF file during a custom phase.
+Method OnCustomPhase(pCustomPhase As %String, ByRef pParams) As %Status
```
isc-kiyer, line 197: I like this abstraction! However, I think we may want to put this in a another intermediate subclass for non-document processor classes. Document processors always execute during regular phases since they don't do things in just one phase (need to do things across reload and compile at a minimum)

src/cls/IPM/Storage/Module.cls
```diff
@@ -350,7 +362,20 @@ ClassMethod ExecutePhases(pModuleName As %String, pPhases As %List, pIsComplete
         }
         Quit:$$$ISERR(tSC)
 
-        If 'tIsCustomPhase {
+        If tIsCustomPhase {
+			// Handle CPF merges in custom phases
```
isc-kiyer, line 366: Nit: Remove this comment since it isn't specific to CPF merges necessarily.
Also, in a few of these files, the spacing seems off. Not sure if that has to do with github though.

isc-kiyer (changes requested):
@isc-shuliu few more small comments/thoughts

src/cls/IPM/ResourceProcessor/CPF.cls
```diff
@@ -0,0 +1,96 @@
...
+Property Directory As %IPM.DataType.ResourceDirectory [ InitialExpression = "cpf" ];
+
+/// FileN ame of the CPF merge file
+Property Name As %IPM.DataType.ResourceName [ Required ];
+
+/// The phase before/after which the CPF file should be merged
+Property Phase As %IPM.DataType.PhaseName [ InitialExpression = "Initialize" ];
+
+/// If provided, the Phase property will be ignored. The CPF merge will happen during this custom phase.
+Property CustomPhase As %IPM.DataType.CustomPhaseName;
```
isc-kiyer, line 22: Nit: No need to have CustomPhase property here since it is inherited

isc-kiyer (approved):
@isc-shuliu one very minor comment. Approving

isc-kiyer (approved):

src/cls/IPM/Storage/Module.cls
```diff
@@ -177,11 +177,27 @@ Method GetCustomPhases(Output pPhases)
...
+		Set tResource = ..Resources.GetNext(.key)
+		If key = "" {
+			Quit
+		}
+		Set tProcessor = tResource.Processor
+		// %IsA() only returns true if the class is the "primary" superclass, while %Extends() works for other superclasses (such as mixins).
+		If $IsObject(tProcessor) && tProcessor.%Extends("%IPM.ResourceProcessor.CustomPhaseMixin") {
+			Set cp = tProcessor.CustomPhase
+			If cp '= "" {
+				Set pPhases($$$lcase(cp)) = cp
```
isc-eneil, line 197: @isc-shuliu Apologies for the late review on this.

Should there be an additional check to make sure the custom phase hasn't already been added (since an Invoke and a CPF resource could technically give the same customer phase)?

cc @isc-kiyer

Issue #653, opened 2024-12-10
Add a new command to update a module's version (common use before publishing)


isc-eneil:
As discussed with HSDevOps, we would like to be able to update the version of the module to include info that is required for artifact traceability before publication. For example, we'd always like to append the build number to the module version, and for dev builds, the project branch and user who triggered the build, if applicable.

This should behave similar to npm's version command: https://docs.npmjs.com/cli/v8/commands/npm-version

PR #710, opened 2025-01-22
Add a `module-version` command that updates the version of a module

isc-shuliu:
Implement #653 

TODO
- [x] Investigate whether we need to update module stream/module.xml as well
- [x] Add integration tests


src/cls/IPM/Main.cls
```diff
@@ -714,6 +714,32 @@ generate /my/path -export 00000,PacketName2,IgnorePacket2^00000,PacketName3,Igno
 	</example>
 </command>
 
+<command name="module-version" aliases="modver">
+	<description>
+		Updates the version of the module in the current namespace.
+	</description>
+	<parameter name="module" description="Name of the module to update the version for." />
```
isc-kiyer, line 721: Module name should be required

src/cls/IPM/Main.cls
```diff
@@ -714,6 +714,32 @@ generate /my/path -export 00000,PacketName2,IgnorePacket2^00000,PacketName3,Igno
...
+	<parameter name="version" description="Either a valid SemVer string, or 'major', 'minor', 'patch'. If it's 'major', 'minor', or 'patch', will bump the cooresponding part of SemVer." />
```
isc-kiyer, line 722: Instead of overloading this, I think if version provided, it should be used as is. If not, then provide one of `-major, -minor, -patch, -pre-release or -build`. With the last 2, need to provide a value as you have below

src/cls/IPM/Main.cls
```diff
@@ -3258,6 +3286,83 @@ ClassMethod UnmapIPM(ByRef pCommandInfo)
...
+		$$$ThrowOnError($$$ERROR($$$GeneralError,"Module "_modName_" not found."))
+	}
+	Set stream = ##class(%IPM.StudioDocument.ModuleStream).NameOpen(modName)
+	If stream = "" {
+		$$$ThrowOnError($$$ERROR($$$GeneralError,"Module stream "_modName_" not found."))
+	}
+	If 'quiet {
+		Write !,$$$FormatText("Current version of module %1 is: %2", modName, mod.VersionString)
+	}
+
```
isc-kiyer, line 3308: Should check if one of major, minor or patch is provided. Cannot provide multiple together. To be done once they are modifiers

src/cls/IPM/Main.cls
```diff
@@ -3258,6 +3286,83 @@ ClassMethod UnmapIPM(ByRef pCommandInfo)
...
+	If $$$HasModifier(pCommandInfo,"prerelease") {
+		Set semver = semver.WithPrerelease($$$GetModifier(pCommandInfo,"prerelease"))
+	}
+	If $$$HasModifier(pCommandInfo,"build") {
+		Set semver = semver.WithBuild($$$GetModifier(pCommandInfo,"build"))
+	}
+
+	If 'quiet {
```
isc-kiyer, line 3335: This should be below the two if blocks below since the version isn't actually being set until those checks pass

src/cls/IPM/Main.cls
```diff
@@ -3258,6 +3286,83 @@ ClassMethod UnmapIPM(ByRef pCommandInfo)
...
+		Return
+	}
+
+	Set initTLevel = $TLevel
+	TSTART
+	Try {
+		Set mod.Version = semver
+		Set mod.VersionString = semver.ToString()
+		$$$ThrowOnError(mod.%Save())
+		Do stream.UpdateVersion(semver)
```
isc-kiyer, line 3356: @isc-tleavitt how will this integrate with source control hooks to ensure the local file system version is also updated?

isc-kiyer (changes requested):
@isc-shuliu few small comments

src/cls/IPM/General/SemanticVersion.cls
```diff
@@ -206,6 +206,68 @@ Method Satisfies(pExpression As %IPM.General.SemanticVersionExpression) As %Bool
...
+	Set tVer = ..%New()
+	Set tVer.Major = ..Major
+	Set tVer.Minor = ..Minor
+	Set tVer.Patch = ..Patch
+	Set tVer.Prerelease = ..Prerelease
+	Set tVer.Build = pBuild
+	Quit tVer
+}
+
+Method IsLowerThan(pVersion As %IPM.General.SemanticVersion) As %Boolean
```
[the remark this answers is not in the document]
isc-shuliu: So `Follows` have this weird restraint that when you have pre-release info, 2 semvers with different major.minor.patch cannot be compared. In other words, if
```objectscript
set u = ##class(%IPM.General.SemanticVersion).FromString("1.2.3-beta")
set v = ##class(%IPM.General.SemanticVersion).FromString("4.5.6-beta")
```
then both `u.Follows(v)` and `v.Follows(u)` will be false. So In our code, 
* if we assert `'oldVer.Follows(newVer)` before proceeding with version updates, then we would allow updating from `4.5.6-beta` to `1.2.3-beta` without requiring `-permit-downgrade`.
* if we assert `newVer.Follows(oldVer)` before proceeding with version updates, the we would disallow updating from `1.2.3-beta` to `4.5.6-beta` unless `-permit-downgrade` is set.

Neither seems a very good option, that's why I wrote this method.

src/cls/IPM/Main.cls
isc-shuliu: If a version string `1.2.3` is provided, then we shouldn't bump `major`, `minor`, or `patch` any further, and vice versa. Also, at most one of {`major`, `minor`, `patch`} should can be set at a time. So the { version string, `major`, `minor`, `patch` } are 4 mutually exclusive values and that's why I decided to overload the `version` parameter. Separating them requires more complicated validation logic and is more error prone.

On the other hand, `prerelease` and `build` are safe to use with any of the 4, because a user might want to set a  semver `1.2.3` and add info to it, such as `1.2.3-snapshot+buildhash`, or bump a major version and add some info, such as `2.0.0-snapshot`.

src/cls/IPM/Main.cls
isc-shuliu: I just added functionality and tests for updating module.xml on the filesystem

isc-kiyer: @isc-shuliu thanks but I'm not sure we want IPM directly modifying local files. Likely should call out to source control hooks. For example, if this is used in perforce, need to check out the file before editing it.

src/cls/IPM/Main.cls
isc-kiyer: Ah fair enough. I like this reasoning.

src/cls/IPM/General/SemanticVersion.cls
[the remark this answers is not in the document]
isc-kiyer: @isc-shuliu Oh then that's a bug in Follows. It should allow comparing pre-release versions. Per the semver spec, 1.2.3-beta < 4.5.6-beta since comparison of versions should first be done on major, then minor, then patch and finally pre-release only if major.minor.patch is the same. At least that is my understanding on re-reading the spec: https://semver.org/
@isc-tleavitt Please confirm.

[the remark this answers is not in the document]
isc-tleavitt: @isc-kiyer @isc-shuliu the intent is for Follows to support semantic version expressions, so this behavior supports (from the link Keshav sent):
"For example, the range >1.2.3-alpha.3 would be allowed to match the version 1.2.3-alpha.7, but it would not be satisfied by 3.4.5-alpha.9, even though 3.4.5-alpha.9 is technically "greater than" 1.2.3-alpha.3 according to the SemVer sort rules."

The point about semantic versions vs. semantic version expressions is fair, but really the only use case prior to this for comparing semantic versions is to evaluate a semantic version expression.

In the case of prerelease designations specifically, "downgrade" isn't a great term, but our default behavior is correct and has parity with node-semver.

src/cls/IPM/StudioDocument/ModuleStream.cls
```diff
@@ -1,8 +1,59 @@
...
+Method UpdateVersion(semver As %IPM.General.SemanticVersion)
+{
+	// Can't use %IPM.Repo.XSLTProvider because we need to pass a parameter
+	Set compiledTransfrom = ..CompiledUpdateVersionXSLT(semver)
+
+	$$$ThrowOnError(..Contents.Rewind())
+	$$$ThrowOnError(##class(%XML.XSLT.Transformer).TransformStreamWithCompiledXSL(..Contents, compiledTransfrom, .updatedContents))
+	$$$ThrowOnError(..Contents.Clear())
+	$$$ThrowOnError(..Contents.CopyFrom(updatedContents))
+	$$$ThrowOnError(..%Save())
```
isc-kiyer, line 54: Rather than saving the stream and persistence independently, should you perhaps use `##class(%IPM.StudioDocument.Module).UpdatePersistentFromStream()`? I see this calls out to the source control hooks method `##class(%IPM.ExtensionBase.SourceControl.Interface).%OnModifyResources(pModule)`. I guess I'm not sure where the source control hooks are invoked when a studio document changes. @isc-tleavitt do you know?

isc-shuliu: This is handled in Tim's commit [32f1b2f](https://github.com/intersystems/ipm/pull/710/commits/32f1b2f2039c78ef3739f34dcd30d5124b64e204), which calls %Library.RoutineMgr to update the stream and calls the source control hooks.

src/cls/IPM/StudioDocument/Module.cls
```diff
@@ -344,6 +344,65 @@ Method ImportFromXML(stream As %RegisteredObject, flags As %String) As %Status
...
+	$$$ThrowOnError(tSC)
+
+	Quit compiledStyleSheet
+}
+
+/// Updates the version of <var>moduleName</var> to <var>semver</var>
+ClassMethod UpdateVersion(moduleName As %String, semver As %IPM.General.SemanticVersion)
+{
+	// Can't use %IPM.Repo.XSLTProvider because we need to pass a parameter
+	Set compiledTransfrom = ..CompiledUpdateVersionXSLT(semver)
```
isc-eneil, line 392: Nit: should this be compiledTrans**form**?

isc-eneil (approved):

isc-tleavitt (approved):
I think this is good after the typo fix. @isc-shuliu and I had a pair programming session today on the source control integration and have confirmed it working with the Perforce extension, although we found an issue there due to a changed return type - should be fixed in the Perforce extension itself.

PR #714, opened 2025-01-23
fix(ci): bump version of upload/donwload artifact to v4

isc-shuliu:
Bump version of `actions/upload-artifact` and `actions/download-artifact` to v4

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -40,8 +40,8 @@ jobs:
           done;
           echo images="[${images%?}]" >> $GITHUB_OUTPUT
 
-          // The syntax GitHub expects is `main=string`, rather than `main="string"`
-          // I think this is way too confusing and makes no sense.
-          // Even for bash, `main=string` and `main="string"` should be equivalent.
+          # The syntax GitHub expects is `main=string`, rather than `main="string"`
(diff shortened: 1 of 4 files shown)


isc-shuliu:
It seems that the syntax for the `if` is causing some problems.

On one hand
```yaml
uses: actions/upload-artifact@v4
if: matrix.image == ${{ needs.prepare.outputs.main }}
```
causes all matrix jobs to run the upload-artifact step, leading to a name conflict.

On the other hand
```yaml
uses: actions/upload-artifact@v4
if: ${{ matrix.image ==  needs.prepare.outputs.main }}
```
causes all matrix jobs to skip this step, leading to another problem during download-artifact

Will investigate futher

isc-shuliu:
The CI files should be in good shape now. 

Now we're dealing with a failed [integration test](https://github.com/intersystems/ipm/actions/runs/12938280043/job/36088888044) on `irishealth-community:2023.1`. I'm unable to reproduce it on a container of the same tag locally. Might need to enable more verbose output during tests to investigate.

isc-shuliu:
And it just magically recovered without any change. Will leave the debugging code in there in case it happens again in the future.

PR #716, opened 2025-01-24
V0.10.x feat publish external name

isc-shuliu:
Added support to publish under external name by passing `-use-external-name` or `-use-ext`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #609 Added support for `-export-deps` when running the "Package" phase of lifecycle
 - #541 Added support for ORAS repository
 - #704 Added support for passing in env files via `-env /path/to/env1.json;/path/to/env2.json` syntax
+- #716 Added support to publish under external name by passing `-use-external-name` or `-use-ext`. 
 
 ### Changed
 - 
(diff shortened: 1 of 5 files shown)


isc-shuliu:
@isc-tleavitt @isc-kiyer @isc-eneil Do we need to update the published module.xml as well?

isc-eneil:
> @isc-tleavitt @isc-kiyer @isc-eneil Do we need to update the published module.xml as well?

I don't believe so. I think it is sufficient for the module.xml to keep the same <Name> and <ExternalName> tags.

isc-shuliu:
> > @isc-tleavitt @isc-kiyer @isc-eneil Do we need to update the published module.xml as well?
> 
> I don't believe so. I think it is sufficient for the module.xml to keep the same and tags.

In that case, when a user installs a package by its external name, it will probably still show as the original name in `list-modules`. If that's the expected behavior, I'll mark this PR ready for review

isc-tleavitt:
> > > @isc-tleavitt @isc-kiyer @isc-eneil Do we need to update the published module.xml as well?
> > 
> > 
> > I don't believe so. I think it is sufficient for the module.xml to keep the same and tags.
> 
> In that case, when a user installs a package by its external name, it will probably still show as the original name in `list-modules`. If that's the expected behavior, I'll mark this PR ready for review

This is expected behavior, to be addressed via zpm "info" in subsequent issues

isc-tleavitt (commented):
@isc-eneil @isc-kiyer can you confirm this does what you need / the specific use cases and targets you have in mind?

There's a comment in the issue that doesn't seem to be addressed here:
"Each repo type also likely has different naming requirements so maybe need a method in the IPublishService interface for determining valid package name."

isc-kiyer:
> @isc-eneil @isc-kiyer can you confirm this does what you need / the specific use cases and targets you have in mind?
> 
> There's a comment in the issue that doesn't seem to be addressed here: "Each repo type also likely has different naming requirements so maybe need a method in the IPublishService interface for determining valid package name."

@isc-tleavitt @isc-shuliu Yes so I can elaborate on what I meant by that comment. Currently for publishing, we check the name and version based on semantic versioning requirements + IPM module naming conventions. However, based on the repo type, requirements may be slightly different (for example, ORAS and perforce have different requirements). So worthwhile having methods in IPublishService for IsValidPackageName() and IsValidPackageVersion()
cc @isc-eneil

isc-kiyer (approved):

PR #726, opened 2025-02-03
Load dependencies from `.modules` subfolder

### LINKED ISSUE #721: Load tarball - IPM look for dependency from repo instead
Environment: 
IPM v0.10.beta-4 on an I4H instance with no repo configured.

Background:
The module has dependencies. The tarball includes all the dependency modules.

Behavior:  
When I do zpm "load <tarball>", it stopped at the first dependency hs.rest, "Could not find satisfactory version of hs.rest in any repositories. Required by: healthshare.hp.pas: ^2.1.0". I think it is looking for the dependency from the repository, not from the tarball.

In contrast, on a instance with a repo configured, when I load the tarball, IPM would install the dependencies from the repo, not from the tarball.

@isc-shuliu

isc-shuliu:
When finished, this will fix #721

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -139,6 +139,7 @@ This command is an alias for `module-action module-name package`
 <modifier name="quiet" aliases="q" dataAlias="Verbose" dataValue="0" description="Produces minimal output from the command." />
 <modifier name="verbose" aliases="v" dataAlias="Verbose" dataValue="1" description="Produces verbose output from the command." />
 <modifier name="path" aliases="p" dataAlias="Path" value="true" description="Use specified path to export package." />
+<modifier name="export-deps" value="true" dataAlias="ExportDependencies" description="If specified, controls whether dependencies are exported. If omitted, defaults to the value of the #EXPORTDEPENDENCIES in lifecycle class." />
 </command>
 
 <command name="verify" dataPrefix="D">
--- a/src/cls/IPM/Storage/Module.cls
+++ b/src/cls/IPM/Storage/Module.cls
@@ -561,6 +561,24 @@ Method LoadDependencies(pPhaseList As %List, ByRef pParams) As %Status
 		// Resolve and validate dependencies
 		Set tReloadSnapshots = +$Get(pParams("UpdateSnapshots"))
 		Set tPermitDowngrade = +$Get(pParams("PermitDowngrade"))
+
+		// When loading a module from a local folder, there might be a <mod root>/.modules/ folder containining dependencies.
+		// It's easier to configure a temporary repository than to handle this case in the dependency resolution code.
+		If $Get(pParams("RootDirectory")) '= "" {
+			Set dotModules = ##class(%File).NormalizeDirectory(".modules", pParams("RootDirectory"))
+			If ##class(%File).DirectoryExists(dotModules) {
+				Set repoNameBase = "ipm-temp-dot-modules-"
+				Set count = 1
+				For {
+					Set repoName = repoNameBase _ count
+					If '##class(%IPM.Repo.Definition).ServerDefinitionKeyExists(repoName) {
+						Quit
+					}
+				}
+				Do ##class(%IPM.Main).Shell("repo -fs -name "_repoName_" -path "_dotModules_" -snapshots 1 -prereleases 1")
+			}
+		}
+
 		Set tSC = ..BuildDependencyGraph(.tDependencyGraph,,tReloadSnapshots,.qstruct,pPhaseList,,,,,,tPermitDowngrade)
 		If $$$ISERR(tSC) {
 			Quit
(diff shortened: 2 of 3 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -561,6 +561,24 @@ Method LoadDependencies(pPhaseList As %List, ByRef pParams) As %Status
 		// Resolve and validate dependencies
 		Set tReloadSnapshots = +$Get(pParams("UpdateSnapshots"))
 		Set tPermitDowngrade = +$Get(pParams("PermitDowngrade"))
+
+		// When loading a module from a local folder, there might be a <mod root>/.modules/ folder containining dependencies.
+		// It's easier to configure a temporary repository than to handle this case in the dependency resolution code.
+		If $Get(pParams("RootDirectory")) '= "" {
```
isc-kiyer, line 567: I like this idea. Some notes: 
- would be good to have this repo created as early as possible
- this repo should be the first in the list of repos to resolve against I believe? Or at least provide a way to signify its priority?
- should we clean up this repo? I think we probably should at the end of the load command. Since this could only happen from a load command, the creation and clean up of this repo could be part of load command in Main()?

src/cls/IPM/Repo/Definition.cls
```diff
@@ -31,6 +31,11 @@ Property Snapshots As %Boolean [ InitialExpression = 0 ];
 
 Property Prereleases As %Boolean [ InitialExpression = 0 ];
 
+/// If set to a non-empty value, this will be used as the sort order for this repository, bypassing the GetSortOrder() method.
+/// This should be only configured by IPM during certain processes (e.g., installing depdenencies from .modules directory).
+/// There is no current plan to support configuring this from the package manager shell.
+Property OveriddenSortOrder As %Integer;
```
isc-kiyer, line 37: Nit: Would perhaps mark this as Internal then

src/cls/IPM/Storage/Module.cls
```diff
@@ -561,6 +561,7 @@ Method LoadDependencies(pPhaseList As %List, ByRef pParams) As %Status
 		// Resolve and validate dependencies
 		Set tReloadSnapshots = +$Get(pParams("UpdateSnapshots"))
 		Set tPermitDowngrade = +$Get(pParams("PermitDowngrade"))
+
```
isc-kiyer, line 564: Nit: this file can be removed from review

src/cls/IPM/Utils/Module.cls
```diff
@@ -1206,8 +1206,9 @@ ClassMethod LoadNewModule(pDirectory As %String, ByRef pParams, pRepository As %
 		}
 		Merge $$$ZPMHandledModules($Namespace) = pParams("Multicompile","ModuleContext")
 
-    $$$ThrowOnError(##class(%IPM.Storage.Module).CheckSystemRequirements(tModuleName))
-  
+		$$$ThrowOnError(##class(%IPM.Storage.Module).CheckSystemRequirements(tModuleName))
+		Set pParams("RootDirectory") = pDirectory
```
isc-kiyer, line 1210: What is the purpose of setting RootDirectory here instead of below? That seems to be the only functional change so if its not needed, changes to this file can be reverted as well

isc-kiyer (commented):
@isc-shuliu few very minor comments. Looks great! We had discussed during the call today that this could also be supported for installing from a repo using the install command but in that case the local .modules directory would be used last in the ordering instead of first. Do you want to implement that in a separate review or this one?

isc-kiyer (approved):

src/cls/IPM/Main.cls
```diff
@@ -2088,23 +2088,48 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
...
+		Set tempRepo.Name = repoName
+		Set tempRepo.Root = dotModules
+		Set tempRepo.Snapshots = 1
+		Set tempRepo.Prereleases = 1
+		Set tempRepo.OverriddenSortOrder = -1000 // Make sure this is the first repo to be found by SQL query in %IPM.Repo.Manager:SearchRepositoriesForModule
+		$$$ThrowOnError(tempRepo.BuildCache(1,1,1))
+	}
+	Set tSC = ##class(%IPM.Utils.Module).LoadNewModule(tTargetDirectory, .tParams)
+	If $Data(tempRepo) # 2 {
+		Set tSC = $$$ADDSC(tSC, tempRepo.%DeleteId(tempRepo.%Id()))
```
isc-tleavitt, line 2130: Nitpick: this should be in a pseudo-"Finally" block so that if BuildCache fails we still clean up. We can make that a followup given how critical this is, though.

isc-tleavitt (approved):

src/cls/IPM/Main.cls
isc-shuliu: I logged an issue for that.

Issue #742, opened 2025-02-18
Add support for passing OS environment variables to the -env JSON file


isc-eneil:
Per discussion with HSDevOps, AppServices, HSOPS, Corporate Security, and HS Payer Services on 2/18/2025, we'd like for IPM to support OS environment variables in the JSON file passed to the `module-action`, `install`, or `load` commands via the `-env` flag. 

Dev Notes:
- At the IRIS level, it is possible to access OS environment variables via `$System.Util.GetEnviron()` 
- Example key-value pairs and syntax:
--  "userName":"${DTR_INTEROP_USERNAME}"
 and "password":"${DTR_INTEROP_PASSWORD}"

cc @isc-kiyer @isc-shuliu @timleavitt @isc-jili @isc-jlechtne

PR #743, opened 2025-02-19
Supporting `${EnvVar}` syntax in `env.json`

isc-shuliu:
Implement #742 

Environment variables can be included in env.json using `{secret: "${varname}"}` syntax. IPM will call $System.Util.GetEnviron() to populate the JSON object accordingly.

### DIFF
--- a/src/cls/IPM/General/EnvironmentConfig.cls
+++ b/src/cls/IPM/General/EnvironmentConfig.cls
@@ -24,6 +24,7 @@ Method Load(path As %String)
             $$$ThrowStatus($$$ERROR($$$GeneralError, "Invalid configuration file. Expected a JSON object, not a JSON array"))
         }
         Do ..MergeDynamicObjects(..Config, newConfig)
+        Set ..Config = ..%Evaluate(..Config)
     } Catch ex {
         Do ..Clear()
         Throw ex
@@ -41,7 +42,7 @@ Method GetArg(package As %String, args... As %String) As %IPM.DataType.Any
         }
         Set obj = obj.%Get(args(i))
     }
-    Return ..%Evaluate(obj)
+    Return obj
 }
 
 /// The input could be a primitive type, a dynamic object, or a dynamic array
--- a/src/cls/IPM/General/EnvironmentConfig.cls
+++ b/src/cls/IPM/General/EnvironmentConfig.cls
@@ -41,7 +41,7 @@ Method GetArg(package As %String, args... As %String) As %IPM.DataType.Any
         }
         Set obj = obj.%Get(args(i))
     }
-    Return %Evaluate(obj)
+    Return ..%Evaluate(obj)
 }
 
 /// Recursively evaluate ${myVarName} to the value of the variable myVarName
--- a/src/cls/IPM/General/EnvironmentConfig.cls
+++ b/src/cls/IPM/General/EnvironmentConfig.cls
@@ -41,6 +41,38 @@ Method GetArg(package As %String, args... As %String) As %IPM.DataType.Any
         }
         Set obj = obj.%Get(args(i))
     }
+    Return %Evaluate(obj)
+}
+
+/// Recursively evaluate ${myVarName} to the value of the variable myVarName
+/// The input could be a primitive type, a dynamic object, or a dynamic array
+/// "CopyCreated" should only be used internally to avoid computation overhead. It should not be used by the any caller other than the recursive call in itself.
+ClassMethod %Evaluate(obj As %IPM.DataType.Any, CopyCreated As %Boolean = 0) As %IPM.DataType.Any [ Internal ]
+{
+    If '$IsObject(obj) {
+        // All numbers, bools, null, etc. will always fail this test. 
+        // Only strings of the form "${myVarName}" will pass
+        If ($Length(obj) > 3) && ($Extract(obj, 1, 2) = "${") && ($Extract(obj, *) = "}") {
+            Set varName = $Extract(obj, 3, *-1)
+            Set varName = $ZStrip(varName, "<>WC")
+            Return $System.Util.GetEnviron(varName)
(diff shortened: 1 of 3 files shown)


src/cls/IPM/General/EnvironmentConfig.cls
```diff
@@ -41,6 +41,38 @@ Method GetArg(package As %String, args... As %String) As %IPM.DataType.Any
         }
         Set obj = obj.%Get(args(i))
     }
+    Return ..%Evaluate(obj)
+}
+
+/// Recursively evaluate ${myVarName} to the value of the variable myVarName
+/// The input could be a primitive type, a dynamic object, or a dynamic array
+/// "CopyCreated" should only be used internally to avoid computation overhead. It should not be used by the any caller other than the recursive call in itself.
```
isc-kiyer, line 49: Nit: Instead of expecting this from the dev, perhaps create an internal private version of the method with the CopyCreated arg and only keep the obj arg in this signature?
Or could make this method itself private since its only intended for the above GetArg (for now)

src/cls/IPM/General/EnvironmentConfig.cls
```diff
@@ -41,6 +41,38 @@ Method GetArg(package As %String, args... As %String) As %IPM.DataType.Any
...
+ClassMethod %Evaluate(obj As %IPM.DataType.Any, CopyCreated As %Boolean = 0) As %IPM.DataType.Any [ Internal ]
+{
+    If '$IsObject(obj) {
+        // All numbers, bools, null, etc. will always fail this test. 
+        // Only strings of the form "${myVarName}" will pass
+        If ($Length(obj) > 3) && ($Extract(obj, 1, 2) = "${") && ($Extract(obj, *) = "}") {
```
isc-kiyer, line 55: Nit: Rather than doing this, evaluation on each get, perhaps could do regex matching on the stream itself and do the replacements once when the file is loaded into memory?

tests/unit_tests/Test/PM/Unit/ParseEnvVarJSON.cls
```diff
@@ -0,0 +1,87 @@
...
+
+Parameter ENVNAME = "mySecretEnvVarName";
+
+Parameter ENVVALUE = "This is a secret value";
+
+XData SampleJSON [ MimeType = application/json ]
+{
+{
+    "package": {
+        "nested": {
```
isc-kiyer, line 12: Worthwhile adding examples with env vars at top level node, nested node, in an array and perhaps multiple env vars in a single value

isc-kiyer (changes requested):
@isc-shuliu some comments/thoughts up

src/cls/IPM/General/EnvironmentConfig.cls
isc-shuliu: Will create a private helper method. I didn't make this method itself private because I wanted to do unit test on this method.

src/cls/IPM/General/EnvironmentConfig.cls
isc-shuliu: I wanted evaluate it for every `GetArg` call because I wasn't sure if environment variables can change during installing/loading packages. If the environment variables won't change, I can move this logic into `Load` so that is only runs once.

isc-kiyer: I don't think these variables should be changing during the process of install. If someone is doing so, that sounds like bad practice to me because if we think about other platforms/tools like docker, env variables are set once and consumed @isc-tleavitt please confirm.

isc-eneil (commented):
@isc-shuliu I took a look and don't have any further substantial comments. Once the changes Keshav suggested are in, you'll have my up-vote!

tests/unit_tests/Test/PM/Unit/ParseEnvVarJSON.cls
```diff
@@ -0,0 +1,87 @@
...
+    Do os.putenv(..#ENVNAME, ..#ENVVALUE)
+}
+
+ClassMethod UnsetEnvironmentVariable()
+{
+    Set os = ##class(%SYS.Python).Import("os")
+    Do os.environ.pop(..#ENVNAME, "")
+}
+
+Method TestarseEnvVarJSON()
```
isc-eneil, line 65: Nit: More descriptive test method name (not quite sure what this one means).

isc-eneil (changes requested):

src/cls/IPM/General/EnvironmentConfig.cls
isc-tleavitt: Confirmed.

isc-tleavitt (approved):

isc-kiyer (approved):

src/cls/IPM/General/EnvironmentConfig.cls
```diff
@@ -44,6 +45,46 @@ Method GetArg(package As %String, args... As %String) As %IPM.DataType.Any
     Return obj
 }
 
+/// The input could be a primitive type, a dynamic object, or a dynamic array
+/// This function wll evaluate ${myVarName} to the value of the environment variable $myVarName
+/// If the environment variable is not set, it will be empty string
+ClassMethod %Evaluate(obj As %IPM.DataType.Any) As %IPM.DataType.Any
+{
+    Return ..%EvaluateHelper(obj, 0)
```
isc-eneil, line 53: Nit: not a merge blocker but rather a coding style suggestion. It seems like the copyCreated arg is only 0 in the first recursive call to %EvaluateHelper. I think it would be a bit more elegant to create a copy of the object in %Evaluate() then pass in the copy to %EvaluateHelper() and eliminate the copyCreated arg.
@isc-shuliu

isc-eneil (approved):

Issue #613, opened 2024-10-31
Additional info in the welcome screen


SCanzano:
When you start ZPM you get something like
```
========================================
|| Welcome to the Package Manager Shell (ZPM). version 0.7.3               ||
|| Enter q/quit to exit the shell. Enter ?/help to view available commands ||
|| Current registry https://registry.readycomputing.com/registry           ||
========================================
```

Often times I get issues reported to me about FAILURES that occur with an install.  
I'm making a recommendation that the above screen show additional information like

- the results of $ZV or if in a HealthShare installation the full HealthShare version string
- whether or not this is the primary mirror member if part of a mirror or if on a backup
- the results of what is found in  `  Set $Namespace = "%SYS" $$$THROWONERROR(tSC, ##class(Config.Startup).Get(.tProperties)) 		Set pSystemMode = $Get(tProperties("SystemMode"))` so that we could see if this is a TEST, STAGE or PROD environment.

- Maybe some information about the %SYS("HealthShare","Instances" types so we can see if this might be an instance where HSANALYTICS, HSPI, the CV etc are currently installed.  This is important because InterSystems does not deliver all of the code to each HealthShare instance but rather only the code that is appropriate for the instance type.  Some of our modules have dependencies on things like HSPI_Data.Patient and can only compile if the instance has HSPI installed.

Having this information might prevent me from having to ask for additional informaiton if we make it easier for people to capture.

isc-kiyer:
All of that sounds like very useful info! However, I think it makes more sense to have it be available when the terminal shell is started (can be done via the ZSTART routine I believe) and doesn't belong on the IPM side

janihur:
Related to the welcome screen. It reports only 
```
Current registry https://pm.community.intersystems.com  
```
Even I have two other active registries:
```
zpm:IPMTEST1>repo -list


local
        Source:                 http://localhost:52773/registry/
        Enabled?                Yes
        Available?              Yes
        Use for Snapshots?      Yes
        Use for Prereleases?    Yes
        Is Read-Only?           No
        Deployment Enabled?     Yes

registry
        Source:                 https://pm.community.intersystems.com
        Enabled?                Yes
        Available?              Yes
        Use for Snapshots?      Yes
        Use for Prereleases?    Yes
        Is Read-Only?           No
        Deployment Enabled?     No

localfs
        Source:                 /home/irisowner/work/objectscript-ex/ipm/*/module.xml
        Enabled?                Yes
        Available?              Yes
[...]

PR #749, opened 2025-02-28
Additional debugging info in welcome banner

isc-shuliu:
Implement #613 

- [x] the results of $ZV or if in a HealthShare installation the full HealthShare version string
- [x] whether or not this is the primary mirror member if part of a mirror or if on a backup
- [x] the results of what is found in   `Set $Namespace = "%SYS" $$$THROWONERROR(tSC, ##class(Config.Startup).Get(.tProperties)) Set pSystemMode = $Get(tProperties("SystemMode"))` so that we could see if this is a TEST, STAGE or PROD environment.
- [x] Maybe some information about the %SYS("HealthShare","Instances" types so we can see if this might be an instance where HSANALYTICS, HSPI, the CV etc are currently installed.

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -823,6 +823,47 @@ ClassMethod GetVersion(ModuleName, ByRef out, list)
 	Quit sc
 }
 
+ClassMethod GetSystemMode()
+{
+	New $Namespace
+	Set $Namespace = "%SYS"
+	If $$$ISOK(##class(Config.Startup).Get(.tProperties)) {
+		Quit $Get(tProperties("SystemMode")) 
+	}
+	Quit ""
+}
+
+ClassMethod BuildIntroMessages() As %List
+{
+	Do ##class(%IPM.Main).GetVersion("zpm",.out)	
+	If $Get(out) = "" {
+		Set registryInfo = $$$FormattedLine($$$Cyan, "No registry configured")
+	} Else {
+		Set registryInfo = "Current registry "_$$$FormattedLine($$$Cyan, out)
+	}
+	Set systemMode = ..GetSystemMode()
+	If systemMode = "" {
+		Set systemMode = "System Mode: <unset>"
+	} Else {
+		Set systemMode = "System Mode: " _ systemMode
+	}
+	Set cls = "%ZHSLIB.Version", mthd = "GetFullBuildNumber"
+	If $System.CLS.IsMthd(cls, mthd) {
+		Set version = "HealthShare: " _ $ClassMethod(cls, mthd)
+	} Else {
+		Set version = $zv
+	}
+	Write !, $ZVersion // $ZV is too long for the box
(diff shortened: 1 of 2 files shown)


isc-shuliu:
@isc-eneil @isc-kiyer @isc-tleavitt 
Currently the welcome banner looks like this

image

Any idea how to do the remaining 2 items? 
1. whether or not this is the primary mirror member if part of a mirror or if on a backup
2. Some information about the `^%SYS("HealthShare","Instances")` (The HS instance on my device doesn't have `^%SYS("HealthShare","Instances")`, it only has `^%SYS("HealthShare")`, `^%SYS("HealthShare","ComponentDatabase", ...)`, and `^%SYS("HealthShare","SystemStartup","HS.HC.Util.SystemStartup")`)

isc-kiyer:
> @isc-eneil @isc-kiyer @isc-tleavitt Currently the welcome banner looks like this
> 
> image
> 
> Any idea how to do the remaining 2 items?
> 
> 1. whether or not this is the primary mirror member if part of a mirror or if on a backup
> 2. Some information about the `^%SYS("HealthShare","Instances")` (The HS instance on my device doesn't have `^%SYS("HealthShare","Instances")`, it only has `^%SYS("HealthShare")`, `^%SYS("HealthShare","ComponentDatabase", ...)`, and `^%SYS("HealthShare","SystemStartup","HS.HC.Util.SystemStartup")`)

@isc-shuliu if you use `##class(%ZHSLIB.HealthShareMgr).VersionInfo()`, it will give you the full HS version string which includes the product + what components are included (is the $zv equivalent for HS). Would recommend checking if the method exists before invoking it.

For mirroring, check if it is a mirror member and if so can display something like `Mirror Status: Primary/Backup`. Can use methods in `$System.Mirror` for that

src/cls/IPM/Main.cls
```diff
@@ -823,6 +823,47 @@ ClassMethod GetVersion(ModuleName, ByRef out, list)
...
+	} Else {
+		Set systemMode = "System Mode: " _ systemMode
+	}
+	Set cls = "%ZHSLIB.Version", mthd = "GetFullBuildNumber"
+	If $System.CLS.IsMthd(cls, mthd) {
+		Set version = "HealthShare: " _ $ClassMethod(cls, mthd)
+	} Else {
+		Set version = $zv
+	}
+	Write !, $ZVersion // $ZV is too long for the box
```
isc-kiyer, line 856: Why is $zv written both before and after? VersionInfo() for HS includes $zv as well so it can be there just once, either inside or outside the box. You could also manipulate the version string to fit in the box by adding new line characters as needed

isc-shuliu: Sorry that was some debugging code that I didn't remove

isc-shuliu:
Here's improved layout
image
image

isc-jili (approved):

isc-kiyer (approved):

Issue #722, opened 2025-01-31
Unify modifiers between `ModuleAction` and `RunOnePhase`


isc-shuliu:
Currently, lifecycle phases can be run through either `ModuleAction` or `RunOnePhase` (except for `CustomPhase`s). However, the modifiers supported between the 2 are different.

For example, when packaging a module,
```objectscript
zpm "<mod-name> package -exports-deps 1"  // this works
zpm "package <mod-name> -exports-deps 1"  // this doesn't work
```
Conversely,
```objectscript
zpm "<mod-name> package -path /my/path"  // this doesn't work
zpm "package <mod-name> -path /my/path"  // this works
```

To make IPM more user friendly, we need to support them both (or deprecate one in favor of the other).

PR #750, opened 2025-02-28
Unify modifiers between ModuleAction and directly calling lifecycle phase

isc-etamarch:
See #722 

This unifies modifiers between lifecycle actions and moduleaction to make the 2 interchangeable

### DIFF
--- a/src/cls/IPM/CLI.cls
+++ b/src/cls/IPM/CLI.cls
@@ -35,6 +35,10 @@ ClassMethod %Help(ByRef pCommandInfo) [ Final ]
 	If (tCommand '= "") && (tCommand '= +tCommand) && $Data(tCommandStruct(tCommand)) {
 		Kill tOneCommandStruct
 		Merge tOneCommandStruct = tCommandStruct(tCommand)
+		set phases = $select(tCommand = "module-action":##class(%IPM.Main).#STANDARDPHASES,1:$lb("module-action"))
+		if ((tCommand = "module-action") || ($listfind(##class(%IPM.Main).#STANDARDPHASES,tCommand))) {
+			do ..UnifyLifecycles(.tOneCommandStruct,.tCommandStruct,phases)
+		}
 		Do ..%HelpForCommand(tCommand,.tOneCommandStruct,1,tMarkdown)
 	} Else {
 		// For markdown mode only, merge aliases into main part of tCommandStruct array.
@@ -69,6 +73,10 @@ ClassMethod %Help(ByRef pCommandInfo) [ Final ]
 			
 			Kill tOneCommandStruct
 			Merge tOneCommandStruct = tCommandStruct(tCommand)
+			set phases = $select(tCommand = "module-action":##class(%IPM.Main).#STANDARDPHASES,1:$lb("module-action"))
+			if ((tCommand = "module-action") || ($listfind(##class(%IPM.Main).#STANDARDPHASES,tCommand))) {
+				do ..UnifyLifecycles(.tOneCommandStruct,.tCommandStruct,phases)
+			}
 			Do ..%HelpForCommand(tCommand,.tOneCommandStruct,tVerbose,tMarkdown,$Get(tCommandStruct(-1, "maxLength"),0))
 			
 			Write !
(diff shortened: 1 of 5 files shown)


src/cls/IPM/CLI.cls
```diff
@@ -438,22 +438,44 @@ ClassMethod %ParseCommandInput(pCommandString As %String, Output pCommandInfo) A
...
 						Set tModifier = tAliasFor
+					} ElseIf (($data(pCommandInfo("parameters","actions"),phases)) && (pCommandInfo = "module-action")) || ($listfind(##class(%IPM.Main).#STANDARDPHASES,pCommandInfo)) {
+						
+						// Lifecycle phases can be called directly or through module-action, we need to share modifiers
+						if (pCommandInfo = "module-action") {
+							set phases = $listfromString(phases)
+						} else {
+							set phases = $lb("module-action")
+						}
+						set llen = $LISTLENGTH(phases)
```
isc-tleavitt, line 454: Nitpick: always use [$listnext](https://docs.intersystems.com/iris20243/csp/docbook/Doc.View.cls?KEY=RCOS_flistnext#RCOS_flistnext_listnext_and_performance) to iterate over $listbuild lists.

isc-tleavitt (changes requested):
Love seeing the unit test on this!!

One minor nitpick.

isc-shuliu (approved):
@isc-etamarch 
Looks amazing! Does this mean that we can remove some of the repeated modifiers defined in `%IPM.Main:(XData)Commands`? For example, the `<modifier name="export-deps" .../>` is specified 3 times, in `moduel-action`, `package`, and `publish` respectively.

isc-kiyer (commented):
@isc-etamarch This looks great! Question: should the modifiers be removed for the individual phase commands (test, reload, compile etc.) that are duplicates in module-action? Also, if I were to look at the help output for one of the phase commands, would I see the options for module-action?

isc-etamarch:
> @isc-etamarch Looks amazing! Does this mean that we can remove some of the repeated modifiers defined in `%IPM.Main:(XData)Commands`? For example, the `<modifier name="export-deps" .../>` is specified 3 times, in `moduel-action`, `package`, and `publish` respectively.

> @isc-etamarch This looks great! Question: should the modifiers be removed for the individual phase commands (test, reload, compile etc.) that are duplicates in module-action? Also, if I were to look at the help output for one of the phase commands, would I see the options for module-action?

I modified the code so that help also includes the proper modifiers, and have removed the duplicates from the aliases

src/cls/IPM/CLI.cls
```diff
@@ -438,22 +481,43 @@ ClassMethod %ParseCommandInput(pCommandString As %String, Output pCommandInfo) A
...
+						
+						// Lifecycle phases can be called directly or through module-action, we need to share modifiers
+						if (pCommandInfo = "module-action") {
+							set phases = $listfromString(phases)
+						} else {
+							set phases = $lb("module-action")
+						}
+						set ptr = 0
+						while $listnext(phases,ptr,currPhase) {
+							if $data(tCommandStructure(currPhase,"modifiers",tAccum)) {
```
isc-kiyer, line 499: Am having a bit of an issue following the structure of the subscripted array here. Although slightly out of scope for this work, could you add documentation to this method to elaborate on the structure beyond the top level subscript?

isc-kiyer (commented):
@isc-etamarch minor comments. Looks good! Could you attach a screenshot as well of what the help command looks like now for say `help module-action` vs `help package`?

isc-etamarch:
> @isc-etamarch minor comments. Looks good! Could you attach a screenshot as well of what the help command looks like now for say `help module-action` vs `help package`?
@isc-kiyer I added a description of the array subscripts to the Commands.cls file, since that is where they're generated. Here are the help outputs: 
module-action:
image
image

package:
image

isc-kiyer (approved):
@isc-etamarch Approving since this looks good! General q for future: currently for each phase command, the summary is that it is an alias for `module-action`. However, would it be worthwhile to flip it and make each `phase` command what we expect someone to use and only use `module-action` for custom phases? Reasoning is specifically related to the help command. As a user, if I want to say publish a module, I want to see the help for the publish command and only the modifiers relevant to that. `module-action` currently lists modifiers for all commands (not just the ones that are common but also the ones specific to a phase. For example, `export-deps` is only relevant to package but is listed in `module-action`)
cc @isc-shuliu @isc-tleavitt

PR #748, opened 2025-02-28
Allow publishing of package with deployed code without being in dev mode

isc-etamarch:
See #745

This PR allows packages with deploy code to be published without having to be created in developer mode.

PR #753, opened 2025-03-05
Update v0.10.x with commits from main

isc-etamarch:
Merges  #748 and other into v0.10.x


module.xml
```diff
@@ -2,7 +2,7 @@
 <Export generator="Cache" version="25">
 <Document name="ZPM.ZPM"><Module>
   <Name>ZPM</Name>
-  <Version>0.9.2-SNAPSHOT</Version>
+  <Version>0.9.3-SNAPSHOT</Version>
```
isc-tleavitt, line 5: Should be 0.10.0-SNAPSHOT in v0.10.x?? Not sure why the diff is showing as this.

isc-tleavitt (changes requested):
One thing to worry about (module.xml version); otherwise good to go.

isc-shuliu (approved):

isc-tleavitt (approved):

PR #754, opened 2025-03-05
V0.10.x fix oras deployed code

isc-shuliu:
With this PR, deployed items are bundled into a %StudioProject for installing and uninstalling.

This PR also makes ORAS registries compatible with deployed items.


src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -258,23 +278,7 @@ Method OnExportDeployedItem(pFullExportPath As %String, pItemName As %String, By
     Set fileExt = $$$ucase($Piece(pItemName, ".", *))	
     Set tName = $Piece(pItemName, ".", 1, * - 1)	
     Quit:'($ListFind($ListBuild("CLS","MAC","INT"), fileExt))
-    Set tFullExportPath = pFullExportPath	
-    Set tFullExportPath = tFullExportPath _ ".deploy"	
-    Write:tVerbose !,"Exporting Deployed '",pItemName,"' to '",tFullExportPath,"'"	
-    Set list = ""	
-    If fileExt = "CLS" {	
-      Set list = ..ListGlobals4Class(tName)
```
isc-kiyer, line 266: Should these referenced methods of ListGlobals4Class/Routine be deleted?

src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -172,6 +172,26 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
                 Set tSC = $$$OK
                 Quit
               }
+              If ..ResourceReference.Deploy {
+                If tVerbose {
+                  Write !, "Skipping deployed resource: " _ ..ResourceReference.Name
+                }
+                Set ..CompileFromProject = 1
+                Set tDeployedDirectory = ##class(%File).NormalizeDirectory("studio-project", ..ResourceReference.Module.Root)
```
isc-kiyer, line 180: Could all "magic strings" be macros or helper methods? So that includes "studio-project", "__deployed" and "Deployed.xml". Perhaps helper methods to construct the file name for a given module's deployed studio project file + a helper method for the studio project name for a module's deployed code?

src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -172,6 +172,26 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
...
+                $$$ThrowOnError($System.OBJ.Load(tDeployedDirectory_"Deployed.xml", "<none>", , .tLoadedList))
```
isc-kiyer, line 181: Should Load() be used to import studio projects? Usually we use %Studio.Project:ImportFromFile(). This will also ensure any ^INSTALL routine packaged with the studio project is run. Not sure how we want to handle the ^INSTALL routine in this case though (or if we even want it to run) since this isn't really the same concept as packaging the entire module as a studio project so it can be imported in a context without IPM. Here, we would import the studio project in the reload phase and then continue with regular activate/configure to run invokes and installer class code so don't need the ^INSTALL routine. @isc-tleavitt thoughts?

src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -172,6 +172,26 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
                 Set tSC = $$$OK
                 Quit
               }
+              If ..ResourceReference.Deploy {
+                If tVerbose {
+                  Write !, "Skipping deployed resource: " _ ..ResourceReference.Name
```
isc-kiyer, line 177: I'm a bit confused by this. Is the resource getting skipped or is it just getting loaded from a studio project so regular handling is skipped?

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -883,14 +887,26 @@ Method %Compile(ByRef pParams) As %Status
...
+				If tResource.Processor.%IsA("%IPM.ResourceProcessor.AbstractCompilable") && tResource.Processor.CompileFromProject && ('tDeployedItemsLoaded) {
+					Set tDeployedItemsLoaded = 1
+					Merge deployedList = tResource.Processor.CompileFromProjectList
+					Set deployedItem = ""
+					For {
+						Set deployedItem = $Order(deployedList(deployedItem))
+						If deployedItem = "" {
+							Quit
+						}
+						Set tCompileFlagMap("ck", deployedItem, deployedItem) = ""
```
isc-kiyer, line 907: For a studio project, do we need to recompile if we just call InstallFromFile()? Seems like we wouldn't need to so don't need to do this additional accumulation.
Also, if a module has any deployed resources, do we put all compilable resources into the studio project? In that case, could just skip regular compile logic

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -114,22 +118,27 @@ Method ListModules(pSearchCriteria As %IPM.Repo.SearchCriteria) As %ListOfObject
...
 
 				Set tModRef = ##class(%IPM.Storage.ModuleInfo).%New()
 				Set tModRef.Name = artifactMetadata.ImageTitle
 				Set tModRef.Repository = artifactMetadata.ImageSource
 				Set tModRef.VersionString = artifactMetadata.ImageVersion
 				Set tModRef.Description = artifactMetadata.ImageDescription
 				Set tModRef.Deployed = artifactMetadata.IPMDeployed
+				Set tModRef.PlatformVersion = artifactMetadata.IPMPlatformVersion
+				If artifactMetadata.IPMPlatformVersion '= "" {
+					$$$ThrowOnError(tModRef.PlatformVersions.Insert(artifactMetadata.IPMPlatformVersion))
```
isc-kiyer, line 140: Maybe a stuipd question but what is the purpose of PlatformVersion vs PlatformVersions and why do we need 2 properties to encapsulate this? PlatformVersions in %IPM.Storage.ModuleInfo doesn't have a description on it describing its purpose

isc-kiyer (changes requested):
@isc-shuliu Looks good but a few comments and action items. Also, general q: for listing modules in a repo + being able to install them, what is the defined behavior? I can think of the following cases:
- Module exists in non-deployed mode: in this case, no special handling since it can be installed on any system
- Module exists in deployed mode for version v1 and v2 and user is on v1 or v2: I think list-modules should display the versions for each platform available in list-modules but when the user tries to install, it picks the version they are on and installs the module. Perhaps could grey out or have an additional flag to list modules that are available in the repo but not installable on the current platform? By default I agree a user shouldn't be able to see module versions built for other platforms.
- Module exists in deployed mode for version v1 and v2 and user is on v3: based on the above, a user would fail to install the module but would be worthwhile them getting a better error message indicating the module exists in the repo but is incompatible with their platform version and offering them a command to view all platform versions available a module (some variation of list-modules)

src/cls/IPM/ResourceProcessor/Default/Document.cls
isc-tleavitt: I don't think we should support use of ^INSTALL in an IPM context. Better to use invokes to be more explicit/transparent about what's happening.

src/cls/IPM/Lifecycle/Base.cls
isc-tleavitt: Agreed, we don't need to recompile.

We only put deployed resources into the studio project.

src/cls/IPM/Repo/Oras/PackageService.cls
isc-tleavitt: The inheritance hierarchy is a little weird here. In some cases we're interested in the PlatformVersion of a specific ModuleInfo; when we're getting back results from a repo we might want a list of all the available PlatformVersions.

Some of this might be still be leftover from the evolution of old deployed code support, where (perhaps) initially there could only be one PlatformVersion.

I wonder if the right thing here is to accumulate PlatformVersions by Name and not use the PlatformVersion property? As it stands this would only have one PlatformVersion at most. We need to test with multiple PlatformVersions on the same repo; any issues along these lines should pop out quickly.

src/cls/IPM/Lifecycle/Base.cls
isc-shuliu: My concern was that if a user runs `<package> reload -only` or `<package> compile -only`, they would expect only to load the items or compile items that have already been loaded. Do we want to combine the two for %Studio.Project of deployed items? If so, do we want to place the `InstallFromFile` step in which phase?

src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -258,23 +278,7 @@ Method OnExportDeployedItem(pFullExportPath As %String, pItemName As %String, By
...
-    }	
-    Quit:list=""	
-    Set ptr = 0	
-    While $ListNext(list, ptr, item) {
-      Set list(item _ ".gbl") = ""	
-    }	
-    $$$ThrowOnError(##class(%File).CreateDirectoryChain(##class(%File).GetDirectory(tFullExportPath)))	
-    Set flags = $Select(tVerbose: "d", 1: "-d")	
-    $$$ThrowOnError($System.OBJ.Export(.list, tFullExportPath, flags))	
+    // TODO verify that the deployed item is in the %Studio.Project
```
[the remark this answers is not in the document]
isc-shuliu: So this goes back to the question of how do we map the resource name, such as `Foo.Bar.CLS` to deployed item names, such as `Foo.Bar.0.OBJ` and `Foo.Bar.1.OBJ`. You mentioned today that all .CLS will generate a .INT and a .0.OBJ and .1.OBJ. Can you point me to the direction of what gets generated from other resources (routines, inc, etc) ?

src/cls/IPM/Lifecycle/Base.cls
isc-kiyer: Right that's a good question. I think with studio project, it is fundamentally compiled code so loading is effectively compiling. I think it makes sense to put it into reload since compile is effectively skipped in this case. So place InstallFromFile in reload IMO.

src/cls/IPM/ResourceProcessor/Default/Document.cls
isc-kiyer: In that case, should ensure ^INSTALL isn't packaged in the studio project in this case. I know there is some logic in the studio project lifecycle class for it which is why I asked.

src/cls/IPM/ResourceProcessor/Default/Document.cls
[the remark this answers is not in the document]
isc-kiyer: @isc-shuliu  Taking a step back, what is the purpose of this method? We package any deployed items into a studio project in the lifecycle class + the GetStudioProject() method on the module so is this needed? We should either stick with that or adapt this to behave like general lifecle + resource processor handling. 

My suggestion: do not use GetStudioProject() in the module class since it deviates from our general lifecycle + processor class handling. Instead, in %Export() instantiate a new studio project and pass it down to the resource processors if their corresponding resource is marked deployed (perhaps via pParams or a singleton of some sort) and resource processors can accordingly add to the studio project in their OnPhase() methods. Then once resource processor delegation is done, in the main lifecycle class can export the studio project to the file system. How does that sound?

cc @isc-tleavitt

src/cls/IPM/Repo/Oras/PackageService.cls
isc-shuliu: `PlatformVersions` was introduced in [this commit](https://github.com/intersystems/ipm/commit/188b63227c2eda5d1d9df12b1957a965edba424d#diff-717bb46727b5a8ff132aa006a2c5434ddcb15d3fb63c3c4de02308ad467c58abR5), as part of the effort to support the corresponding support for deployed code on zpm-registry. 

From [this implementation](https://github.com/intersystems-community/zpm-registry/commit/6c53e0b52da376a5c0f5172928720054d8e98181#diff-33884809558a308b624a52b752be0d243d34b058d4d31dcede97bcf8b15c2567R206) of ZPM registry, it appears `PlatformVersions` is meant to be an output-only JSON field that aggregates all `PlatformVersion` from input. But for ORAS, there isn't a server side API that aggregates tags by their prefix, so I'll pull a list of tags and aggregate the `PlatformVersion` part from all tags into a `PlatformVersions` on the IPM side.

isc-shuliu:
> @isc-shuliu Looks good but a few comments and action items. Also, general q: for listing modules in a repo + being able to install them, what is the defined behavior? I can think of the following cases:
> 
> * Module exists in non-deployed mode: in this case, no special handling since it can be installed on any system
> * Module exists in deployed mode for version v1 and v2 and user is on v1 or v2: I think list-modules should display the versions for each platform available in list-modules but when the user tries to install, it picks the version they are on and installs the module. Perhaps could grey out or have an additional flag to list modules that are available in the repo but not installable on the current platform? By default I agree a user shouldn't be able to see module versions built for other platforms.
> * Module exists in deployed mode for version v1 and v2 and user is on v3: based on the above, a user would fail to install the module but would be worthwhile them getting a better error message indicating the module exists in the repo but is incompatible with their platform version and offering them a command to view all platform versions available a module (some variation of list-modules)

Opened a separate issue for this. For now, `list-modules` will list the latest package version regardless of platform version compatibility. When installing, we default to the latest package version and verify platform compatibility. Users can optionally install another version.

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -56,6 +60,115 @@ Method HasModule(pModuleReference As %IPM.Storage.ModuleInfo) As %Boolean
...
+		Set $ListBuild(version, platformVersion) = $ListFromString(tag, $$$OrasTagPlatformSeparator)
+		Set value = $Get(aggregatedPlatformVersion(version))
+		Set aggregatedPlatformVersion(version) = value _ $ListBuild(platformVersion) 
+
+		// prepare for sorting
+		Set semver = ##class(%IPM.General.SemanticVersion).FromString($$$Tag2Semver(version))
+		Set sorter(+semver.Major, +semver.Minor, +semver.Patch, " "_semver.Prerelease, " "_semver.Build) = version
+	}
+
+	Set major = ""
```
isc-tleavitt, line 86: This is a great use case for [$query](https://docs.intersystems.com/iris20243/csp/docbook/Doc.View.cls?KEY=RCOS_fquery) but don't need to do that right now.

isc-tleavitt (approved):

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -890,7 +894,11 @@ Method %Compile(ByRef pParams) As %Status
 			Quit:tKey=""
 			
 			If '$Data(tHandledMap(tResource.Name)) && $IsObject(tResource.Processor) {
-				If tResource.Processor.%IsA("%IPM.ResourceProcessor.AbstractCompilable") {
+				If tResource.Processor.%IsA("%IPM.ResourceProcessor.AbstractCompilable") && tResource.Processor.CompileFromProject {
+					If tVerbose {
+						Write !, tResource.Name_" was loaded from studio project, skipping compilation."
```
isc-kiyer, line 899: Nit: rather than saying studio-project here, perhaps says "was loaded from deployed code file" since the format of deployed code is irrelevant. Also, instead of "skipping compilation", maybe "so compilation isn't needed"? cc @isc-tleavitt thoughts on wording?

isc-kiyer (changes requested):
@isc-shuliu looks good! Some minor comments and questions

src/cls/IPM/Lifecycle/Base.cls
isc-tleavitt: Let's use wording:
tResource.Name_" was loaded as deployed code, so compilation isn't needed."

src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -172,6 +174,21 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
                 Set tSC = $$$OK
                 Quit
               }
+              If ..ResourceReference.Deploy {
```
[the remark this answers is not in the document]
isc-tleavitt: In that case we won't get here because the resource will exist. (Might be worth adding a comment to that effect.)

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -351,6 +351,10 @@ Method %Clean(ByRef pParams) As %Status
 							If $$$ISERR(tDelSC) {
 								Write !,"WARNING: ",$System.Status.GetErrorText(tDelSC)
 							}
+						} ElseIf $System.CLS.IsMthd(tClassKey, "%Extends") {
```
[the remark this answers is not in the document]
isc-tleavitt: Yes - apparently they don't have either, this was news to me. (I thought they still had some stub of one or the other.)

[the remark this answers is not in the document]
isc-kiyer: Hmm I just checked on IRIS 24.1 and see contents in both oddDEF and oddCOM for some deployed classes in HS kits in HSLIB. Do we have an example of a class that has no contents when deployed?

src/cls/IPM/ResourceProcessor/Default/Document.cls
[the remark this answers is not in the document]
isc-kiyer: Ah oops missed that part

isc-tleavitt (approved):

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1656,24 +1678,22 @@ Method %MakeDeployed(ByRef pParams) As %Status
...
-			If (tVerbose) {
-				Set tClass = ""
-				Write !,"The following classes are flagged for deployment:"
-				#; Write !,"The following classes "_$Select(tDev:"would be flagged",1:"are flagged")_" for deployment:"
-				For {
-					Set tClass = $Order(tDeployClassArray(tClass))
-					Quit:tClass=""
-					Write !,$Char(9),tClass
+			If $Data(tDeployClassArray) {
+				Set ..HasDeployedResources = 1
```
isc-kiyer, line 1684: @isc-shuliu  small thing but you set this property twice and also the $Data() check for the routine array went away for some reason. You can move the setting of .. properties + save to be the first thing in the outer if block and then do the separate if blocks for classes vs routines.

isc-kiyer (approved):

Issue #689, opened 2025-01-07
IPM shell banner rework and "info" command


isc-tleavitt:
Add a new command, zpm "info"

zpm "info" and zpm "info -verbose" are "get my bearings" and "answer all of Support's preliminary questions" use cases, respectively.

Display names of top-level packages (not dependencies) in the namespace are part of the base command, verbose variant shows External Name (internal-name) as relevant.
 
And the IPM shell welcome text will call "info"
 
In "info -verbose" and "list-installed" put internal name in parentheses after external name

Related to #613 - everything covered there falls into either "get my bearings" or "answer all of Support's preliminary questions"

isc-eneil:
One more requirement emerged based on a discussion between HSDevOps and AppServices today, 2/18/2025: **the build info part of a module's version should not be displayed in the `zpm "info"` command.**

Background: 
IPM supports publishing modules with build info (in semantic versioning, this is everything after the +). E.g. For a module of version 1.0.0-beta+234523, the build info is "234523". The module's full version should be displayed as the module's version on `zpm "list-installed"`. This aligns with other ISC utilities show system info, e.g. $ZV which prints the full build version string of an instance. However, for the `zpm "info"` command, the build info part of the module's version should be ignored so that it is easier for a customer to quickly get their bearings on the modules present in their namespace.

cc @isc-tleavitt @isc-shuliu @isc-kiyer @isc-jili @isc-jlechtne

PR #755, opened 2025-03-10
V0.10.x info command

isc-shuliu:
This PR adds an `info` command per #689 . It lists top level packages and their external name (if existent) and the semver without the `build` part. If a `-verbose/-v` modifier is passed, it will include default name in parentheses. 

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #720 Support offline installation of oras using fixed version of pure python wheels and an adaptor for rpds.
 - #746: Added support for loading modules synchronously without multiprocessing
 - #749: Added more debugging information in the welcome banner
+- #755: Added an `info` command which prints external name (optionally including the real name) of top-level packages without the `build` part of semver. 
 
 ### Changed
 - #702 Preload now happens as part of the new `Initialize` lifecycle phase. `zpm "<module> reload -only"` will no longer auto compile resources in `/preload` directory.
--- a/src/cls/IPM/General/SemanticVersion.cls
+++ b/src/cls/IPM/General/SemanticVersion.cls
@@ -32,6 +32,14 @@ _ $Case(..Prerelease, "": "", : "-" _ ..Prerelease)
 _ $Case(..Build, "": "", : "+" _ ..Build)
 }
 
+Method ToStringWithoutBuild() As %String [ CodeMode = expression ]
+{
+..Major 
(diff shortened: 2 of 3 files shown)


isc-kiyer (approved):

isc-tleavitt (changes requested):
Going back to the spec in https://github.com/intersystems/ipm/issues/689 there are a few more pieces than have been addressed here:

zpm "info" and zpm "info -verbose" are "get my bearings" and "answer all of Support's preliminary questions" use cases, respectively.

- [x] Display names of top-level packages (not dependencies) in the namespace are part of the base command, verbose variant shows External Name (internal-name) as relevant.
- [x] The IPM shell welcome text will call "info"
- [x] In "info -verbose" and put internal name in parentheses after external name
- [ ] In "list-installed" put internal name in parentheses after external name

For "answer all of Supports' preliminary questions I think we will want the build in the -verbose mode.

Please confirm: @isc-kiyer @isc-eneil @isc-jili

isc-shuliu:
My concern is the `info` command can be very long when users install many packages, since each package takes a row. It may be longer than the height of the terminal screen and can clutter the terminal when user runs `zpm`. Are we sure this won't impact UX?

> Going back to the spec in #689 there are a few more pieces than have been addressed here:
> 
> zpm "info" and zpm "info -verbose" are "get my bearings" and "answer all of Support's preliminary questions" use cases, respectively.
> 
> * [x]  Display names of top-level packages (not dependencies) in the namespace are part of the base command, verbose variant shows External Name (internal-name) as relevant.
> * [ ]  The IPM shell welcome text will call "info"
> * [x]  In "info -verbose" and put internal name in parentheses after external name
> * [ ]  In "list-installed" put internal name in parentheses after external name
> 
> For "answer all of Supports' preliminary questions I think we will want the build in the -verbose mode.
> 
> Please confirm: @isc-kiyer @isc-eneil @isc-jili

isc-tleavitt:
> @isc-tleavitt My concern is the `info` command can be very long when users install many packages, since each package takes a row. It may be longer than the height of the terminal screen and can clutter the terminal when user runs `zpm`. Are we sure this won't impact UX?
> 
> > * [ ]   The IPM shell welcome text will call "info"

Good point - let's flip this around: zpm "info" should also show the information in the shell welcome text (version, etc.) as a one-stop shop for everything Support would need.

isc-shuliu:
@isc-tleavitt 
This is the updated behavior.
Google Chrome 2025-03-13 12 15 08

isc-tleavitt (approved):

PR #756, opened 2025-03-10
V0.10.x feat run commands with external name

isc-shuliu:
* Support running commands using external names of packages.
* External name of packages are now unqiue and can no longer conflict with the real name of another packages.

### DIFF
--- a/src/cls/IPM/Storage/Module.cls
+++ b/src/cls/IPM/Storage/Module.cls
@@ -213,6 +213,7 @@ ClassMethod ExecutePhases(pModuleName As %String, pPhases As %List, pIsComplete 
 		Set tModule = ..NameOpen(pModuleName,,.tSC)
 		If '$Isobject(tModule) {
 			Set tModule = ..ExternalNameOpen(pModuleName,,.tSC)
+			Set pModuleName = tModule.Name
 		}
 		If '$Isobject(tModule) {	
 			$$$ThrowStatus($$$ERROR($$$GeneralError, $$$FormatText("Module '%1' not found.", pModuleName)))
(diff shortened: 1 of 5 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -209,7 +211,11 @@ ClassMethod ExecutePhases(pModuleName As %String, pPhases As %List, pIsComplete
 	Try {
 		Set tVerbose = $Get(pParams("Verbose"),0)
 		Set tModule = ..NameOpen(pModuleName,,.tSC)
-		If '$isobject(tModule) {
+		If '$Isobject(tModule) {
+			Set tModule = ..ExternalNameOpen(pModuleName,,.tSC)
+			Set pModuleName = tModule.Name
```
isc-kiyer, line 216: This would throw an undefined error when accessing Name property if the module doesn't exist. So need to move the $IsObject check to be inside the if block.

src/cls/IPM/Storage/Module.cls
```diff
@@ -1216,6 +1222,11 @@ Method %OnValidateObject() As %Status [ Private, ServerOnly = 1 ]
 				Quit
 			}
 		}
+		Set query = "SELECT %DLIST(Name) AS NameList FROM %IPM_Storage.ModuleItem"
+		Set rs = ##class(%SQL.Statement).%ExecDirect(,query)
+		If rs.%Next() && $ListFind(rs.%Get("NameList"), ..ExternalName) {
```
isc-kiyer, line 1227: Nit: Rather than doing this, why not query with a where clause on ExternalName? Would be more efficient since the table could use the index on ExternalName.

isc-kiyer: Also, should exclude this module from the query in case this module has the same name and external name (weird edge case but its possible and would be a confusing error)

isc-kiyer (changes requested):
@isc-shuliu few minor comments

src/cls/IPM/Storage/Module.cls
isc-shuliu: Completely agree. I wasn't thinking straight on this.

isc-kiyer (approved):

isc-tleavitt (approved):

PR #656, opened 2024-12-11
Fix #652: Don't create mappings unless they're actually needed

isc-tleavitt:
@isc-kiyer @isc-eneil and others - any concerns on this from a HS perspective?

PR #757, opened 2025-03-11
Fix IPM bug where mappings are not getting created when they should

isc-jili:
Bug where mappings are not getting created when they should was introduced in https://github.com/intersystems/ipm/pull/656/
This is because the the original change contains inverted logic where mappings are created if the routine is already mapped to the current default routine DB. We want it to be the other way around: to create mappings only if the routine is NOT already mapped to the current default routine DB.
This merge will fix this bug.


isc-kiyer (changes requested):
@isc-jili Changes look good! Worth adding some unit tests though. Test the following cases:
- behavior of package, class, routine, loc file when there are no existing package mappings present (no mappings to be created)
- behavior of package, class routine, loc file when a package mapping exists for a parent package (mappings created) e.g. module containts package Foo.Bar.PKG, Foo.Bar.Errors.LOC etc. and a package mapping already exists for Foo from another database. In that case, a new mapping should be created for Foo.Bar. Can check if mappings were correctly created both by using %SYS.Namespace:Get*Dist() methods and Exists() method for Config.Map* classes

isc-shuliu (approved):
Can we add a unit/integration test to verify it's working as intended?

isc-eneil:
@isc-jili No further comments from me-- will take a look back once unit tests have been added.

isc-jili:
Added unit tests in Mappings.cls @isc-kiyer @isc-eneil @isc-jlechtne @isc-shuliu @isc-tleavitt

tests/unit_tests/Test/PM/Unit/Mappings.cls
```diff
@@ -0,0 +1,463 @@
...
+    Set ..Default.Class = "%IPM.ResourceProcessor.Default.Routine"
+
+    // Add the default setting to the module
+    Do ..ResourceReferenceClass.Module.Defaults.Insert(..Default)
+
+    Set testObjMAC = ##class(%IPM.ResourceProcessor.Default.Routine).%New(..ResourceReferenceClass)
+    Set ns = "USER"
+    Set parentRoutine = "Foo"
+    Set childRoutine = "Foo.Bar"
+    Set db = "HSLIB"
```
isc-kiyer, line 46: @isc-jili Looks good but HSLIB doesn't exist on vanilla IRIS so would throw an error. You can create your own database at the beginning of the test and delete it by the end (OnBefore/AfterAllTests) or could map from say IRISSYS/another db that always exists?

isc-tleavitt (approved):

PR #759, opened 2025-03-11
Replaced blue in terminal with default

### LINKED ISSUE #751: Colours in the list option show blue on black for the version
It's very difficult to read the output of the list option because the version is shown as blue on black.

Image

isc-etamarch:
Fixes #751 (Blue output in terminal is now white)

image


isc-tleavitt:
Oh, that was easy. Great!

isc-kiyer (approved):

PR #761, opened 2025-03-17
Adding module name as first key in resource array

isc-jlechtne:
To fix FileCopy issue as well as others that we haven't run into yet but would otherwise run into the same bug

### DIFF
--- a/src/cls/IPM/Lifecycle/Base.cls
+++ b/src/cls/IPM/Lifecycle/Base.cls
@@ -1088,9 +1088,13 @@ Method %Export(ByRef pParams, ByRef pTargetDirectory As %String, Output pDepende
 		
 		// First pass: if we're exporting dependencies, build an array of modules from which we are deriving things.
 		If tExportDependencies {
+			Set tModuleName = ""
+			For {
+				Set tModuleName = $Order(tResourceArray(tModuleName))
+				Quit:(tModuleName="")
 			Set tFullResourceName = ""
 			For {
-				Set tFullResourceName = $Order(tResourceArray(tFullResourceName))
+					Set tFullResourceName = $Order(tResourceArray(tModuleName, tFullResourceName))
 				Quit:(tFullResourceName="")
 				
 				Set tExt = $Piece(tFullResourceName,".",*)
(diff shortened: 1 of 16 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -1006,7 +1006,7 @@ Method GetResolvedReferences(Output pReferenceArray, pLockedDependencies As %Boo
 			}
 		}
 		
-		Set pReferenceArray(..Name_".ZPM") = ""
+		Set pReferenceArray(..Name, ..Name_".ZPM") = ""
```
isc-kiyer, line 1009: Nit: This is not needed since for the base module, don't want module scoping and for dependencies, this is handled by using tempArray below and copying back into the subscripted array with first subscript as module name

src/cls/IPM/Storage/Module.cls
```diff
@@ -1038,7 +1040,7 @@ Method GetResolvedReferences(Output pReferenceArray, pLockedDependencies As %Boo
 				}
 				
 				// Don't pass along the phase
-				Set pReferenceArray(tModName_".ZPM") = ..Name
+				Set pReferenceArray(tModName, tModName_".ZPM") = ..Name
```
isc-kiyer, line 1043: For same reason as above, can remove this.

isc-kiyer (changes requested):
@isc-jlechtne few small comments. Please add unit tests. For at least the following cases:
- Package of FileCopy of a module with no dependencies
- Package of FileCopy of a module with dependencies where dependency FileCopy resources have different names
- Package of FileCopy of a module with dependencies where dependency FileCopy resources have the same name
- Confirm install behavior of FileCopy has no issues.
- MakeDeployed for just a module as well as recursively for its dependencies as well

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1732,6 +1733,11 @@ Method ExportSingleModule(
 		
 		Set tFullPath = ##class(%File).NormalizeFilename(pTargetDirectory_tRelativePath)
 		Set tFullPath = ##class(%File).NormalizeFilenameWithSpaces(tFullPath)
+
+		// Handle FileCopy different --> set target directory from dependency graph
+		If ($FIND($GET(pResourceArray(tFullResourceName, "Processor")), "%IPM.ResourceProcessor.FileCopy")) {
```
isc-shuliu, line 1738: I'd recommend using `processor.%IsA("%IPM.ResourceProcessor.FileCopy")` instead of casting processor obj to a string.

src/cls/IPM/Storage/Module.cls
isc-jlechtne: While GetResolvedReferences is called recursively, the info added to the pReferenceArray is the <module>.ZPM entry created here and the resources themselves, taken care of in ResolveChildren(). Since <module>.ZPM isn't a part of the tempArray, we do have to add the first subscript in here.

src/cls/IPM/Storage/Module.cls
isc-kiyer: So the above one is needed but this one isn't since it calls GetResolvedReferences() and not ResolveChildren()

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1807,7 +1721,129 @@ Method GetResourceRelativePath(pResource As %String, pExtension As %String) As %
...
+			Set tRelativePath = pDependencyGraph(tName,"Location")_tRelativePath
+		} ElseIf (tSrcModule '= "") && (tSrcModule '= ..Module.Name) {
+			Set tRelativePath = pDependencyGraph(tSrcModule,"Location")_tRelativePath
+		}
+		
+		Set tFullPath = ##class(%File).NormalizeFilename(pTargetDirectory_tRelativePath)
+		Set tFullPath = ##class(%File).NormalizeFilenameWithSpaces(tFullPath)
+
+		// Handle FileCopy different --> set target directory from dependency graph
+		If ($Get(pResourceArray(tFullResourceName, "Processor")) '= "") && $Get(pResourceArray(tFullResourceName, "Processor")).%IsA("%IPM.ResourceProcessor.FileCopy") {
```
isc-kiyer, line 1758: Would such explicit hard coding for FileCopy affect other resource processors? For example, in HS, we have AngularArtifact and JavaLibrary. Ideally we should be consistent with what is passed to OnExportItem for the base module and dependencies so it can behave the same way but its inputs will guarantee that the files go to their correct locations (whether its for the base module or under .modules/ for dependencies)

isc-tleavitt: Agreed, we should just do this across the board.

isc-jlechtne:
> > @isc-jlechtne few small comments. Please add unit tests. For at least the following cases:
> > 
> > * Package of FileCopy of a module with no dependencies
> > * Package of FileCopy of a module with dependencies where dependency FileCopy resources have different names
> > * Package of FileCopy of a module with dependencies where dependency FileCopy resources have the same name
> > * Confirm install behavior of FileCopy has no issues.
> > * MakeDeployed for just a module as well as recursively for its dependencies as well
> 
> I'll add a couple more cases that seem relevant from my initial testing:
> 
> * Install and Package of a module with a FileCopy resource with source and dest _directories_ where:
>   -- Overlay = 1 and Overlay - 0
>   -- Target directory exists beforehand and Target directory does not exist
>   -- Target directory is a system variable that references a directory relative to IRIS instance root directory

@isc-eneil I'm not so sure those cases are within the scope of this issue. Those tests would be related to just general FileCopy install behavior (and flags) whereas this issue is focused on packaging and making sure dependency locations are created correctly.

isc-kiyer (approved):

isc-jili (approved):

isc-shuliu (approved):

PR #774, opened 2025-04-01
Only add PrepareDeploy phase if module is in dev mode

isc-jlechtne:
Relating to HSIEO-12012 regarding publishing modules with deployed code. Larger issue with temporary namespace created when executing the phases with deployed code. Talked with @isc-kiyer and a solution to the problem that faces us at the moment is to only add the PrepareDeploy phase if code is in Dev mode.

(Created a different jira for the fix with the larger NS issue)


isc-kiyer:
Note: We only need PrepareDeploy in dev mode since that is when we want to retain the current state of files having installed from source code. If in non-dev mode, then have installed only for publishing so its fine to deploy the code in the current namespace.

isc-jili (approved):

PR #775, opened 2025-04-03
fix: fix a bug when discovering packages from oras repo

isc-shuliu:
The version info in metadata can be different from the version encoded in tag name


isc-tleavitt (approved):
Needs CHANGELOG, otherwise looks good.

isc-kiyer (approved):

Issue #781, opened 2025-04-08
Workaround for file paths exceeding 256 characters


isc-jlechtne:
When testing install of DTR and PAS modules built and deployed with 0.10.0-beta.30 IPM, runs into the following error:
```
zpm:USER>install test/intersystems_payer_services_hp_dtr 1.1.0-snapshot+CMS-0057-WIP-jlou-11706-7731740
ERROR! ObjectScript error: <NAMEADD>MoveToEnd+10^%Stream.FileBinary.1
zpm:USER>install hshp/intersystems_payer_services_hp_pas 1.1.0-snapshot+CMS-0057-WIP-jlou-11714+7744270
ERROR! ObjectScript error: <NAMEADD>MoveToEnd+10^%Stream.FileBinary.1
```
This is due to file paths being over 256 characters. Example file path: /usr/irissys/mgr/.modules/USER/intersystems_payer_services_hp_dtr/1.1.0-snapshot_CMS-0057-WIP-11750/.modules/healthconnect.fhirmeta/dev/fhir/fhir-metadata/packages/hl7.fhir.r4.core/SearchParameter-diagnosticreport-genetic-DiagnosticReport-assessed-condition.json.

Temporary workaround was discussed to use a system-generated temporary directory instead of something under /mgr.

isc-tleavitt:
Fixed in #783

PR #783, opened 2025-04-09
Use system temporary directory instead of long path when installing modules

isc-jlechtne:
To address https://github.com/intersystems/ipm/issues/781

### DIFF
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -389,7 +389,7 @@ ClassMethod LoadModuleFromArchive(pModuleName As %String, pModuleVersion As %Str
 	Try {
 		Set tVerbose = $Get(pParams("Verbose"))
 		
-		Set tTargetDirectory = ##class(%SYS.System).TempDirectory()_"/"_##class(%PopulateUtils).StringMin(5,5)
+		Set tTargetDirectory = ##class(%SYS.System).TempDirectory()_"/"_pModuleName
 		If ##class(%File).DirectoryExists(tTargetDirectory) {
 			// Delete it.
 			Set tSC = ##class(%IPM.Utils.File).RemoveDirectoryTree(tTargetDirectory)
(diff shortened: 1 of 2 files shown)


src/cls/IPM/Utils/Module.cls
```diff
@@ -389,8 +389,7 @@ ClassMethod LoadModuleFromArchive(pModuleName As %String, pModuleVersion As %Str
 	Try {
 		Set tVerbose = $Get(pParams("Verbose"))
 		
-		Set tTargetDirectory = ##class(%File).ManagerDirectory()_".modules\"_$Namespace_"\"_pModuleName_"\"_pModuleVersion_"\"
-		Set tTargetDirectory = ##class(%File).NormalizeDirectory(tTargetDirectory)
+		Set tTargetDirectory = ##class(%SYS.System).TempDirectory()
```
isc-tleavitt, line 392: Really want a subdirectory of this.

isc-tleavitt (changes requested):
Needs a random *subdirectory* of ##class(%SYS.System).TempDirectory() and CHANGELOG.md

isc-jili (changes requested):
Perhaps add update to CHANGELOG?

src/cls/IPM/Utils/Module.cls
isc-kiyer: Subdirectory can be the module name and should check if the directory exists first and delete it if so. Also, probably good to delete the directory at the end of the run

isc-kiyer (approved):

src/cls/IPM/Utils/Module.cls
```diff
@@ -419,6 +418,13 @@ ClassMethod LoadModuleFromArchive(pModuleName As %String, pModuleVersion As %Str
 		If $$$ISERR(tSC) {
 			Quit
 		}
+		// Delete directory at the end of the run
```
isc-tleavitt, line 421: If something goes horribly wrong, we won't delete the directory.

I think this is a feature.

isc-jili (approved):

PR #792, opened 2025-04-11
Fix 786: Resources with scope should not be exported

isc-jili:
Fix for issue 786:

Resources with phase and scope are being exported even though they should not.

The following changes should be addressed:

Adding throw on error to %IPM.Lifecycle.Base ExportSingleModule() to avoid silently failing
In %IPM.Storage.Module GetResolvedReferences() adding check to tResource.Attributes.GetAt(“Phase”) to not include resource for export if there is a phase (in addition to scope checking)
In %IPM.ResourceProcessor.Test change Phase property to be a comma separated string instead of a List because it is not necessary to be a list

### DIFF
--- a/src/cls/IPM/ResourceProcessor/Test.cls
+++ b/src/cls/IPM/ResourceProcessor/Test.cls
@@ -8,7 +8,7 @@ Parameter DESCRIPTION As STRING = "Loads unit tests (extending %UnitTest.TestCas
 Parameter ATTRIBUTES As STRING = "Phase,Package,Class,ManagerClass,Format,ExportFlags";
 
 /// Phase in which the unit tests run - "test" (dev namespace), "verify" (separate, clean namespace), or "test,verify" (to run in both cases).
-Property Phase As %String [ InitialExpression = "test", Required ];
+Property Phase As %IPM.DataType.ListOfOptions(VALUELIST = ",test,verify") [ InitialExpression = {$ListBuild("test")}, Required ];
 
 // NOTE: Either Package or Class MUST be defined.
 
@@ -83,7 +83,7 @@ Method TestsShouldRun(pPhase As %String, ByRef pParams) As %Boolean
 			// always equals 1
 		}
 	}
-	Quit tPathMatch && ($Find(..Phase,$ZConvert(pPhase,"L")) > 0)
+	Quit tPathMatch && ($ListFind(..Phase,$ZConvert(pPhase,"L")) > 0)
 }
 
 Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boolean = 0) As %Status
@@ -207,7 +207,7 @@ Method OnResolveChildren(ByRef pResourceArray) As %Status
 		Set tResourceInfo = tModuleName
 		Set tResourceInfo("Generated") = 0
 		Set tResourceInfo("Preload") = 0
-		Set tResourceInfo("Scope") = ..Phase
+		Set tResourceInfo("Scope") = $ListToString(..Phase, ",")
 		Set tResourceInfo("Deploy") = 0
 		Set tResourceInfo("Processor") = $This
 		Set tResourceInfo("UnitTest") = 1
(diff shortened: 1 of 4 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -1016,7 +1016,8 @@ Method GetResolvedReferences(Output pReferenceArray, pLockedDependencies As %Boo
 			#dim tResource As %IPM.Storage.ResourceReference
 			Set tResource = ..Resources.GetNext(.tKey)
 			Quit:tKey=""
-			If '..HasScope(pPhases,tResource.Scope) {
+			Set tPhases = tResource.Attributes.GetAt("Phase")
```
isc-tleavitt, line 1019: This is just a thing for unit tests, not resources with scope more generally.

isc-tleavitt (changes requested):
%IPM.Lifecycle.Base changes are great; I think the unit test scope change should be done differently.

src/cls/IPM/Storage/Module.cls
```diff
@@ -1016,13 +1028,17 @@ Method GetResolvedReferences(Output pReferenceArray, pLockedDependencies As %Boo
...
```
isc-tleavitt, line 1031: We don't need this diff any more.

src/cls/IPM/Storage/Module.cls
```diff
@@ -550,7 +550,19 @@ ClassMethod HasScope(pPhases As %List, pScope As %String) [ Private ]
 		Quit 1
 	}
 	Set tPhases = $zcvt(pPhases,"L")
-	Quit $ListFind(tPhases,pScope)
+	// If pScope contains a comma, treat it as a comma-separated list
+    If $Find(pScope, ",") {
```
isc-tleavitt, line 554: Would it not ultimately be more readable / cleaner code to keep this as a $ListBuild list, check $ListValid, and iterate over the list with $ListNext rather than $Length/$Piece?

src/cls/IPM/ResourceProcessor/Test.cls
```diff
@@ -207,13 +207,14 @@ Method OnResolveChildren(ByRef pResourceArray) As %Status
 		Set tResourceInfo = tModuleName
 		Set tResourceInfo("Generated") = 0
 		Set tResourceInfo("Preload") = 0
-		Set tResourceInfo("Scope") = ..Phase
+		Set tResourceInfo("Scope") = $ListToString(..Phase, ",")
```
isc-tleavitt, line 210: Leave this as a $ListBuild list!

src/cls/IPM/Storage/Module.cls
```diff
@@ -549,8 +549,24 @@ ClassMethod HasScope(pPhases As %List, pScope As %String) [ Private ]
...
 		Quit 1
 	}
-	Set tPhases = $zcvt(pPhases,"L")
-	Quit $ListFind(tPhases,pScope)
+	Set tPhases = $ZConvert(pPhases, "L")
+	If $ListValid(pScope) {
+		// Iterate through each item in pScope list
+		Set i = 0
+		For {
+			Set i = $ListNext(pScope, i, scopeItem)
```
isc-tleavitt, line 557: General $ListNext pattern:
```
Set pointer = 0
While $ListNext(pScope,pointer,scopeItem) {
// do something with scopeItem
}
```

src/cls/IPM/Storage/Module.cls
isc-tleavitt: Still don't need this line of code, tPhases is unused.

isc-tleavitt (changes requested):
Sorry for the continual nitpicking here.

isc-tleavitt (approved):

isc-tleavitt:
@isc-jili confirmed the last commit's changes worked for the original problem use case.

src/cls/IPM/Storage/Module.cls
```diff
@@ -1022,7 +1034,11 @@ Method GetResolvedReferences(Output pReferenceArray, pLockedDependencies As %Boo
 			
 			kill tempArray
 			Set tSC = tResource.ResolveChildren(.tempArray)
-			Merge pReferenceArray(..Name) = tempArray
+			Set resourceScope = tResource.Scope
+			Set attrScope = $Get(tempArray("Scope"))
+			If ..HasScope(pPhases, resourceScope) && ..HasScope(pPhases, attrScope) {
```
isc-kiyer, line 1039: Nit: the first check if already done above so only the second check needs to be done here

PR #798, opened 2025-04-23
Fix #797 DetectPipCaller Returning Null for Windows

### LINKED ISSUE #797: DetectPipCaller Not Using Python Runtime Library Path in 2024.2+
The check:
`If $Data(info("CPF_PythonRuntimeLibrary"), tPyDylib) && tPyDylib {`
in the DetectPipCaller method of IPM.Lifecycle.Base is always false. This is because tPyDylib is a string and resolves as 0 logically. Need to instead check that it is not null or similar as currently the logic within this block is never run. This led to the following errror when installing packages in 2025.1:
ERROR! General exception Name 'Could not find a suitable pip caller. Consider setting UseStandalonePip and PipCaller' Code '' Data ''

isc-egabhart:
Fixes #797 by checking if tPyDylib not null

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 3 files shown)


isc-kiyer (approved):

isc-tleavitt:
The latest-preview failures are concerning but probably unrelated and thus not blocking.

PR #803, opened 2025-04-25
Fix: make CI pass on 2025.2

### LINKED ISSUE #778: Possible deadlock during IPM installation?
I think I may have run into a deadlock trying to install IPM. Locks show the following:

Image

and then for the system processes holding that lock on 

Image

Image

It seems the IPM installation is competing for locks on `^%SYS.NAMESPACE` and `^Config.Namespaces` that are held for processing stuff after an upgrade (see last global ref `^SYS("FirstRestartSinceInstall")`). Possibly some namespace mapping work by sharding that's not cleaning up as it should, which I'll pursue separately with the sharding dev team.

However, I believe the fact that the IPM install tries to take the lock with no timeout is a bug. It should come back after some reasonable timeout and tell me that it couldn't take the locks it wanted (and recommend restarting/reinstalling)

isc-tleavitt:
Fixes #778 

Solution (in theory) is just to purge cached queries so we aren't stuck with stale queries in the container when it boots.

### DIFF
--- a/.github/workflows/images.yml
+++ b/.github/workflows/images.yml
@@ -34,11 +34,6 @@ jobs:
                 then
                   continue
               fi
-              # Skip latest-preview due to #778
-              if [ "$tag" = "latest-preview" ];
-                then
-                  continue
-              fi
               images+='"'${n}:${tag}'",';
             done
           done;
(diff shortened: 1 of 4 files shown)


isc-tleavitt:
That does it - passing on 2025.2 again!

isc-egabhart (approved):

isc-kiyer (approved):

Issue #811, opened 2025-04-30
Intersystems Upgrade - Reinstall IPM Issue


jbglytec:
I am doing an upgrade on my HealthConnect and when installing IPM I am getting this error. An older version of IPM was install prior to the upgrade and in the %SYS namespace. The namespace I was attempting to install IPM in had 1 package (git-source-control) installed already.

Install command used:
s version="0.9.2" s r=##class(%Net.HttpRequest).%New(),r.Server="[pm.community.intersystems.com](http://pm.community.intersystems.com/)",r.SSLConfiguration="ISC.FeatureTracker.SSL.Config" d r.Get("/packages/zpm/"_version_"/installer"),$system.OBJ.LoadStream(r.HttpResponse.Data,"c")

![Image](https://github.com/user-attachments/assets/615ae74f-6779-4f93-acd1-27f3953ec289)

isc-tleavitt:
I think there are at least two separate things to fix here:
* The missing include file (theory: creation of a mapping was missed, which might be an upstream issue with %ZHSLIB.HealthShareMgr)
* When IPM installation fails at this level it needs to clean up after itself better

isc-tleavitt:
I incidentally reproduced something similar to this, but not even on an IRIS for Health instance (just IRIS community edition with 2024.1, IPM installed, upgrade to 2025.1, then reinstall IPM). Still working toward root cause.

isc-tleavitt:
Debugging methodology:
After loading errored installer, note directory in output:
```
2025-05-21 13:24:13 3 Evaluate: ${MODDIR} -> C:\Windows\Temp\mU7moFHh\
```

Then:
```
USER>do ##class(IPM.Installer).setup("C:\Windows\Temp\mU7moFHh\",3)
... (error happens again) ...

USER>zw %objlasterror
[...]

Issue #809, opened 2025-04-29
Version 0.10.X installation error


lauracysnet:
I have tried to install 0.10.X version (latest, and I have tried with 0.10.0 too), and I get this error after compiling all files:

```
USER>s version="latest" s r=##class(%Net.HttpRequest).%New(),r.Server="pm.community.intersystems.com",r.SSLConfiguration="ISC.FeatureTracker.SSL.Config" d r.Get("/packages/zpm/"_version_"/installer"),$system.OBJ.LoadStream(r.HttpResponse.Data,"c")
....
Compilation finished successfully in 3.294s.

Load finished successfully.

Compiling routine : %ZLANGC00.mac
Compiling routine : %ZLANGF00.mac2025-04-29 10:34:43 3 Evaluate: ${REGISTRY} -> https://pm.community.intersystems.com
2025-04-29 10:34:43 3 Evaluate: ${TRACKINGID} -> UA-126752118-1
2025-04-29 10:34:43 3 Evaluate: ${MODDIR} -> /tmp/dirOsRUJP/

[USER|ZPM]      Initialize START
Compilation started on 04/29/2025 10:34:47
Compiling class IPM.Installer
Compiling routine IPM.Installer.1
Compilation finished successfully in 0.054s.

[USER|ZPM]      Initialize SUCCESS
[USER|ZPM]      Reload START (/tmp/dirOsRUJP/)
[USER|ZPM]      Reload SUCCESS
[ZPM]   Module object refreshed.
[USER|ZPM]      Validate START
[USER|ZPM]      Validate SUCCESS
[USER|ZPM]      Compile START
[USER|ZPM]      Compile SUCCESS
[USER|ZPM]      Activate START
[ZPM]   Activate FAILURE
ERROR! File not copied: -32025-04-29 10:34:59 0 IPM.Installer: ERROR #5001: File not copied: -3
2025-04-29 10:34:59 0 IPM.Installer: Installation failed at 2025-04-29 10:34:59
2025-04-29 10:34:59 0 %Installer: Elapsed time 24.978745s
Time spent 24.985555
ERROR #5001: File not copied: -3
Detected 1 errors during load.
```

I can access zpm, but it doesn't show version:

```
USER > zpm
=============================================================================
|| Welcome to the Package Manager Shell (ZPM). Version:                    ||
|| Enter q/quit to exit the shell. Enter ?/help to view available commands ||
|| No registry configured                                                  ||
|| System Mode: <unset>                                                    ||
[...]

Issue #796, opened 2025-04-22
Installation issues with 0.10.x in an environment without flexible python runtime configured


isc-tleavitt:
ERROR! ObjectScript error: <OBJECT DISPATCH>InstallORASIfMissing+1^IPM.Installer.1 *Failed to Load Python: Check documentation and messages.log, Check CPF parameters:[PythonRuntimeLibrary,PythonRuntimeLibraryVersion], Check sys.path setup in: $INSTANCE/lib/python/iris_site.py

This happens in 2025.1 but not 2024.1 - probably related to Flexible Python Runtime changes.

jinlouISC:
At payer services, we have a few developers on Windows on I4H 2024.2 ran into this too when installing IPM v0.10.0. Their machines have python. 

The temporary workaround we discovered is to follow step 1-7 in https://docs.intersystems.com/irislatest/csp/docbook/DocBook.UI.Page.cls?KEY=GEPYTHON_prereqs#GEPYTHON_prereqs_flexible.

isc-tleavitt:
Desired behavior here:
Installation succeeds without Embedded Python; features that require Embedded Python fail helpfully at runtime.

PR #824, opened 2025-05-23
Fix/improve installation edge cases

isc-tleavitt:
* Install now works in containers with durable %SYS without jumping through extra hoops (#809)
* Install behaves properly if Flexible Python Runtime is not configured, with descriptive runtime errors on ORAS use (#796)
* Installing in an environment with a separate broken IRISSYS install no longer fails (#811)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-ARG BASE=containers.intersystems.com/intersystems/irishealth-community:2024.1
+ARG BASE=irepo.intersystems.com/intersystems/iris-community:2025.1
 FROM ${BASE}
 
 ARG REGISTRY=https://pm.community.intersystems.com
--- a/preload/cls/IPM/Installer.cls
+++ b/preload/cls/IPM/Installer.cls
@@ -147,7 +147,7 @@ ClassMethod InstallORASIfMissing()
       }
     }
   } else {
-    Write !!,"WARNING: Flexible Python Runtime is not configured on this instance. Not all IPM features will work. If you wish to use ORAS registries, configure Flexible Python Runtime then reinstall IPM.",!
+    Write !!,"WARNING: Flexible Python Runtime is not configured on this instance. Not all IPM features will be available. If you wish to use ORAS registries, configure Flexible Python Runtime and then reinstall IPM.",!
   }
   Quit $$$OK
 }
--- a/preload/cls/IPM/Installer.cls
+++ b/preload/cls/IPM/Installer.cls
@@ -131,19 +131,23 @@ ClassMethod MapIfLegacy()
 
 ClassMethod InstallORASIfMissing()
 {
-  if '..ORASInstalled() {
-    // Need to run in a background process to pick up new %IPM.Storage.Module definition
-    Write !,"Installing ORAS Python components..."
-    Job ..InitializeAndSignal()
-    Set result = $System.Event.WaitMsg("",30)
-    Set $ListBuild(msg,sc) = result
-    If (msg = 0) {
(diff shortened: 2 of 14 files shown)


tests/sandbox/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASE=containers.intersystems.com/intersystems/iris-community:2024.1
+ARG BASE=irepo.intersystems.com/intersystems/iris-community:2025.1
```
isc-kiyer, line 1: Same issue here with irepo vs containers

src/cls/IPM/Main.cls
```diff
@@ -1083,7 +1083,12 @@ ClassMethod GetListModules(pNamespace As %String = {$Namespace}, pSearch As %Str
       Set ns = $Order(listNS(ns)) 
       Quit:ns=""
       Kill listModules
-      Do ..GetListModules(ns, pSearch, .listModules, pExtraFields, pRepository)
+      Try {
+        // It is possible that GetListModules will fail in some namespaces. Allow it to do so silently.
```
isc-kiyer, line 1087: Enumerate at least one reason it could fail and why it is safe to ignore. Should we at least report which namespaces it failed in but we expected IPM to be present in?

isc-kiyer (changes requested):
@isc-tleavitt a few comments but looks good overall!

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG BASE=containers.intersystems.com/intersystems/irishealth-community:2024.1
+ARG BASE=irepo.intersystems.com/intersystems/iris-community:2025.1
```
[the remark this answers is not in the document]
isc-tleavitt: (Also, having trouble with containers.intersystems.com internally for some reason.)

preload/cls/IPM/Installer.cls
```diff
@@ -153,6 +157,17 @@ ClassMethod InitializeAndSignal()
   do $System.Event.Signal($zparent,sc)
 }
 
+ClassMethod ValidatePythonRuntime() As %Boolean
```
isc-kiyer, line 160: @isc-tleavitt Final comment: While doing this, worth checking the min python version and notifying the user if its not meeting the minimum requirement to use oras?

preload/cls/IPM/Installer.cls
```diff
@@ -19,22 +19,27 @@ XData PM [ XMLNamespace = INSTALLER ]
 <Default Name="TRACKINGID" Value="UA-126752118-1" />
 
 <Namespace Name="${CURNS}" Code="${CURNSRoutineDB}" Data="${CURNSGlobalDB}" Create="no" Ensemble="0">
-  <Configuration>
-      <ClassMapping Package="%IPM" From="${CURNSRoutineDB}" />
-      <RoutineMapping Routines="%IPM.*" Type="ALL" From="${CURNSRoutineDB}" /> 
-  </Configuration>
+  <!-- Support use in a bootstrapped ^INSTALL context, detected by presence of %Log -->
+  <If Condition="($Data(%Log)#2)">
```
isc-kiyer, line 23: Should this happen when %Log is not defined? Currently a regular install (not in ^INSTALL) would not create the mappings since %Log is not defined right?

isc-tleavitt: This is just plain wrong. Going to kill it.

isc-tleavitt:
Should be good here. Note, I wasn't able to test the negative case (just the positive case) for the version check due to a really weird upstream IRIS issue (internal ref: DP-441981)

PR #826, opened 2025-05-30
Fix issue 816: zpm "install" command not pulling specified versions of artifacts

isc-jili:
Fix issue 816 zpm "install" command not pulling specified versions of artifacts. Do so by matching on the exact build metadata in the version string if specified in the install command.

### DIFF
--- a/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
+++ b/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
@@ -92,6 +92,10 @@ Method Evaluate(pVersion As %IPM.General.SemanticVersion) As %Boolean
       Quit:((val1 = "*")||(val2 = "*"))
       Set tEquals = tEquals && (val1 = val2)
     }
+	// Compare build if both versions specify it
+  	If (pVersion.Build '= "") && (..Build '= "") {
+    	Set tEquals = tEquals && (pVersion.Build = ..Build)
+  	}
   }
 	If (+##class(%IPM.Repo.UniversalSettings).GetValue("SemVerPostRelease") = 0) || ($Extract(..Operator,1) = "=")  {
 		// If SemVerPostRelease is disabled, then we should not consider post-release versions as equal to the base version
--- a/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
+++ b/src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
@@ -85,7 +85,7 @@ Method Evaluate(pVersion As %IPM.General.SemanticVersion) As %Boolean
 {
 	Set tEquals = (..Operator [ "=")
   If tEquals {
-    For part = "Major", "Minor", "Patch", "Prerelease" {
+    For part = "Major", "Minor", "Patch", "Prerelease", "Build" {
       Set val1 = $Property(pVersion, part)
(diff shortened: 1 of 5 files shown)


tests/integration_tests/Test/PM/Integration/OrasVersionWithBuild.cls
```diff
@@ -0,0 +1,71 @@
...
+    // Publish the clean version
+    Do ..AssertVersion(..#CleanVersion)
+    Set sc = ##class(%IPM.Main).Shell("publish "_..#Module_" -r oras -verbose -only")
+    Do $$$AssertStatusOK(sc, "Published clean prerelease version")
+
+    // Uninstall all local copies
+    Do $$$AssertStatusOK(##class(%IPM.Main).Shell("uninstall "_..#Module))
+
+    // Install the version with build explicitly
+    Set sc = ##class(%IPM.Main).Shell("install oras/"_..#Module_" "_..#BuildVersion)
```
isc-kiyer, line 41: Would be good to add a test case for multiple modules published with different build versions and then attempt installing an older build version (not the latest published one)

isc-jili:
As I'm wrapping up this work, noting down a few relevant items:
- In the Steps to reproduce, the command zpm "install isc/intersystems_payer_services_hp_dtr/1.1.0-snapshot_CMS-0057-WIP-15090__2024.2" is incorrect. The correct command to specify install of a particular version is zpm "install isc/intersystems_payer_services_hp_dtr 1.1.0-snapshot+CMS-0057-WIP-15090".
- It turns out that IPM intentionally was ignoring the build metadata in a version string (ie. the CMS-0057-WIP-15090 part) so this behavior started off as expected behavior. This is because according to the semantic versioning documentation https://semver.org/#spec-item-10 "Build metadata MUST be ignored when determining version precedence".
- We decided via a discussion with DevOps and Tim L. that if the build metadata is specified in an install command's version, then we should be taking that specific version of the artifact. This PR encompasses that change.

isc-tleavitt (dismissed):
Nitpick: we have true unit tests around everything Semantic Versioning, and it would be great to have an update there as well.

That said, the existing test is sufficient, and I don't want to block this further.

src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
```diff
@@ -85,12 +85,14 @@ Method Evaluate(pVersion As %IPM.General.SemanticVersion) As %Boolean
...
   If tEquals {
-    For part = "Major", "Minor", "Patch", "Prerelease" {
+    For part = "Major", "Minor", "Patch", "Prerelease", "Build" {
       Set val1 = $Property(pVersion, part)
       Set val2 = $Property($this, part)
-      Quit:((val1 = "")||(val2 = ""))
       Quit:((val1 = "*")||(val2 = "*"))
-      Set tEquals = tEquals && (val1 = val2)
+	  // Enforce exact match if comparator specifies the part (e.g., Prerelease or Build)
+	  If val2 '= "" {
```
isc-kiyer, line 93: Why not keep the original quit in which checks for if either val1 or val2 is empty?

isc-tleavitt (approved):
@isc-jili thanks for the extra attention on testing, glad it was worthwhile!

isc-tleavitt (approved):

PR #830, opened 2025-06-04
Fix getting relative path of items when Directory="" on resource

### LINKED ISSUE #829: Export fails if resources have Directory=""
Say I want my code repository to look like this:
```
src/
    Package/
        Class.cls
        Class2.cls
```
If I follow the hints in the documentation my module.xml will include `<SourcesRoot>src</SourcesRoot>` and each of my resources will have the attribute `Directory=""`. But if I do that, IPM load fails on the export step:
```
ERROR #5002: ObjectScript error: <NOTOPEN>ExportUDL+26^%apiOBJ
exporting new version of Package.Class.CLS to C:\Users\pbarton\Projects\TestCoverage\src\\Package\Class.cls
```

isc-pbarton:
Fixes #829

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.10.3] - Unreleased
 
 ### Fixed
+- #829: Fixed export of resources with null Directory attribute
 
 ## [0.10.2] - 2025-06-04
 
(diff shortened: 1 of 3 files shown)


isc-kiyer (commented):
@isc-pbarton looks good! Please add a changelog update

isc-tleavitt (approved):
On principle need another review too since I contributed, but this looks good.

PR #834, opened 2025-06-06
Fix: Selectively undeploying classes within a package

isc-jlechtne:
- Issue seems to happen in %IPM.ResourceProcessor.Default.Document::OnPhase() when pPhase="Reload"
- In this scenario, the if statement: "If ##class(%IPM.Utils.File).Exists(.tResourcePath)" evaluates to true for the package level resource.
        - This stops us from reaching the "If ..ResourceReference.Deploy" code block, where the deployed files are loaded in.
        - Since at least one class isn't deployed, the class directory exists so stops execution. Issue is there are other resources in the 
          package that are deployed that we didn't handle

Fix:
- Swapped order of if-statements in OnPhase() to run steps if deployed whether or not the resource path exists to account for this case described above.
- 2 passes over the resource list in %Reload(). First pass: to split up resources between deployed and not deployed. Second pass: Process the resources--deployed first then not deployed.
        - Loading a non-deployed resource should override anything that was one with deployed resources

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - #829: Fixed export of resources with null Directory attribute
 - #832: Fixed places that export to possibly-nonexistent directories by adding /createdirs (needed in 2025.2+)
-- #834: Fixed selectively undeploying classes within a package
+- #823: Fixed selectively undeploying classes within a package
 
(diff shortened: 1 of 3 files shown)


isc-jlechtne:
Realized my local changes were off of the v0.10.x branch, not main. Closing this request/branch and creating a new one with changes to the main branch to avoid accidentally reverting unrelated changes.

New pull request: https://github.com/intersystems/ipm/pull/835

PR #835, opened 2025-06-06
Fix: Selectively undeploying classes within a package

isc-jlechtne:
- Issue seems to happen in %IPM.ResourceProcessor.Default.Document::OnPhase() when pPhase="Reload"
- In this scenario, the if statement: "If ##class(%IPM.Utils.File).Exists(.tResourcePath)" evaluates to true for the package level resource.
        - This stops us from reaching the "If ..ResourceReference.Deploy" code block, where the deployed files are loaded in.
        - Since at least one class isn't deployed, the class directory exists so stops execution. Issue is there are other resources in the
          package that are deployed that we didn't handle

Fix:
- Swapped order of if-statements in OnPhase() to run steps if deployed whether or not the resource path exists to account for this case described above.
- 2 passes over the resource list in %Reload(). First pass: to split up resources between deployed and not deployed. Second pass: Process the resources--deployed first then not deployed.
        - Loading a non-deployed resource should override anything that was one with deployed resources

### DIFF
--- a/src/cls/IPM/Lifecycle/Base.cls
+++ b/src/cls/IPM/Lifecycle/Base.cls
@@ -642,10 +642,29 @@ Method %Reload(ByRef pParams) As %Status
 		
 		// Standard resource processing
 		Set tSC = $$$OK
+		Set deployedResources = []
+		Set nonDeployedResources = []
 		Set tKey = ""
 		For {
 			Set tResource = ..Module.Resources.GetNext(.tKey)
 			Quit:tKey=""
+			If tResource.Deploy {
+				do deployedResources.%Push(tResource)
+			}
+			Else {
+				do nonDeployedResources.%Push(tResource)
+			}
+		}
+		// Process deployed resources first
+		set iter = deployedResources.%GetIterator()
+		while iter.%GetNext(, .tResource) {
+			If $IsObject(tResource.Processor) {
+				Set tSC = $$$ADDSC(tSC,tResource.Processor.OnPhase("Reload",.pParams))
+			}
+		}
+		// Then process non deployed resources
+		set iter = nonDeployedResources.%GetIterator()
+		while iter.%GetNext(, .tResource) {
 			If $IsObject(tResource.Processor) {
 				Set tSC = $$$ADDSC(tSC,tResource.Processor.OnPhase("Reload",.pParams))
 			}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - #829: Fixed export of resources with null Directory attribute
 - #832: Fixed places that export to possibly-nonexistent directories by adding /createdirs (needed in 2025.2+)
+- #823: Fixed selectively undeploying classes within a package
 
 ## [0.10.2] - 2025-06-04
 
--- a/src/cls/IPM/ResourceProcessor/Default/Document.cls
+++ b/src/cls/IPM/ResourceProcessor/Default/Document.cls
@@ -170,10 +170,6 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
                   }	
                 }
               }
-              If ##class(%IPM.Utils.File).Exists(.tResourcePath) {
-                Set tSC = $$$OK
-                Quit
-              }
               If ..ResourceReference.Deploy {
                 If tVerbose {
                   Write !, "Loading deployed resource from Studio project: "_..ResourceReference.Name
@@ -188,6 +184,10 @@ Method OnPhase(pPhase As %String, ByRef pParams, Output pResourceHandled As %Boo
                   Set $$$DeployedProjectInstalled(tDeployedProjectPath) = ""
                 }
                 Return $$$OK
+              }
+              If ##class(%IPM.Utils.File).Exists(.tResourcePath) {
+                Set tSC = $$$OK
+                Quit
               }
               Set tSC = $$$ERROR($$$GeneralError, "Resource path '" _ tResourcePath _ "' not found")
             }


tests/integration_tests/Test/PM/Integration/SelectiveDeployResource.cls
```diff
@@ -0,0 +1,113 @@
...
+        // Load module from package
+        Set tSC = ##class(%IPM.Main).Shell("load -v "_packageDir_".tgz")
+        Do $$$AssertStatusOK(tSC,"Installed selectively-undeploy successfully.")
+        Set tSC = ##class(SelectivelyUndeploy.D1).Main()
+        Do $$$AssertStatusOK(tSC, "Class SelectivelyUndeploy.D1 installed correctly")
+        Set tSC = ##class(SelectivelyUndeploy.D2).Main()
+        Do $$$AssertStatusOK(tSC, "Class SelectivelyUndeploy.D2 installed correctly")
+        
+        // Check deployed status after loading
+        Set tSC = ##class(%Dictionary.CompiledClass).%OpenId("SelectivelyUndeploy.D1").Deployed
```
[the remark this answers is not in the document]
isc-eneil: Same in the test below.

tests/integration_tests/Test/PM/Integration/SelectiveDeployResource.cls
```diff
@@ -0,0 +1,113 @@
...
+        Set tSC = ##class(%IPM.Main).Shell("uninstall selectively-deploy")
+        Do $$$AssertStatusOK(tSC,"Uninstalled selectively-deploy successfully.")
+        Set tSC=##class(%Dictionary.CompiledClass).%ExistsId("SelectivelyDeploy.D1")
+        Do $$$AssertNotTrue(tSC, "Class SelectivelyDeploy.D1 DNE")
+        Set tSC=##class(%Dictionary.CompiledClass).%ExistsId("SelectivelyDeploy.D2")
+        Do $$$AssertNotTrue(tSC, "Class SelectivelyDeploy.D2 DNE")
+
+        // Load module from package
+        Set tSC = ##class(%IPM.Main).Shell("load -v "_packageDir_".tgz")
+        Do $$$AssertStatusOK(tSC,"Installed selectively-deploy successfully.")
```
isc-eneil, line 97: @isc-jlechtne Nit: Loaded, not installed (that applies wherever you've used `load` and not `install`)

isc-eneil (changes requested):
@isc-jlechtne Left a couple of comments. Logic looks good to me!

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -640,13 +640,29 @@ Method %Reload(ByRef pParams) As %Status
 			Set ^Sources("MODULE",..Module.Name) = tAbsolutePrefix_tRoot
 		}
 		
-		// Standard resource processing
-		Set tSC = $$$OK
+		// Process resources based on their package depth
```
isc-eneil, line 643: @isc-jlechtne Nit: I like explanation that "(higher level first so more granular resource will override)" so I'd actually keep that!

tests/integration_tests/Test/PM/Integration/SelectiveDeployResource.cls
```diff
@@ -0,0 +1,113 @@
...
+
+        // Package module
+        Set packageDir = $$$FileTempDir_"selectively-deploy"
+        Set sc = ##class(%IPM.Main).Shell("selectively-deploy package -only -v -DPath="_packageDir)
+        Do $$$AssertStatusOK(sc, "Successfully packaged module to "_packageDir)
+
+        // Uninstall module
+        Set tSC = ##class(%IPM.Main).Shell("uninstall selectively-deploy")
+        Do $$$AssertStatusOK(tSC,"Uninstalled selectively-deploy successfully.")
+        Set isClassExist=##class(%Dictionary.CompiledClass).%ExistsId("SelectivelyDeploy.D1")
```
isc-eneil, line 90: Grammar nit: `doesClassExist`

isc-eneil (changes requested):

isc-jili (approved):

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -640,13 +640,29 @@ Method %Reload(ByRef pParams) As %Status
...
-		Set tSC = $$$OK
+		// Process resources based on their package depth (higher level first so more granular resource will override)
+		// e.g. <Resource Name="Depth1.PKG"/> will be processed before <Resource Name="Depth1.Depth2.CLS"/>
 		Set tKey = ""
 		For {
 			Set tResource = ..Module.Resources.GetNext(.tKey)
 			Quit:tKey=""
 			If $IsObject(tResource.Processor) {
+				Set depth = $LENGTH(tResource.Name, ".")
+				If $GET(resourceList(depth)) = "" {
```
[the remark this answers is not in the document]
isc-jlechtne: This works, however, there are some cases where the resource extension causes the sorting to get messed up. Take for example the 2 resources `MyPackage.PKG` and `MyPackage.Class1.CLS`. Since 'C' (from ".Class1") comes before 'P' (from ".PKG"), the array would end up looking like:
```
resourceList("MtPackage.Class1.CLS")
resourceList("MyPackage.PKG")
```
My idea around this would be to strip everything after the last '.' out of the name key (or keep as is if there are no periods). I'm curious if there's and edge case where this approach wouldn't work, nothing is coming to my mind right now.

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -640,15 +640,49 @@ Method %Reload(ByRef pParams) As %Status
...
+		// Process resources based on their package depth (higher level first so more granular resource will override)
+		// e.g. <Resource Name="Depth1.PKG"/> will be processed before <Resource Name="Depth1.Depth2.CLS"/>
+		Set nonDocumentResources = []
 		Set tKey = ""
 		For {
 			Set tResource = ..Module.Resources.GetNext(.tKey)
 			Quit:tKey=""
 			If $IsObject(tResource.Processor) {
+				// Strip extension from resource name as that caused improper sorting (i.e. MyPackage.MyClass.CLS gets sorted before MyPackage.PKG)
+				Set tProcessorClassName = tResource.Module.Lifecycle.GetDefaultResourceProcessor(tResource.Name)
```
isc-kiyer, line 652: To figure out if the processor extends Default.Document, can simple do the following: 
`If tResource.Processor.%Extends("%IPM.ResourceProcessor.Default.Document")`

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -640,15 +640,49 @@ Method %Reload(ByRef pParams) As %Status
...
+				if (tProcessorClassName '= "") {
+				Set tProcessorClass = ##class(%Dictionary.CompiledClass).%OpenId(tProcessorClassName)
+					Set extendsDefaultDocument = $FIND(tProcessorClass.Super, "%IPM.ResourceProcessor.Default.Document")
+				}
+				else {
+					Set extendsDefaultDocument = "0"
+				}
+				if (extendsDefaultDocument) {
+					Set tResourceName = $Piece(tResource.Name,".",1,*-1)
+					Set:tResourceName="" tResourceName = tResource.Name
```
isc-kiyer, line 662: Would tResource ever be empty? I think for a default doc, its always going to include at least one period. Also, nit: avoid using post-conditionals (exception is for quit in a loop)

isc-kiyer (changes requested):
@isc-jlechtne few small comments and a question. To get this done and merged, I recommend creating a separate issue and PR for generalizing this lifecycle processing order

isc-eneil (approved):
@isc-jlechtne Looks good! I don't think this should stop the merge, but there's a little difference in style (your changes all use lowercase ObjectScript, but the IPM codebase capitalizes commands).

Issue #848, opened 2025-07-07
Update Resource Processing Order


isc-jlechtne:
This issue came out as part of the result of the selectively undeploying resources bug fix (#823).

Part of that fix included updating the resource processing in %Reload to first process document resources by depth order and then non-document resources after. That similar resource processing order should be changed for all lifecycle phases that use resource processing in this manner.

PR #849, opened 2025-07-07
Initial commit. Separating out list generation into a helper method

isc-jlechtne:
Addresses #848 

Idea is to create a helper method which compiles all resources into one list in their proper processing order. Lifecycle methods can then use this `orderedResourceList.GetNext(.tKey)` instead of `..Module.Resources.GetNext(.tKey)`

### DIFF
--- a/src/cls/IPM/Storage/Module.cls
+++ b/src/cls/IPM/Storage/Module.cls
@@ -85,6 +85,55 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
 
 Property Deployed As %Boolean(XMLPROJECTION = "Element");
 
+/// This method iterates over resources of a module and returns a list of objects in the correct order that resources should be processed
+/// Process resources based on their package depth (higher level first so more granular resource will override)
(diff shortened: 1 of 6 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,55 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
 
 Property Deployed As %Boolean(XMLPROJECTION = "Element");
 
+/// This method iterates over resources of a module and returns a list of objects in the correct order that resources should be processed
+/// Process resources based on their package depth (higher level first so more granular resource will override)
```
isc-eneil, line 89: @isc-jlechtne Isn't the package depth processing just for Document resources, then the rest of resources are processed in the order in which they are declared in the module's module.xml?

isc-eneil (changes requested):
I suggested one small method description change, but otherwise looks good to me!

isc-jili (dismissed):

isc-jili (approved):

isc-kiyer (approved):

PR #851, opened 2025-07-25
Small improvements for the benefit of contributors

gjsjohnmurray:
- Use public registry when getting Docker images.
- Avoid recommending a superseded extension.
- Remove need to change settings.json if port had to be altered in docker-compose.yml
- Set the container up to produce test coverage reports in VS Code using the InterSystems Testing Manager extension


isc-tleavitt (approved):
@gjsjohnmurray sorry this has sat for a ridiculously long time!

This is uncontroversial and should just be merged.

PR #854, opened 2025-07-30
Fix #820: $(packagename) not working for FileCopy

### LINKED ISSUE #820: No replacement for ${packagename} in FileCopy
When using the following resource in module.xml, the ${packagename} parameter is not substituted, although all other parameters are correctly replaced.

`<FileCopy Name="/src/assets/" InstallDirectory="{$mgrdir}/assets/${namespaceRoutineDB}/${packagename}"/>`

isc-dchui:
Fixes #820. 

Should also now be working for all resource processors and not just FileCopy.


isc-kiyer (approved):
@isc-dchui Once this PR is merged, please also update 0.9.x with the fix

isc-jili (approved):

PR #852, opened 2025-07-28
Upgrade framework jlechtne

isc-jlechtne:
Initial pass of the update framework with creating new VersionBase, UpdateStep, and Controller classes as well as updates to %IPM.Storage.Module

PR #855, opened 2025-07-31
IPM Update Framework new classes

isc-jlechtne:
Continuation of https://github.com/intersystems/ipm/pull/852, but in a new, clean branch with all and only update framework changes.

Issue: #550


src/cls/IPM/Storage/UpdateStep.cls
```diff
@@ -0,0 +1,102 @@
...
+	methodName As %String = "",
+	mirrored As %Boolean = 1,
+	hostName As %String = "",
+	instanceName As %String = "") As %Status
+{
+    set ..ClassName = className
+    set ..MethodName = methodName
+    set ..Mirrored = mirrored
+    if hostName = "" {
+        set ..HostName = ##class(%SYS.System).GetNodeName()
```
isc-kiyer, line 35: Rather than doing this here, could simply have this be InitialExpression for HostName and InstanceName properties

src/cls/IPM/Storage/UpdateStep.cls
```diff
@@ -0,0 +1,102 @@
...
+
+/// Returns %IPM.Storage.UpdateStep object corresponding to unique index in storage table
+/// If object does not exist, this will return an empty object
+ClassMethod GetUpdateStep(
+	className As %String,
+	methodName As %String,
+	hostName As %String,
+	instanceName As %String) As %IPM.Storage.UpdateStep
+{
+    do ..UpdateClassMethodIdxExists(className, methodName, hostName, instanceName, .stepId)
```
isc-kiyer, line 53: This should be an `if` and only if it returns true, return the %OpenId(). If it doesn't exist, return ""

src/cls/IPM/Utils/Update/Controller.cls
```diff
@@ -0,0 +1,85 @@
...
+    set tSC = $$$OK
+    set tInitTLevel = $tlevel
+    try {
+        tstart
+        // For each base module dependency
+        set processList = ..GetModulesToUpdateOrderedList(module)
+
+        set ptr = 0
+        while $listnext(processList, ptr, moduleName) {
+            set nextModule = ##class(%IPM.Storage.Module).NameOpen(moduleName)
```
isc-kiyer, line 30: Should open with concurrency level of 0 here to not acquire exclusive lock since you are just reading from the object

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,129 @@
...
+{
+    set sc = $$$OK
+    try {
+        set iter = methodList.%GetIterator()
+        while iter.%GetNext(, .methodName) {
+            set stepObj = ##class(%IPM.Storage.UpdateStep).GetUpdateStep(className, methodName, hostName, instanceName)
+            if '$isobject(stepObj) {
+                set stepObj = module.AddUpdateStep(className, methodName, mirrored)
+            }
+            if 'newStepToApply && 'stepObj.TimeStampEnd {
```
isc-kiyer, line 93: What is the purpose of newStepToApply? Also, it is never initialized so you would get an error on the first iteration.

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,129 @@
...
+                set newStepToApply = 1
+            }
+            if newStepToApply {
```
isc-kiyer, line 96: Need to handle the case somewhere in here for when we find a step for a given major version that hasn't been run and then any subsequent steps are to be re-run. 
@isc-jlechtne @isc-jili @isc-eneil  Not sure if this is overkill but how about separating the persistence of the contents of an update step to the actual running of it? So UpdateStep would have the class name, method name and whether it should run on only primary and then a child table would have the various runs for each update step? In most cases there would be a single child record for each record in UpdateStep for steps to be run only on primary and 1 per mirror member. But if there are errors, we have a new entry for each time it has run and the error reported to see a full history of errors. Thoughts?

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,69 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
 
+Relationship UpdateSteps As %IPM.Storage.UpdateStep(XMLPROJECTION = "NONE") [ Cardinality = many, Inverse = ModuleHome ];
+
+Property UpdatePackage As %String(MAXLEN = 255);
+
+/// Add an update step and connect it to this module via relationship
+Method AddUpdateStep(
+	className As %String,
+	methodName As %String,
+	mirrored As %Boolean) As %IPM.Storage.UpdateStep
```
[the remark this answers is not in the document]
isc-jlechtne: Yes I had that same thought as well, but more so for the VersionBase GetMirroredMethodNames()/GetNonMirroredMethodNames() methods since that was a point of confusion when we were discussing things yesterday.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,69 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+
+        for version=1:1:..Version.Major {
+            set className = ..UpdatePackage_".V"_version
+            if '##class(%Dictionary.ClassDefinition).%ExistsId(className) {
+                // TODO: Log that this class is being skipped?
+                continue
+            }
+
+            // Seed MirrorMethodNames steps
+            set mirroredMethodNames = $classmethod(className, "GetOrderedMirroredMethodNames")
```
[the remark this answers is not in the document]
isc-jlechtne: I agree (as mentioned in response to your comment in %IPM.Storage.Module.cls)

src/cls/IPM/Storage/Module.cls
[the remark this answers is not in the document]
isc-eneil: I guess this is moot given our discussion of the UpdateStep storage model change.

src/cls/IPM/Utils/Update/Controller.cls
```diff
@@ -0,0 +1,85 @@
+Include %IPM.Common
+
+Class %IPM.Utils.Update.Controller
+{
+
+/// Iterates through a module and its dependencies and applies their update steps that haven't been run up through the current version
+/// Starts with the most dependent module and works its way backwards to the base module we are calling "update" on
+ClassMethod ApplyUpdatesInOrder(
```
isc-jili, line 8: Would it be possible to have this method take another argument toVersion? This is to allow the -version flag for the update command to specify a particular version to update to.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,69 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+{
+	set newStep = ##class(%IPM.Storage.UpdateStep).%New(className, methodName, mirrored)
+	$$$ThrowOnError(..UpdateSteps.Insert(newStep))
+	$$$ThrowOnError(..%Save())
+	return newStep
+}
+
+/// Seed update steps for all versions prior to the current major version by setting TimeStampEnd to current time
+/// Steps that already have TimeStampEnd defined (already been run) will be skipped
+Method SeedUpdateSteps() As %Status [ Internal ]
```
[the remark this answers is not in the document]
isc-jili: That would be sufficient! Thank you.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,69 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+{
+    set sc = $$$OK
+    try {
+        set hostName = ##class(%SYS.System).GetNodeName()
+        set instanceName = ##class(%SYS.System).GetInstanceName()
+
+        for version=1:1:..Version.Major {
+            set className = ..UpdatePackage_".V"_version
+            if '##class(%Dictionary.ClassDefinition).%ExistsId(className) {
+                // TODO: Log that this class is being skipped?
```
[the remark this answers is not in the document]
isc-jlechtne: This part kind of deleted itself when implementing the class query you mentioned in the previous comment, so I agree.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,280 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+}
+
+/// Helper method for ApplyStepsInOrder()
+/// Iterates over a method name list and calls their update steps.
+/// If an update step does not already exist, a new one gets created.
+/// Once reaching the first method that hasn't been run on the instance (no TimeStampEnd value),
+/// newStepToApply gets marked as true to tell us to run all remaining steps in the list
+/// Can carry this over to subsequent calls by passing the newStepToApply variable to those ExecuteMethodList() calls
+/// i.e. We find a new step to apply in the mirroredMethodNames list so we know to execute all steps in the nonMirroredMethodNames list
+Method ExecuteMethodList(
```
isc-jlechtne, line 224: Update name to make it clear this is UpdateStep specific

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,280 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+        write !, "Releasing lock on ^IPM.Update("_..Name_")"
+    }
+    $$$UnlockUpdateModule(..Name)
+    quit tSC
+}
+
+/// Get list of dependency modules in order of how we want to update them.
+/// We want to perform updates in reverse order of dependencies depth.
+/// i.e. Call update steps for highest depth dependent module all the way up back the base module being updated
+Method GetModulesToUpdateOrderedList() As %List [ Private ]
```
isc-jili, line 142: I also put this comment on the previous version of this review, but pasting it forward:
Just making a note here that from offline discussion with @isc-kiyer this dependency resolution logic may not be needed since as part of running through the lifecycle phases, IPM will apply the update to each dependent module in the correct order. And because %Update() will call ApplyUpdatesInOrder() for each module, all the dependent modules will have update steps applied.

isc-kiyer (changes requested):
@isc-jlechtne another round of comments up!

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,280 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+Method GetAddUpdateStep(
+	className As %String,
+	methodName As %String,
+	primaryOnly As %Boolean) As %IPM.Storage.UpdateStep.Base
+{
+    if primaryOnly {
+        set stepObj = ##class(%IPM.Storage.UpdateStep.PrimaryOnly).GetUpdateStep(className, methodName)
+        if '$isobject(stepObj) {
+            set stepObj = ##class(%IPM.Storage.UpdateStep.PrimaryOnly).%New(className, methodName)
+            // TODO: Replace this with actual foreign key
```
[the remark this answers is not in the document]
isc-jlechtne: For this update, I included ForeignModule as a string of the module name. The TODO is to remind me to go back and change that datatype to be a foreign key instead of string.

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,48 @@
...
+ClassMethod GetUpdatePackageClassList(
+	updatePackage As %String,
+	Output updatePackageClassList) As %Status
+{
+    set sc = $$$OK
+    try {
+        set tRes = ##class(%SQL.Statement).%ExecDirect(,
+            "SELECT * "_
+            "FROM %Dictionary.CompiledClass "_
+            "WHERE Name LIKE '"_updatePackage_".%'")
```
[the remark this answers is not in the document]
isc-eneil: And since you're really checking for V# classes, it should be

```
WHERE Name %STARTSWITH '"_updatePackage_".V%'"
```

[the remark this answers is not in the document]
isc-eneil: (I just added a "V" after the .)

src/cls/IPM/Storage/UpdateStep/Base.cls
```diff
@@ -0,0 +1,37 @@
...
+
+Method %OnNew(
+	className As %String = "",
+	methodName As %String = "",
+	hostName As %String = "",
+	instanceName As %String = "") As %Status
+{
+    set ..ClassName = className
+    set ..MethodName = methodName
+    if hostName = "" {
```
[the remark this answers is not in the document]
isc-jlechtne: No, I've had a similar thought because the stuff I've implemented here does let them default to the values below. Left this in for a hypothetical future case where the user would want to specify host and instance on step creation that's not these default values. Personally, I haven't thought of a specific use case like this so, if nobody else does either, I'm fine with getting rid of the conditional and setting to GetNodeName() and GetInstanceName() in all cases.

isc-eneil (changes requested):
Left another round of comments @isc-jlechtne !

src/cls/IPM/Storage/UpdateStep/PrimaryOnly.cls
```diff
@@ -0,0 +1,58 @@
+Class %IPM.Storage.UpdateStep.PrimaryOnly Extends (%Persistent, %IPM.Storage.UpdateStep.Base)
+{
+
+Index UpdateClassMethodIdx On (ClassName, MethodName) [ Unique ];
```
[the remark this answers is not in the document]
isc-jlechtne: After some investigation, we found that index cannot be moved to the super class in the way that the exists method is generated. Have to keep the indices in the subclasses as they are here.

src/cls/IPM/Storage/Module.cls
```diff
@@ -167,20 +168,17 @@ Method GetModulesToUpdateOrderedList() As %List [ Private ]
     return processList
 }
 
+/// For a given version, apply update steps listed in the module's update package version class
+/// First run are steps that can be executed on any mirror member, followed by those that can only be run on the primary
+/// Previously executed steps will be skipped until a new one has been found.
+/// If a step has not been run, all subsequent steps will be run, regardless of if they already have or not.
```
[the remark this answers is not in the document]
isc-jlechtne: @isc-eneil My understanding was that we ran everything after the first non-seeded step for every version but treated each individual version standalone. So will run all subsequent steps for a module version if new step found but won't carry over that flag between versions.

i.e. New step in v2 so we run the last X steps but won't run any steps in v3 until we find the first non-seeded step for v3.

src/cls/IPM/Storage/UpdateStep/PrimaryOnly.cls
```diff
@@ -0,0 +1,85 @@
...
+
+/// Home module which this update step belongs to
+Property HomeModule As %IPM.Storage.Module;
+
+/// ForeignKey pointing to the module which this update step belonds to
+ForeignKey HomeModKey(HomeModule) References %IPM.Storage.Module() [ OnDelete = cascade ];
+
+/// Description
```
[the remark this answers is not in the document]
isc-jlechtne: haha good point. Must be leftover from an autocomplete

src/cls/IPM/Storage/UpdateStep/AnyMirrorMember.cls
```diff
@@ -0,0 +1,58 @@
+Class %IPM.Storage.UpdateStep.AnyMirrorMember Extends (%Persistent, %IPM.Storage.UpdateStep.Base)
+{
+
+Index UpdateClassMethodIdx On (ClassName, MethodName) [ Unique ];
+
+/// Home module which this update step belongs to
+Property HomeModule As %IPM.Storage.Module;
+
+/// ForeignKey pointing to the module which this update step belonds to
```
isc-jili, line 9: nit: found another "belonds" lol

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+Method ApplyUpdatesInOrder(
+	logFile As %Library.File = "",
+	verbose As %Boolean = 0) As %Status
+{
+    if verbose {
+        write !, "Acquiring lock on ^IPM.UpdateLock("_..Name_")"
+    }
+    $$$LockUpdateModule(..Name)
+    if ('$test) {
+        quit
```
isc-kiyer, line 100: Needs to return $$$OK or put everything in the try catch

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,52 @@
+Class %IPM.Utils.Update.VersionBase [ Abstract ]
```
isc-kiyer, line 1: This should be %IPM.General.Update.VersionBase. Utils package is really just a collection of methods and no base classes as much as possible

isc-kiyer: Also add a description to the class with intended usage and what should be subclassing it

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,52 @@
...
+{
+
+/// Returns dynamic array of update method names to apply in order that can be run on both primary and backup
+/// Example output list: ["methodA", "methodB", "methodC"]
+/// When mirror enabled: These steps will run on all members
+/// When mirror not enabled: These steps will be run on the instance
+ClassMethod GetOrderedMethodsAnyMirrorMember() As %Library.DynamicArray [ Abstract ]
```
isc-kiyer, line 8: I would remove "AnyMirrorMember" and just call this GetOrderedMethods() since it implies its only relevant for mirrored instances.

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,52 @@
+Class %IPM.Utils.Update.VersionBase [ Abstract ]
+{
+
+/// Returns dynamic array of update method names to apply in order that can be run on both primary and backup
+/// Example output list: ["methodA", "methodB", "methodC"]
```
isc-kiyer, line 5: Nit: I would reword this description to the following:
```
Returns a dynamic array of update method names to apply in order. <br />
These are always run synchronously when the zpm "update" command is run (during the ApplyUpdateSteps lifecycle phase). <br />
When the instance is mirrored, this will run on every mirror member. <br />
<example>
   Output: ["methodA", "methodB", "methodC"]
</example>
```

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,52 @@
...
+/// When mirror enabled: These steps will run on all members
+/// When mirror not enabled: These steps will be run on the instance
+ClassMethod GetOrderedMethodsAnyMirrorMember() As %Library.DynamicArray [ Abstract ]
+{
+}
+
+/// Returns dynamic array of update method names to apply in order that must be run on the primary
```
isc-kiyer, line 12: Similarly, update the description to:
```
Returns a dynamic array of update method names to apply in order. This method has different behavior when the instance is mirrored versus not. <br />
When the instance is not mirrored, these are always run synchronously when the zpm "update" command is run (during the ApplyUpdateSteps lifecycle phase). <br />
When the instance is mirrored, this will run only when the instance becomes the primary member. <br />
<example>
   Output: ["methodA", "methodB", "methodC"]
</example>
```

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,52 @@
...
+/// Example output list: ["methodA", "methodB", "methodC"]
+/// When mirror enabled: These steps will run ONLY on the primary mirror member
+/// When mirror not enabled: These steps will be run on the instance
+ClassMethod GetOrderedMethodsPrimaryOnly() As %Library.DynamicArray [ Abstract ]
+{
+}
+
+/// For a given UpdatePackage name, compiles a list of the update step classes ordered by number
+/// Example output for module with update steps V1, V2, V3, & V5 classes:
+///     updatePackageClassList(1) = "MyModUpdatePackage.V1"
```
isc-kiyer, line 22: Nit: Put example in `<example>` tag

src/cls/IPM/Utils/Update/VersionBase.cls
```diff
@@ -0,0 +1,52 @@
...
+///     updatePackageClassList(2) = "MyModUpdatePackage.V2"
+///     updatePackageClassList(3) = "MyModUpdatePackage.V3"
+///     updatePackageClassList(5) = "MyModUpdatePackage.V5"
+ClassMethod GetUpdatePackageClassList(
+	updatePackage As %String,
+	Output updatePackageClassList) As %Status
+{
+    set sc = $$$OK
+    try {
+        set tRes = ##class(%SQL.Statement).%ExecDirect(,
```
isc-kiyer, line 32: Nit: do not use `t` prefix for net new code

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+    set tInitTLevel = $tlevel
+    try {
+        tstart
+
+        $$$ThrowOnError(##class(%IPM.Utils.Update.VersionBase).GetUpdatePackageClassList(..UpdatePackage, .classList))
+        set keyVersion = ""
+        for {
+            // Iterate over update package class list for steps to apply. Apply all of those up to and including this major version
+            set keyVersion = $order(classList(keyVersion))
+            if (keyVersion = "") || (keyVersion > ..Version.Major) {
```
isc-kiyer, line 113: You don't need the major version check since it should not be possible to have a class for a major version higher than the current version installed. So you can simply go through all classes found. Anything that was seeded, will get skipped

src/cls/IPM/Storage/Module.cls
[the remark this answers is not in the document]
isc-kiyer: I agree with @isc-jlechtne. For each major version, find the first step that has not been run and run from there to the end. This should be made clearer in the method description. Right now it sounds like if say we are on v3, if there is a step found on v1 that has not been run, everything gets re-run from there (but should indicate it is scoped by major version)

src/cls/IPM/Storage/Module.cls
isc-kiyer: We should however (as a separate issue, out of scope for here) add a compile time check in VersionBase so any subclass of it can have at most the major version of the module. So for example, if major version is 3, there cannot be a V4 class in the update package.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+        set className = ..UpdatePackage_".V"_version
+        if '##class(%Dictionary.ClassDefinition).%ExistsId(className) {
+            if verbose {
+                write !, "Class """_className_""" does not exist. Skipping running its update steps"
+            }
+            quit
+        }
+
+        // Regardless of mirror status, run these methods first
+        set anyMirrorMemberMethodNames = $classmethod(className, "GetOrderedMethodsAnyMirrorMember")
```
isc-kiyer, line 159: Have if $System.CLS.IsMthd(className, <method name>) for each of these to ensure we only run it if its been overridden.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+///                  Can carry this over to subsequent calls by passing the newStepToApply variable to those ExecuteUpdateStepMethodList() calls
+Method ExecuteUpdateStepMethodList(
+	methodList As %Library.DynamicArray,
+	className As %String,
+	primaryOnly As %Boolean,
+	logfile As %Library.File,
+	verbose As %Boolean = 0,
+	ByRef newStepToApply As %Boolean = 0) As %Status [ Private ]
+{
+    set sc = $$$OK
```
isc-kiyer, line 186: No need to return a status for this method. Can let it throw errors

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+        if '##class(%Dictionary.ClassDefinition).%ExistsId(className) {
+            if verbose {
+                write !, "Class """_className_""" does not exist. Skipping running its update steps"
+            }
+            quit
+        }
+
+        // Regardless of mirror status, run these methods first
+        set anyMirrorMemberMethodNames = $classmethod(className, "GetOrderedMethodsAnyMirrorMember")
+        set sc = ..ExecuteUpdateStepMethodList(anyMirrorMemberMethodNames, className, 0, logFile, verbose, .newStepToApply)
```
isc-kiyer, line 160: Make these `do` once the methods re updated to not return statuses

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+            if newStepToApply {
+                if verbose {
+                    write !, "Running "_className_"::"_methodName_"()"
+                }
+                if primaryOnly {
+                    set stepObj.HostName = ##class(%SYS.System).GetNodeName()
+                    set stepObj.InstanceName = ##class(%SYS.System).GetInstanceName()
+                }
+                set stepObj.TimeStampStart = $zdatetime($horolog,3)
+                set stepObj.Status = $$$OK
```
isc-kiyer, line 206: Nit: Should not need to set Status and TimeStampStart. These should have initial expressions.
Also, for TimeStampStart, its initial expression can be `##class(%UTC).NowUTC()`

isc-kiyer (changes requested):
@isc-jlechtne my comments are up!

src/cls/IPM/Utils/Update/VersionBase.cls
isc-kiyer: On further thought, I think its overkill and just need the latest error status. We would only ever have a single update step with an error status (per table - primary only vs any member) since we stop processing once an error is hit

src/cls/IPM/Storage/Module.cls
isc-jlechtne: Yeah I wasn't sure about that being a realistic use case but that additional check makes sense

src/cls/IPM/Storage/UpdateStep/AnyMirrorMember.cls
```diff
@@ -0,0 +1,58 @@
...
+<Value>TimeStampStart</Value>
+</Value>
+<Value name="5">
+<Value>TimeStampEnd</Value>
+</Value>
+<Value name="6">
+<Value>Status</Value>
+</Value>
+<Value name="7">
+<Value>ModuleHome</Value>
```
isc-jili, line 44: This seems to be leftover from a previous design; should no longer be here

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,262 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+    } catch (ex) {
+        set sc = ex.AsStatus()
+    }
+    quit sc
+}
+
+/// Iterates over a provided list of method names + class name and seeds their update steps by setting their TimeStampEnd values to imply they have been run
+/// Steps that already have TimeStampEnd defined (already been run/seeded) will be skipped
+/// Steps that don't exist yet will be created and immediately seeded
+Method SeedUpdateStepsFromList(
```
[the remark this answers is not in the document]
isc-jlechtne: While they do iterate over the update steps in a similar manner, I would lean against combining them into one method for how differently they handle the steps (particularly with the newStepToApply flag which is only in executing the steps but not used in seeding). Think combining them would add confusion to the readability of both cases more than condensing any code written. Curious to hear other people's thoughts, @isc-jili & @isc-eneil

src/cls/IPM/Storage/Module.cls
isc-eneil: I agree that `TimeStampStart` should be empty so we can tell if the method was seeded or not.

isc-eneil: But maybe the value of TimeStampStart can be an arg to %OnNew()?

src/cls/IPM/Storage/Module.cls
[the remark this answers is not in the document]
isc-eneil: I think I agree with @isc-jlechtne -- there's more context to hold onto when executing steps. I'm fine with the common units of logic of creating and getting update steps being pulled out (`GetUpdateStep()` and `AddUpdateStep()`) but the methods that use them staying separate

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,217 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+            write !, $char(9)_"Running "_updateStep.MethodName_"()"
+        }
+        if primaryOnly {
+            set updateStep.HostName = ##class(%SYS.System).GetNodeName()
+            set updateStep.InstanceName = ##class(%SYS.System).GetInstanceName()
+        }
+        set updateStep.TimeStampStart = $zdatetime($horolog,3)
+        set updateStep.Status = $$$OK
+        try {
+
```
isc-jili, line 246: nit: perhaps an extra blank line here

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,217 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+        if verbose {
+            write !, "Acquiring lock on ^IPM.UpdateLock("_..Name_")"
+        }
+        $$$LockUpdateModule(..Name)
+        if ('$test) {
+            return $$$OK
+        }
+    }
+
+    set sc = $$$OK
```
isc-jili, line 108: Is there any reason that sc is not set at the beginning of this method such that if '$test, can return sc instead of $$$OK?

src/cls/IPM/General/Update/VersionBase.cls
```diff
@@ -0,0 +1,61 @@
...
+            "FROM %Dictionary.CompiledClass "_
+            "WHERE Name %STARTSWITH '"_updatePackage_".V'")
+        if (res.%SQLCODE < 0) {
+            throw ##class(%Exception.SQL).CreateFromSQLCODE(res.%SQLCODE,res.%Message)
+        }
+        while res.%Next() {
+            // Class name is of the form UpdatePackage.V#
+            set className = res.Name
+            // This strips everything except numbers following the '.' in the class name so we can isolate the # of this update class itself
+            set classNum = $zstrip($piece(className, ".", *),"*E'N")
```
isc-jili, line 52: This may belong in compile time checks, but perhaps at least add validation that classNum is not empty before setting updatePackageClassList(classNum)?

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,217 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+            if verbose {
+                if seedSteps {
+                    write !, "Starting seeding steps for module = "_..Name_" & class = "_className
+                }
+                else {
+                    write !, "Starting to run update steps for module = "_..Name_" & class = "_className
+                }
+            }
+
+            // Don't want newStepToApply (execution of all subsequent steps when seeing a new one) to persist across versions
```
isc-jili, line 128: Perhaps reword to not use "seeing" because is too similar to "seeding"

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,217 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+	verbose As %Boolean = 0) As %Status
+{
+    // Only need to acquire lock when applying the steps, not when seeding them
+    if 'seedSteps {
+        if verbose {
+            write !, "Acquiring lock on ^IPM.UpdateLock("_..Name_")"
+        }
+        $$$LockUpdateModule(..Name)
+        if ('$test) {
+            return $$$OK
```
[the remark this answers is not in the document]
isc-jlechtne: I believe I did have that but then removed it in response to a previous comment. Additionally, I was not getting caught up in verbose details on this anymore since it'll all probably change when we implement our in depth logging updates.

src/cls/IPM/Storage/Module.cls
```diff
@@ -87,140 +87,173 @@ Property Deployed As %Boolean(XMLPROJECTION = "Element");
...
+        if primaryOnly {
+            set updateStep.HostName = ##class(%SYS.System).GetNodeName()
+            set updateStep.InstanceName = ##class(%SYS.System).GetInstanceName()
+        }
+        set updateStep.TimeStampEnd = $zdatetime($horolog,3)
+        $$$ThrowOnError(updateStep.%Save())
+    }
+    else {
+        if verbose {
+            write !, $char(9)_"Skipping seeding "_updateStep.MethodName_"() (already seeded)"
```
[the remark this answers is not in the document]
isc-jlechtne: Meant "seeded" as a more general term for steps with TimeStampEnd values. Will update to be more explicit on "step already ran or seeded" but, as I mentioned to Jacqueline in a different comment, I don't want to get held up too much on verbose wording here because no matter what we end up on it will change when we get to the standalone logging story.

src/cls/IPM/Storage/Module.cls
```diff
@@ -125,7 +125,7 @@ Method HandleAllUpdateSteps(
                 }
             }
 
-            // Don't want newStepToApply (execution of all subsequent steps when seeing a new one) to persist across versions
+            // Don't want newStepToApply (execution of all subsequent steps when encountering a new one) to persist across versions
```
isc-eneil, line 128: Ha I like @isc-jlechtne

isc-eneil (approved):

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,218 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+        set updateStep.HostName = ##class(%SYS.System).GetNodeName()
+        set updateStep.InstanceName = ##class(%SYS.System).GetInstanceName()
+    }
+    set updateStep.TimeStampStart = $zdatetime($horolog,3)
+    set updateStep.Status = $$$OK
+    try {
+        $$$ThrowOnError($classmethod(updateStep.ClassName, updateStep.MethodName))
+        set updateStep.TimeStampEnd = $zdatetime($horolog,3)
+    } catch (ex) {
+        set updateStep.Status = ex.AsStatus()
```
isc-kiyer, line 251: If a single method errors, should throw the error to stop further processing

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,218 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+    if createNew && '$isobject(stepObj) {
+        set stepObj = ..AddUpdateStep(className, methodName, primaryOnly)
+    }
+    return stepObj
+}
+
+/// Create a new update step object for this module
+/// Update steps are defined by their unique className + methodName combination
+/// primaryOnly - set to 1 if this new update step can only be run on the primary in a mirrored environment
+Method AddUpdateStep(
```
isc-kiyer, line 283: This is just adding. Not updating.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,218 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+	className As %String,
+	methodName As %String,
+	primaryOnly As %Boolean,
+	createNew As %Boolean = 0) As %IPM.Storage.UpdateStep.Base
+{
+    if primaryOnly {
+        set stepObj = ##class(%IPM.Storage.UpdateStep.PrimaryOnly).GetUpdateStep(className, methodName)
+    }
+    else {
+        set stepObj = ##class(%IPM.Storage.UpdateStep.AnyMember).GetUpdateStep(className, methodName)
```
isc-kiyer, line 272: Why not have GetUpdateStep just return a new record if only doesn't exist instead of returning ""? Then below you can just call %Save()

isc-kiyer: If you need to know whether a step exists or not, can have a separate HasUpdateStep() method

isc-kiyer (changes requested):
@isc-jlechtne few more comments

src/cls/IPM/Storage/Module.cls
isc-eneil: @isc-kiyer That method was `GetAddUpdateStep()` a few reviews back but I commented that I thought it was more readable to have a `GetUpdateStep()` and `AddUpdateStep()`. I'll defer to @isc-jlechtne  on adding something like a `HasUpdateStep()` but don't feel strongly at this point
cc @isc-jlechtne

isc-jlechtne: Building off of what @isc-eneil mentioned, I split up Get and Add but added in the 'createNew' flag in the Get method to allow it to create a new step if one doesn't already exist.

A separate HasUpdateStep() doesn't seem necessary to me right now mainly because I don't see a case where I would use it outside of GetUpdateStep() but '$isobject(stepObj) already satisfies that.

Returning "" also makes sense to me since GetUpdateStep() is essentially a wrapper for %OpenId() of the step classes and it just returns what those calls return.

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,216 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+}
+
+/// Attempt to get update step object with this className + methodName index
+/// primaryOnly - set to 1 if this update step is one that runs on the primaryOnly
+/// createNew - set to 1 to create a new step object with these values if one is not found to already exist
+Method GetUpdateStep(
+	className As %String,
+	methodName As %String,
+	primaryOnly As %Boolean,
+	createNew As %Boolean = 0) As %IPM.Storage.UpdateStep.Base
```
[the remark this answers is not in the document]
isc-eneil: Thread on this is in another comment @isc-kiyer

src/cls/IPM/Storage/Module.cls
isc-eneil: @isc-kiyer here's the thread re:
"I don't see any case where you want createNew to be false right? So may as well drop the parameter and always create new. The GetUpdateStep method in PrimaryOnly and AnyMember could be updated to return %New() if no record was found rather than returning "" (which is what %OpenId() returns when no record is found)"

isc-eneil: @isc-jlechtne In fact it seems this always create the step anyways because it calls `%OpenId()` under the hood. So I guess the decision is made for us and we can remove `AddUpdateStep()` altogether? That condition will never be reached.

isc-eneil (changes requested):

src/cls/IPM/General/Update/VersionBase.cls
```diff
@@ -0,0 +1,61 @@
+/// For a module that requires steps to be run when updating to a new version, define those steps in a class named <UpdatePackage>.V#.
+/// To make sure those steps run, have that class extend %IPM.General.Update.VersionBase and
```
isc-jili, line 2: nit: "To make sure those steps run" seems a bit redundant/already implied. Maybe rephrase as "That class must extend %IPM.General.Update.VersionBase and override at least ..."

src/cls/IPM/Storage/Module.cls
```diff
@@ -85,6 +85,216 @@ Property AfterInstallMessage As %String(MAXLEN = "", XMLPROJECTION = "Element");
...
+	updateStep As %IPM.Storage.UpdateStep.Base,
+	primaryOnly As %Boolean,
+	verbose As %Boolean = 0)
+{
+    if 'updateStep.TimeStampEnd {
+        if verbose {
+            write !, $char(9)_"Seeding "_updateStep.MethodName_"()"
+        }
+        if primaryOnly {
+            set updateStep.HostName = ##class(%SYS.System).GetNodeName()
```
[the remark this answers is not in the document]
isc-jlechtne: Yep! Removed these from ExecuteUpdateStep in the prior review, thanks for finding the same issue in SeedUpdateStep

src/cls/IPM/Storage/Module.cls
isc-jlechtne: @isc-eneil It can still be reached under the current implementation. If the parameters for GetUpdateStep() is not an already existing update step then %OpenId() will return "", which will be true for the '$isobject(stepObj) condition.

isc-jili (approved):

isc-eneil (approved):

isc-jili (approved):

PR #856, opened 2025-08-03
Intermediate PR for IPM commands

isc-jili:
This is a intermediate PR for IPM commands. With the addition of the new update command, testing can flow better amongst everyone working on the update framework.

Remaining TODOs:
1. "install" and "load": Add "Force" flag
2. "uninstall": Complete logic for "delete-data" flag
3. "update": Throw a warning to alert users of which update steps have been seeded (if any).
4. "update": Add more verbose logging
5. "update": Flags for log output device + Log file name + Log file location + Local module tarball location + JSON configuration file location + Force
6. "update": Ensure "update" command is not run on the primary instance if it hasn't run on the backup
7. Consider how to hold onto to tune table metadata (automatic exports/imports on update?)

### DIFF
--- a/src/cls/IPM/Lifecycle/Module.cls
+++ b/src/cls/IPM/Lifecycle/Module.cls
@@ -186,4 +186,21 @@ Method %Package(ByRef pParams) As %Status
   quit tSC
 }
 
+Method %Update(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
+    try {
+        set tVerbose = $get(pParams("Verbose"))
+        set tVersion = $get(pParams("Version"))
+        set tModule = ##class(%IPM.Storage.Module).NameOpen(..Module.DisplayName)
+        write:tVerbose !,"Seeding update steps",!
+        $$$ThrowOnError(tModule.SeedUpdateSteps())
+        write:tVerbose !,"Applying update steps",!
+        do ##class(%IPM.Utils.Update.Controller).ApplyUpdatesInOrder(tModule,,tVerbose)
+    } catch e {
+        set tSC = e.AsStatus()
+    }
+  quit tSC
+}
+
 }
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -7,7 +7,7 @@ Class %IPM.Main Extends %IPM.CLI
 
 Parameter DOMAIN = "ZPM";
 
-Parameter STANDARDPHASES = {$listbuild("reload","compile","test","package","verify","publish","makedeployed")};
+Parameter STANDARDPHASES = {$listbuild("reload","compile","test","package","verify","publish","makedeployed","update")};
 
 /// Description of commands to use for this CLI
 XData Commands [ XMLNamespace = "http://www.intersystems.com/PackageManager/CLI" ]
@@ -55,6 +55,8 @@ This is accessible to other instances configured to look at the current namespac
 module repository.
 * publish: saves that artifact to the repository for which deployment is enabled.
 Currently, there may only be one of these per namespace.
+* update: run any update steps for the module to become updated to the latest available version by default.
+Can also specify desired version to upgrade to.
 </description>
 
 <!-- Examples -->
@@ -137,6 +139,26 @@ This command is an alias for `module-action module-name publish`
 <parameter name="module" required="true" description="Name of module on which to perform publish actions" />
 <modifier name="repo" aliases="r" dataAlias="PublishTo" description="Namespace-unique name for the module to publish to (if deployment is enabled)" />
 <modifier name="use-external-name" aliases="use-ext" dataAlias="UseExternalName" dataValue="1" description="Publish the package under the &lt;ExternalName&gt; of the package. If ExternalName is unspecified or illegal, an error will be thrown."/>
+</command>
+
+<command name="update" dataPrefix="D">
+<summary>Updates a module to a newer version.</summary>
+<description>
+	Updates a module to the latest available version by default. Can also specify a version with the -version modifier. Runs all of the update steps from the current version to the newer version.
+</description>
+
+<!-- Examples -->
+<example description="Updates the HS.JSON module from the current version to version 3.0.0.">
+	update HS.JSON -version 3.0.0
+</example>
+
+<!-- Parameters -->
+<parameter name="module" required="true" description="Name of module on which to perform update actions" />
+
+<!-- Modifiers -->
+<modifier name="version" aliases="ver" dataAlias="Version" description="Version to update module to" />
+<modifier name="verbose" aliases="v" dataAlias="Verbose" dataValue="1" description="Produces verbose output from the command." />
+<modifier name="force" aliases="f" dataAlias="Force" dataValue="1" description="If specified, the module will be updated even if there are warnings." />
 </command>
 
 <command name="makedeployed">
@@ -447,6 +469,7 @@ reinstall -env /path/to/env1.json;/path/to/env2.json example-package
 <modifier name="quiet" aliases="q" dataAlias="Verbose" dataValue="0" description="Produces minimal output from the command." />
 <modifier name="verbose" aliases="v" dataAlias="Verbose" dataValue="1" description="Produces verbose output from the command." />
 <modifier name="purge" dataAlias="Purge" dataValue="1" description="Purge data from tables during uninstall." />
+<modifier name="delete-data" dataAlias="DeleteData" dataValue="1" description="Deletes persistent data for the module." />
 
 </command>
 
@@ -2242,6 +2265,12 @@ ClassMethod Install(ByRef pCommandInfo) [ Internal ]
                 }
             }
             $$$ThrowOnError(##class(%IPM.Utils.Module).LoadQualifiedReference(tResult, .tParams, , tSynchronous))
+            if '$get(tParams("DeveloperMode")) {
+                set tModule = ##class(%IPM.Storage.Module).NameOpen(tModuleName)
+                if $isobject(tModule) {
+                    $$$ThrowOnError(tModule.SeedUpdateSteps())
+                }
+            }
         }
     } else {
         set tPrefix = ""
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -1343,6 +1343,10 @@ ClassMethod LoadNewModule(
             tcommit
         }
         $$$ThrowOnError(tSC)
+        // On each load of a new module, we want to seed its update steps
+        if 'tDeveloperMode {
+            $$$ThrowOnError(tModule.SeedUpdateSteps())
+        }
         do tModule.WriteAfterInstallMessage()
     } catch e {
         set tSC = e.AsStatus()


src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +186,21 @@ Method %Package(ByRef pParams) As %Status
   quit tSC
 }
 
+Method %Update(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
+    try {
+        set tVerbose = $get(pParams("Verbose"))
+        set tVersion = $get(pParams("Version"))
```
isc-jili, line 194: tVersion isn't used right now because ApplyUpdatesInOrder() doesn't currently take a parameter for toVersion. I commented on Jimmy's PR (https://github.com/intersystems/ipm/pull/855/files) to request that be an additional parameter.

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +186,48 @@ Method %Package(ByRef pParams) As %Status
   quit tSC
 }
 
+Method %Update(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
+    try {
+        set tVerbose = $get(pParams("Verbose"))
```
isc-jlechtne, line 193: Don't want to use the `t` prefix for new code

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +186,48 @@ Method %Package(ByRef pParams) As %Status
...
+        do ##class(%IPM.Lifecycle.Base).GetDefaultParameters(.tParams)
+        merge tParams = pParams("data")
+        set tParams("DeveloperMode") = $get(tParams("DeveloperMode"), 0)
+        set tParams("cmd") = "install"
+        set tParams("Install") = 1
+
+        write:tVerbose !,"Load new module into storage",!
+        $$$ThrowOnError(##class(%IPM.Utils.Module).LoadQualifiedReference(tResult, .tParams))
+        set tNewModule = ##class(%IPM.Storage.Module).NameOpen(tModuleName)
+        zwrite tNewModule
```
isc-jlechtne, line 219: Remove this zwrite

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +186,48 @@ Method %Package(ByRef pParams) As %Status
...
+{
+    set tSC = $$$OK
+    try {
+        set tVerbose = $get(pParams("Verbose"))
+        set tVersion = $get(pParams("Version"))
+        set tModuleName = ..Module.DisplayName
+        set tOldModule = ##class(%IPM.Storage.Module).NameOpen(tModuleName)
+
+        write:tVerbose !,"Seeding update steps for original module version",!
+        $$$ThrowOnError(tOldModule.SeedUpdateSteps())
```
isc-kiyer, line 199: Why are we ever seeding steps in Update? Seeding of steps would be done in Activate when a module is branch new

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +186,48 @@ Method %Package(ByRef pParams) As %Status
...
+
+        write:tVerbose !,"Resolving new module version to install",!
+        set tSearchCriteria = ##class(%IPM.Repo.SearchCriteria).%New()
+        set tSearchCriteria.Name = $$$lcase(tModuleName)
+        set tSearchCriteria.VersionExpression = tVersion
+        $$$ThrowOnError(##class(%IPM.Repo.Utils).SearchRepositoriesForModule(tSearchCriteria,.tResults))
```
isc-kiyer, line 205: Sooo I think there are some misconceptions about how update should work. When updating a module, is is really just the same as install with the one additional phase just for running update steps. So at this point in the code, the new module code is already installed and we are running update steps for the module and its dependencies.

src/cls/IPM/Main.cls
```diff
@@ -139,6 +141,27 @@ This command is an alias for `module-action module-name publish`
...
 
+<command name="update" dataPrefix="D">
+<summary>Updates a module to a newer version.</summary>
+<description>
+	Updates a module to the latest available version by default. Can also specify a version with the -version modifier. Runs all of the update steps from the current version to the newer version.
+</description>
+
+<!-- Examples -->
+<example description="Updates the HS.JSON module from the current version to version 3.0.0.">
+	update HS.JSON -version 3.0.0
```
isc-kiyer, line 152: Nit: following how other commands include module name and version, both should be parameters (version should not be a modifier)

src/cls/IPM/Main.cls
```diff
@@ -447,6 +470,7 @@ reinstall -env /path/to/env1.json;/path/to/env2.json example-package
 <modifier name="quiet" aliases="q" dataAlias="Verbose" dataValue="0" description="Produces minimal output from the command." />
 <modifier name="verbose" aliases="v" dataAlias="Verbose" dataValue="1" description="Produces verbose output from the command." />
 <modifier name="purge" dataAlias="Purge" dataValue="1" description="Purge data from tables during uninstall." />
+<modifier name="delete-data" dataAlias="DeleteData" dataValue="1" description="Deletes persistent data for the module." />
```
isc-kiyer, line 473: I would say exclude deleta-data for now since that is an enhancment that doesn't technically have anything to do with the update framework. Worthwhile creating a separate github issue for it

src/cls/IPM/Main.cls
```diff
@@ -2242,6 +2266,12 @@ ClassMethod Install(ByRef pCommandInfo) [ Internal ]
                 }
             }
             $$$ThrowOnError(##class(%IPM.Utils.Module).LoadQualifiedReference(tResult, .tParams, , tSynchronous))
+            if '$get(tParams("DeveloperMode")) {
+                set tModule = ##class(%IPM.Storage.Module).NameOpen(tModuleName)
```
isc-kiyer, line 2270: Seeding of update steps should be done in the activate phase per the design doc correct?

src/cls/IPM/Utils/Module.cls
```diff
@@ -1343,6 +1343,10 @@ ClassMethod LoadNewModule(
             tcommit
         }
         $$$ThrowOnError(tSC)
+        // On each load of a new module, we want to seed its update steps
+        if 'tDeveloperMode {
```
isc-kiyer, line 1347: Again, not here but rather in the activate phase in %IPM.Lifecycle.Base. Also, in GetCompletePhasesForOne(), update should run all phases that activate would, in addition to update at the end.

isc-kiyer (changes requested):
@isc-jili my comments are up!

src/cls/IPM/Lifecycle/Module.cls
isc-jili: Thanks for that clarification. Yes that was my understanding for an earlier commit but I ended up overthinking it. I will revert

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -56,6 +56,12 @@ Method %Activate(ByRef pParams) As %Status
                 }
             }
         }
+        if '..Module.DeveloperMode && $isobject(..Module) {
+            if tVerbose {
+                write !, "Seeding update steps for module ",..Module.Name, ", for version ", ..Module.VersionString, !
+            }
+            #; $$$ThrowOnError(..Module.SeedUpdateSteps())
```
isc-jili, line 63: Oops, should uncomment this

src/cls/IPM/Utils/Module.cls
```diff
@@ -1338,7 +1338,11 @@ ClassMethod LoadNewModule(
         if tVerbose || $get(pParams("Multicompile","Verbose"),0) {
             write !,"Loading "_tDisplayName_" in process "_$job
         }
-        set tSC = ##class(%IPM.Storage.Module).ExecutePhases(tModuleName,$listbuild("Activate"),1,.pParams)
+        if $get(pParams("cmd")) = "update" {
```
isc-eneil, line 1341: I know this is what we decided on when we spoke offline because this will allow us to avoid running BuildDependencyGraph and looping through all the modules twice, but I really don't like how hidden this logic is... not sure what the solution is architecturally or if we just need really good comments, but if I were looking at the Update() method in Main.cls, I think I would be confused.

Thoughts @isc-kiyer  @isc-jlechtne @isc-jili ?

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +192,19 @@ Method %Package(ByRef pParams) As %Status
   quit tSC
 }
 
+Method %ApplyUpdateSteps(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
+    try {
+        set verbose = $get(pParams("Verbose"))
+        if verbose {
+            write !,"Applying update steps for new version ", ..Module.VersionString, " of module ", ..Module.Name,!
```
isc-jlechtne, line 201: We will have to coordinate our verbosity because I have pretty much this exact write statement inside of ApplyUpdateInOrder().

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -186,4 +192,19 @@ Method %Package(ByRef pParams) As %Status
...
+        }
+        do ##class(%IPM.Utils.Update.Controller).ApplyUpdatesInOrder(..Module, ,verbose)
```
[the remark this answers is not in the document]
isc-jlechtne: ApplyUpdatesInOrder was one I just moved to %IPM.Storage.Module so will be ..Module.ApplyUpdatesInOrder(<logFile>, verbose)

src/cls/IPM/Utils/Module.cls
isc-jili: I can put comments and method descriptions to hopefully make this a little clearer.
I think it would be difficult to make the logic explicit since Update() calls LoadQualifiedReference() which calls LoadModuleReference() which calls LoadModuleFromDirectory() which calls LoadNewModule() which calls ExecutePhases()...
That amount of nested calls makes it very hard to make it explicit that we are passing params for a potential update as we would have to change multiple method names and/or signatures.
But also open to thoughts from others!

src/cls/IPM/Utils/Module.cls
```diff
@@ -304,10 +304,18 @@ ClassMethod LoadModuleReference(
                     set tSC = $$$ERROR($$$GeneralError, $$$FormatText("Cannot install '%1' over previously installed in developer mode", tInstModule.Name))
                     quit
                 }
+                // TODO: Move the following 2 errors down to LoadNewModule(). But need to find a way to pass in the toVersion in the 'load' case.
```
isc-jili, line 307: Need to finish this TODO in order to finish -force flag for "load"

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -191,22 +192,23 @@ ClassMethod MatchSinglePhase(pOnePhase As %String) As %String
...
+        "makedeployed":     "MakeDeployed",
+        "test":             "Test",
+        "package":          "Package",
+        "verify":           "Verify",
+        "register":         "Register",
+        "publish":          "Publish",
+        "configure":        "Configure",
+        "unconfigure":      "Unconfigure",
+        "applyupdatesteps": "ApplyUpdateSteps",
         :               pOnePhase // return the phase as-is if it's a custom phase name
```
isc-jili, line 212: nit: pad this one line with spaces

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1687,6 +1689,18 @@ Method %MakeDeployed(ByRef pParams) As %Status
...
+Method %ApplyUpdateSteps(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
+    try {
+        set verbose = $get(pParams("Verbose"))
+        do ..Module.ApplyUpdatesInOrder(,verbose)
+    } catch e {
+        set tSC = e.AsStatus()
+    }
+  quit tSC
```
isc-jlechtne, line 1701: Nit: Looks like the spacing on this quit statement is a bit off

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1687,6 +1689,18 @@ Method %MakeDeployed(ByRef pParams) As %Status
     quit tSC
 }
 
+Method %ApplyUpdateSteps(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
+    try {
+        set verbose = $get(pParams("Verbose"))
+        do ..Module.ApplyUpdatesInOrder(,verbose)
```
isc-jlechtne, line 1697: The first parameter here will be the log file. I'm okay with checking this in as is and adding the log file after our meeting on how we want to go about logging but let's make sure not to forget to come back here after.

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -56,6 +56,12 @@ Method %Activate(ByRef pParams) As %Status
                 }
             }
         }
+        if '$get(pParams("Update")) && '..Module.DeveloperMode && $isobject(..Module) {
+            if tVerbose {
+                write !, "Seeding update steps for module ",..Module.Name, ", for version ", ..Module.VersionString, !
+            }
+            $$$ThrowOnError(..Module.SeedUpdateSteps())
```
isc-jlechtne, line 63: SeedUpdateSteps() has a verbose parameter

src/cls/IPM/Utils/Module.cls
```diff
@@ -1338,7 +1347,10 @@ ClassMethod LoadNewModule(
         if tVerbose || $get(pParams("Multicompile","Verbose"),0) {
             write !,"Loading "_tDisplayName_" in process "_$job
         }
-        set tSC = ##class(%IPM.Storage.Module).ExecutePhases(tModuleName,$listbuild("Activate"),1,.pParams)
+        // LoadNewModule() may be called on an install/load or an update. In the case of an update, invoke ApplyUpdateSteps lifecycle phase.
+        // Otherwise, invoke Activate lifecycle phase.
+        set lifeCyclePhase = $select($get(pParams("Update")): "ApplyUpdateSteps", 1: "Activate")
```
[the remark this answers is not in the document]
isc-eneil: @isc-kiyer Isn't that what it does?

The way I read it, if pParams("Update") is set, lifeCyclePhase = "ApplyUpdateSteps" and in any other case, it sets it to "Activate"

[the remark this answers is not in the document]
isc-jili: @isc-kiyer I also think that it is doing the correct behavior; if $get(pParams("Update")) = 1, then it evaluates to "ApplyUpdateSteps", and "Activate" if not

src/cls/IPM/Main.cls
```diff
@@ -139,6 +141,28 @@ This command is an alias for `module-action module-name publish`
 <modifier name="use-external-name" aliases="use-ext" dataAlias="UseExternalName" dataValue="1" description="Publish the package under the &lt;ExternalName&gt; of the package. If ExternalName is unspecified or illegal, an error will be thrown."/>
 </command>
 
+<command name="update" dataPrefix="D">
+<summary>Updates a module to a newer version.</summary>
+<description>
+	Updates a module to a newer version. Also runs all of the update steps from the current version to the newer version.
```
isc-jili, line 147: nit: add "latest available version by default"

isc-eneil (changes requested):
Left a round of comments @isc-jili !

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -56,6 +56,12 @@ Method %Activate(ByRef pParams) As %Status
                 }
             }
         }
+        if '$get(pParams("Update")) && '..Module.DeveloperMode {
+            if tVerbose {
+                write !, "Seeding update steps for module ",..Module.Name, ", for version ", ..Module.VersionString, !
+            }
+            $$$ThrowOnError(..Module.SeedUpdateSteps(tVerbose))
```
isc-jili, line 63: is now "SeedAllUpdateSteps"

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1687,6 +1689,23 @@ Method %MakeDeployed(ByRef pParams) As %Status
     quit tSC
 }
 
+Method %ApplyUpdateSteps(ByRef pParams) As %Status
+{
+    set tSC = $$$OK
```
isc-kiyer, line 1694: Nit: do not use `t` prefix for net new code

src/cls/IPM/Main.cls
```diff
@@ -55,6 +55,8 @@ This is accessible to other instances configured to look at the current namespac
 module repository.
 * publish: saves that artifact to the repository for which deployment is enabled.
 Currently, there may only be one of these per namespace.
+* applyupdatesteps: run any update steps for the module to become updated to the latest available version by default.
```
isc-kiyer, line 58: Nit: "by default" is not needed

src/cls/IPM/Main.cls
```diff
@@ -139,6 +141,28 @@ This command is an alias for `module-action module-name publish`
...
+</description>
+
+<!-- Examples -->
+<example description="Updates the HS.JSON module from the current version to version 3.0.0.">
+	update HS.JSON 3.0.0
+</example>
+
+<!-- Parameters -->
+<parameter name="module" required="true" description="Name of module on which to perform update actions" />
+<parameter name="version" description="Version (or version expression) of module to update; defaults to the latest available if unspecified." />
```
isc-kiyer, line 157: Should also specify that this is ignored if `-path` is provided since it will then update to that version

src/cls/IPM/Main.cls
```diff
@@ -3789,4 +3825,36 @@ ClassMethod Information(ByRef pCommandInfo)
     do ..DisplayModules(.updatedList)
 }
 
+ClassMethod Update(ByRef pCommandInfo)
+{
+    set verbose = $get(pCommandInfo("data","Verbose"))
+    set version = $get(pCommandInfo("parameters","version"))
+    set path = $$$GetModifier(pCommandInfo,"path")
+    set moduleName = $get(pCommandInfo("parameters","module"))
+    set moduleCurrent = ##class(%IPM.Storage.Module).NameOpen(moduleName,.tSC)
```
isc-kiyer, line 3834: Nit: Use NameExists instead. Also, tSC would be an error here if the module did not exist and so would throw an error before anything further runs

src/cls/IPM/Main.cls
```diff
@@ -3789,4 +3825,36 @@ ClassMethod Information(ByRef pCommandInfo)
...
+    $$$ThrowOnError(tSC)
+    if moduleCurrent = "" {
+        write !,"ERROR: Module "_moduleName_" has not been installed. Please use the install or load commands to install the module before calling update.",!
+        quit
+    }
+    if verbose {
+        if version = "" {
+            write !,"Resolving latest available version of module ",moduleName," to install",!
```
isc-kiyer, line 3842: Should check if path is non-empty and indicate it will just take whatever version is in that then

src/cls/IPM/Main.cls
```diff
@@ -2232,8 +2263,13 @@ ClassMethod Install(ByRef pCommandInfo) [ Internal ]
             do ##class(%IPM.Lifecycle.Base).GetDefaultParameters(.tParams)
             merge tParams = pCommandInfo("data")
             set tParams("DeveloperMode") = $get(tParams("DeveloperMode"), 0)
-            set tParams("cmd") = "install"
-            set tParams("Install") = 1
+            set tParams("Force") = tForce
+            // In the case of an update, pCommandInfo has already been pre-populated with the necessary information.
+            // Therefore, tParams needs to be populated if this is an install command.
+            if '$get(tParams("Update"),0) {
+                set tParams("cmd") = "install"
```
isc-kiyer, line 2270: Need similar logic in load to not overwrite "cmd" subscript to "load" instead of "update"

src/cls/IPM/Utils/Module.cls
```diff
@@ -1242,6 +1231,17 @@ ClassMethod LoadNewModule(
 
         set pDirectory = ##class(%File).NormalizeDirectory("",pDirectory)
 
+        set moduleNotPersistent = ..GetModuleObjectFromPath(pDirectory,.found)
```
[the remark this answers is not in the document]
isc-kiyer: Nit: I would just call this variable - newModuleObj

src/cls/IPM/Utils/Module.cls
```diff
@@ -1272,6 +1281,27 @@ ClassMethod LoadNewModule(
 
         set tModule = ##class(%IPM.Storage.Module).NameOpen(tModuleName,,.tSC)
         $$$ThrowOnError(tSC)
+        set toVersion = tModule.Version
+        set toVersionString = tModule.VersionString
+
+        // This means that this is not a new install/load of a module
+        if fromVersion '= "" {
+            // If the version of the module about to be installed is OLDER than the module that is already installed
+            if fromVersion.Follows(toVersion) {
```
isc-kiyer, line 1290: These checks can be moved further up right after GetModuleObjectFromPath to diff the versions and throw the error if needed.

src/cls/IPM/Utils/Module.cls
```diff
@@ -1272,6 +1281,27 @@ ClassMethod LoadNewModule(
...
+                set tSC = $$$ERROR($$$GeneralError,$$$FormatText("Cannot downgrade %1 from version %2 to %3",tModuleName,fromVersionString,toVersionString))
```
isc-kiyer, line 1291: Nit: specify "when running update command" at the end.

isc-kiyer (changes requested):
@isc-jili my comments are up!

src/cls/IPM/Utils/Module.cls
isc-jili: @isc-kiyer I'm actually not sure because I need to get the toVersion from tModule (the one from %IPM.Storage.Module.NameOpen) which doesn't exist at that point.

isc-kiyer: @isc-jili You can get fromVersion using NameOpen on the existing persistence and toVersion from the call to GetModuleObjectFromPath(). Am I missing something?

isc-jili: Ah yes you're correct my bad!

src/cls/IPM/Utils/Module.cls
isc-jili: @isc-kiyer this case is meant to cover the install and load cases as well because it was originally in LoadModuleReference for the install command case.

src/cls/IPM/Utils/Module.cls
```diff
@@ -1386,6 +1372,44 @@ ClassMethod LoadNewModule(
     quit tSC
 }
 
+/// Check whether an update/install/load from a currently installed module to a new version of the module is a valid operation.
+ClassMethod CheckOperationVersionValidity(
```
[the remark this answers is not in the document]
isc-jili: @isc-eneil That's a good point, but I don't think it can be done. In the case that the update command is not called with -version (installs latest version by default), it still needs to call Install() which does a bunch of resolution before it can identify the newModule object.
LoadNewModule() is the first place that the Install() and Load() (and inherently Update()) logic all intersect, and the first place that the newModule object has been guaranteed to be found.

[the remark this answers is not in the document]
isc-jili: But following up to this, there is definitely the case where -version is specified in the update command, in which case that CAN be checked in Update(). I will add in the short circuit version check case there.

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1687,6 +1689,23 @@ Method %MakeDeployed(ByRef pParams) As %Status
     quit tSC
 }
 
+Method %ApplyUpdateSteps(ByRef pParams) As %Status
+{
+    set sc = $$$OK
+    try {
+        if $system.Mirror.IsMember() && $system.Mirror.IsPrimary() {
+            write !,"WARNING: You are running update steps on the primary member of a mirror. It is recommended to start the update process on a backup instance to minimize downtime."
+            write !,"If you have already run update on the backup instance, please disregard this warning."
```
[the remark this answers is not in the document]
isc-jili: @isc-kiyer bumping this thread since I have the same question as @isc-eneil

src/cls/IPM/Utils/Module.cls
```diff
@@ -480,7 +476,12 @@ ClassMethod LoadModuleFromDirectory(
             // Only propagate selected settings that should apply to dependencies.
             set tParams("DeveloperMode") = $get(pOverrideDeveloperMode,0)
             set tParams("Verbose") = tVerbose
-            set tParams("cmd") = "install"
+            // TODO: Test whether load invokes this path
```
isc-jili, line 479: nit: Remove this TODO. Load does not invoke this path

src/cls/IPM/Utils/Module.cls
```diff
@@ -1272,6 +1284,7 @@ ClassMethod LoadNewModule(
 
         set tModule = ##class(%IPM.Storage.Module).NameOpen(tModuleName,,.tSC)
         $$$ThrowOnError(tSC)
+
```
isc-jili, line 1287: nit: remove newline

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1687,6 +1689,23 @@ Method %MakeDeployed(ByRef pParams) As %Status
...
+        }
+        set verbose = $get(pParams("Verbose"))
+        // TODO: Add log file param after logging meeting discussion
+        do ..Module.ApplyUpdatesInOrder(,verbose)
```
isc-jlechtne, line 1702: HandleAllUpdateSteps(, , verbose)

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -56,6 +56,12 @@ Method %Activate(ByRef pParams) As %Status
                 }
             }
         }
+        if '$get(pParams("Update")) && '..Module.DeveloperMode {
+            if verbose {
+                write !, "Seeding update steps for module = ",..Module.Name, ", for version = ", ..Module.VersionString, !
+            }
+            $$$ThrowOnError(..Module.SeedAllUpdateSteps(verbose))
```
isc-jlechtne, line 63: ..Module.HandleAllUpdateSteps(, 1, verbose)

src/cls/IPM/Lifecycle/Module.cls
```diff
@@ -56,6 +56,12 @@ Method %Activate(ByRef pParams) As %Status
                 }
             }
         }
+        if '$get(pParams("Update")) && '..Module.DeveloperMode {
+            if verbose {
+                write !, "Seeding update steps for module = ",..Module.Name, ", for version = ", ..Module.VersionString, !
```
isc-jlechtne, line 61: This same verbose line is in HandleAllUpdateSteps(). Suggest removing it from here entirely and let HandleAllUpdateSteps() handle the logging (like how you have with executing steps in %IPM.Lifecycle.Base

isc-jlechtne: However, all of our logging with this stuff will be changed in the logging story anyways, so this point isn't that big of a deal

src/cls/IPM/Utils/Module.cls
```diff
@@ -1372,14 +1369,11 @@ ClassMethod LoadNewModule(
     quit tSC
 }
 
-/// Check whether an update/install/load from a currently installed module to a new version of the module is a valid operation.
+/// Check whether an update from a currently installed module to a new version of the module is a valid operation.
 ClassMethod CheckOperationVersionValidity(
```
isc-eneil, line 1373: Since these checks only apply to updates, you could rename this `CheckUpdateVersionValidity()`

isc-eneil (changes requested):

src/cls/IPM/Utils/Module.cls
```diff
@@ -1338,7 +1351,10 @@ ClassMethod LoadNewModule(
         if tVerbose || $get(pParams("Multicompile","Verbose"),0) {
             write !,"Loading "_tDisplayName_" in process "_$job
         }
-        set tSC = ##class(%IPM.Storage.Module).ExecutePhases(tModuleName,$listbuild("Activate"),1,.pParams)
+        // LoadNewModule() may be called on an install/load or an update. In the case of an update, invoke ApplyUpdateSteps lifecycle phase.
```
isc-kiyer, line 1354: Everything after the first sentence is unnecessary since its obvious in the code. Explaining the context of LoadNewModule (when it can get called) should be in the method description and not here

isc-kiyer: In general, comments are good as long as they are explaining overarching parts of code or some slightly obscure looking code (like regex/$zstrip). If the code is clear, then comments get in the way and can easily get outdated (outdated comments are worse than no comments IMO)

src/cls/IPM/Utils/Module.cls
```diff
@@ -1353,6 +1369,38 @@ ClassMethod LoadNewModule(
...
+        set fromVersion = moduleCurrent.Version
+        set fromVersionString = moduleCurrent.VersionString
+    } else {
+        // This means that this is a new install/load of a module. Hence no version from/to validity to check.
+        quit
+    }
+
+    // If the version of the module about to be installed is OLDER than the module that is already installed
+    if fromVersion.Follows(toVersion) {
+        set msg = $$$FormatText("Cannot downgrade %1 from version %2 to %3. To downgrade a module, you must first run the uninstall command. Then, run the install command with the desired module version.",moduleName,fromVersionString,toVersionString)
```
isc-kiyer, line 1393: Nit: second sentence should say "run the install or load command with the desired module version"

src/cls/IPM/Utils/Module.cls
```diff
@@ -1353,6 +1369,38 @@ ClassMethod LoadNewModule(
     quit tSC
 }
 
+/// Check whether an update from a currently installed module to a new version of the module is a valid operation.
+ClassMethod CheckOperationVersionValidity(
+	ByRef newModule,
```
isc-kiyer, line 1374: Nit: newModule need not be passed byref (not making changes) and should have a type

isc-kiyer (changes requested):
@isc-jili few more comments

src/cls/IPM/Utils/Module.cls
```diff
@@ -1233,12 +1238,20 @@ ClassMethod LoadNewModule(
 
         set pDirectory = ##class(%File).NormalizeDirectory("",pDirectory)
 
+        set newModuleObj = ..GetModuleObjectFromPath(pDirectory,.foundNewModuleObj)
+        // Note that for update/install commands, the operation version validity has already been checked in %IPM.Main.Install().
+        if isLoad && foundNewModuleObj {
+            // Check that the module currently installed is not being downgraded. Also check that "load" is not being used to update a module, unless the -force flag is passed.
+            do ..CheckOperationVersionValidity(newModuleObj, .pParams)
```
isc-kiyer, line 1245: I just realized one thing: we may want to confirm at some point in this flow when providing a path, that the path provided, matches the module name provided for update. A user could otherwise accidentally have provided an incorrect tarball and update the wrong module unintentionally

isc-kiyer: Yes checking exactly that. You will need to pass that info down to here from Update since that context is not provided for downstream use when -path is provided right now.

src/cls/IPM/Main.cls
```diff
@@ -3789,4 +3838,41 @@ ClassMethod Information(ByRef pCommandInfo)
...
+ClassMethod Update(ByRef pCommandInfo)
+{
+    if $get(pCommandInfo("data","PermitDowngrade"),0) {
+        $$$ThrowStatus($$$ERROR($$$GeneralError,"The 'PermitDowngrade' flag is not allowed for the update command."))
+    }
+    set verbose = $get(pCommandInfo("data","Verbose"))
+    set version = $get(pCommandInfo("parameters","version"))
+    set path = $$$GetModifier(pCommandInfo,"path")
+    set moduleName = $get(pCommandInfo("parameters","module"))
+    if moduleName = "" {
```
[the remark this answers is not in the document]
isc-jili: Noting here that based on an offline discussion we decided to address all required params verification in %IPM.CLI.%ParseCommandInput().

PR #861, opened 2025-08-11
First round of update tests

isc-eneil:
NOTE: For this first testing PR, I created a simple module with no dependencies and a bunch of basic tests that must be run on this module and in future PRs, on other more complex modules. See below for the specific tests included in this PR compared to the total tests that are part of the update framework test plan.

To run: `verify zpm -only -DUnitTest.Case=Test.PM.Integration.Update`

Issue: https://github.com/intersystems/ipm/issues/550

## Update testing To-Dos:

1. Create the following modules:
- [X] Module without dependencies (simplest case)
- [ ] Module with dependencies
- [ ] Module with dependencies added as part of an update
- [ ] Module with dependencies removed as part of an update
- [ ] Module without any update methods (in either the current or updated module)
- [ ] Module with update methods but no seeded steps
- [ ] Module with new update method added to an older major version
- [ ] Module with changed update method order for the major version to update to
- [ ] Module with an update step that throws an error in a major version less than the major version to which to update
- [ ] Module with an update step that throws an error in a previous major version
- [ ] Module with a dependency whose update step that throws an error
- [ ] Module with an <Invoke> that throws an error during a lifecycle phase
- [ ] Module with a dependency with an <Invoke> that throws an error during a lifecycle phase

2. Create a generic update test to be run on all the modules WITHOUT ERRORS that does the following:
- [X] Note: Run it once with -dev flag
- [X] Rows have been added to the UpdateStep table for each update method in the appropriate range of V# classes for 1) the base module and 2) each of its dependencies
- [X] The correct update steps are run (e.g. starting at the correct step within the correct major version)
- [X] (SKIPPING FOR NOW) Logging is correctly outputted to the (optionally provided) log file or standard output.
- [X] Correct versions of dependency modules are installed

3. Create a generic update test to be run on all the modules WITH ERRORS:
All update-related logic is rolled back (the old module is still installed and no new UpdateStep objects exist)
Errors are logged appropriately in the (optionally provided) log file or standard output.

4. Create tests for the following cases (to be run on the simplest module): 
- [X] Running update twice on the same module to new version is successful
- [X] Running update with an older version throws an error immediately
- [X] Running update with a version that doesn't exist in any configured repository throws an error immediately
- [X] Running the same update command twice in a row results in a no-op on all update steps the second time
- [ ] Running update on a module that has not been installed
- [X] Running update with a -path flag (so loading the newer version from a local tarball)
- [ ] Passing in a -path to update with a module name that does not match the given name fails

5. Create tests for trying to update the same module at the same time in the same namespace fails

7. Create test for running update AFTER an an update process was killed mid-update is successful

8. Create test that verifies the correct steps have been run after update depending on mirror membership. 

9. Add the following tests to the suite of install tests:
- [ ]  Verify that the UpdateStep table is properly seeded at install/load, regardless of whether the module is provided in a tarball or local directory
- [ ] Verify that attempting to run install/load without the force flag fails if a module is already installed
- [ ] Verify that rows in the UpdateStep table are properly removed when the module is uninstalled
- [ ] Verify that, if trying to uninstall a dependency module (should throw an error), it errors out and no update metadata is removed
- [ ] Verify that running install/load when a module is already installed (no matter the version) succeeds in dev mode and fails otherwise

10. Update CI pipeline to add another IRIS instance which is mirrored. Initialize the main IRIS container as the primary and the new container as the backup.

11. Update tests above to be mirror-aware.
- [ ] Check that the correct subset of update steps are run on each mirror member corresponding to its mirror membership types


tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,238 @@
...
+{
+  // Load module
+  Do ..AssertModuleInstalledCorrectly(..#ModuleNoDependencies, ..#ModuleTwoVersion)
+
+  // Update module
+  Set tSC = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleOneVersion)
+  Do $$$AssertStatusNotOK(tSC, "Attempt at updating module to older version failed.")
+}
+
+Method TestUpdateToInvalidVersion()
```
isc-jili, line 84: nit: is it worth putting in method descriptions to clarify the expected behavior being tested in each test case?

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,238 @@
...
+}
+
+Method TestUpdateToNewVersionTwice()
+{
+  // Load module
+  Do ..AssertModuleInstalledCorrectly(..#ModuleNoDependencies, ..#ModuleOneVersion)
+
+  // Update module to a 2.x version
+  Set tSC = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleTwoVersion)
+  Do $$$AssertStatusOK(tSC, "Updated module to "_..#ModuleTwoVersion_"without error")
```
isc-jili, line 101: nit: missing space before “without error”

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,238 @@
...
+
+  // Update module to a 3.x version
+  Set tSC = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleThreeVersion)
+  Do $$$AssertStatusOK(tSC, "Updated module to "_..#ModuleThreeVersion_"without error")
```
isc-jili, line 105: nit: missing space before “without error”

isc-jili:
@isc-eneil a note regarding test cases: It may be worth also writing cases for load and install commands given that they may have some overlap in behavior with update. Some test cases I can think of right now:
- Using install to install a newer version of an already installed module (should fail because should use update instead, unless -force or -dev is used)
- Using load for the same above case
- Calling update on a module that wasn't already installed (should fail because should use install/load instead)

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,261 @@
...
+
+    // NOTE: For testing simplicity, we will always assume that in the update test, both the module and its dependencies are being updated
+    // from and to the same version numbers.
+
+    // Check module version has been updated
+    set module = ##class(%IPM.Storage.Module).NameOpen(moduleName)
+    do $$$AssertEquals(module.VersionString, toVersion)
+
+    // Check that dependency versions have been updated
+    set iter = module.Dependencies.%GetIterator()
```
isc-jlechtne, line 213: Want to use buildDependencyGraph here to capture transient dependencies being updated, not just direct dependencies

tests/integration_tests/Test/PM/Integration/_data/update-test/update-modules-new/update/cls/Update/Update/V2.cls
```diff
@@ -0,0 +1,32 @@
+Class Update.Update.V2 Extends %IPM.General.VersionBase
```
isc-jili, line 1: Should be %IPM.Utils.Update.VersionBase

tests/integration_tests/Test/PM/Integration/_data/update-test/update-modules-old/update/cls/Update/Update/V1.cls
```diff
@@ -0,0 +1,20 @@
+Class Update.Update.V1 Extends %IPM.General.VersionBase
```
isc-jili, line 1: Should be %IPM.Utils.Update.VersionBase

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,261 @@
...
+      set className = value.%Get(0)
+      set methodName = value.%Get(1)
+      set isPrimaryOnly = value.%Get(2)
+
+      set updateStep = ..GetUpdateStep(className, methodName, isPrimaryOnly)
+      do $$$AssertEquals(timeStampArr(updateStep.UpdateStepClassMethodIdx), updateStep.TimeStampEnd, "Seeded update step with class="_className_" and method="_methodName_" was correctly not rerun on update.")
+    }
+
+    // Check whether correct UpdateStep rows exist in the correct table after update command is run
+    set updateSteps = ..AssertUpdateStepsExist(updateStepsToRun)
```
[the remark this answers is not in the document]
isc-eneil: Ooh thanks, that was a relic of a previous way of doing this

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,261 @@
...
+    set className = value.%Get(0)
+    set methodName = value.%Get(1)
+    set isPrimaryOnly = value.%Get(2)
+
+    set updateStep = ..GetUpdateStep(className, methodName, isPrimaryOnly)
+    do $$$AssertNotEquals("", updateStep, "Update step with class="_className_" and method="_methodName_" exists.")
+    do updateStepList.%Push(updateStep)
+  }
+
+  return updateStepList
```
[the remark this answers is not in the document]
isc-eneil: Was trying to avoid duplicating the work of iterating through a list of [class, method, isPrimary], but I was unable to deduplicate due to different logic needing to be applied to each step. I'm gonna change it back so it doesn't return anything.

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+	moduleVersion As %String)
+{
+  set tSC = ##class(%IPM.Main).Shell("install -v "_moduleName_" "_moduleVersion)
+  do $$$AssertStatusOK(tSC, "Loaded module successfully")
+}
+
+ClassMethod GetUpdateStep(
+	className As %String,
+	methodName As %String,
+	isPrimaryOnly As %Boolean) As %IPM.Storage.UpdateStep
```
isc-jili, line 252: %IPM.Storage.UpdateStep no longer exists in the new design

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+  // Setup repo pointing to all modules used for tests in this class
+  set tSC = ##class(%IPM.Main).Shell("repo -n integration-tests -fs -path /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/")
+  do $$$AssertStatusOK(tSC,"Created integration-tests repo successfully.")
+  return tSC
+}
+
+Method OnAfterAllTests() As %Status
+{
+  // Clean up repo used for tests
+  set tSC = ##class(%IPM.Main).Shell("repo -delete -name integration-tests")
```
isc-kiyer, line 63: Nit: replace all tSC with sc

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+
+    // Check module version has been updated
+    set module = ##class(%IPM.Storage.Module).NameOpen(moduleName)
+    do $$$AssertEquals(module.VersionString, toVersion)
+
+    // Check that dependency versions have been updated
+    // dependencyGraph format: dependencyGraph(<module name>) = $ListBuild(<depth>, <server>, <version>)
+    set moduleName = ""
+    for {
+      set moduleName = $order(dependencyGraph(moduleName),1,moduleInfo)
```
isc-kiyer, line 216: Dependency graph isn't built anywhere right now so that should be added

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+      quit:moduleName=""
+
+      set moduleVersion = $list(moduleInfo, 2)
```
isc-kiyer, line 219: @isc-eneil @isc-jili @isc-jlechtne  Looking at this made me realize one thing we do need to do (separate subtask to do later as an improvement): when updating a module, we should check all of its dependencies pre and post update and if there were any in pre that aren't in post, they should be uninstalled.

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+}
+
+Method TestUpdateToOlderVersion()
+{
+  // Load module
+  do ..AssertModuleInstalledCorrectly(..#ModuleNoDependencies, ..#ModuleTwoVersion)
+
+  // Update module
+  set tSC = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleOneVersion)
+  do $$$AssertStatusNotOK(tSC, "Attempt at updating module to older version failed.")
```
isc-kiyer, line 84: Nit: worthwhile checking the status and confirming the error message is what is expected (not usually needed but here it is to ensure the user sees an appropriate helpful error)

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+  set twoCStart = stepTwoC.TimeStampStart
+  set twoCEnd = stepTwoC.TimeStampEnd
+
+  // Update module to SAME version
+  set tSC = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleTwoVersion)
+  do $$$AssertStatusOK(tSC, "Updated module to "_..#ModuleTwoVersion_"without error")
+
+  // Check that second update does not rerun update steps
+  // (timestamps on each update step should not have changed)
+  do $$$AssertEquals(twoAStart, stepTwoA.TimeStampStart)
```
isc-kiyer, line 140: After the update, you will want to call GetUpdateStep again to force reload of the persistence or else it will use the one from the first call (can do so using %Reload() as well instead as well instead of calling GetUpdateStep() again)

isc-kiyer (changes requested):
@isc-eneil overall a good first pass! I like the helper method. My comments are up

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+    set timeStampArr = []
+    set iter = updateStepsToSeed.%GetIterator()
+    while iter.%GetNext(.key, .value) {
+      set className = value.%Get(0)
+      set methodName = value.%Get(1)
+      set isPrimaryOnly = value.%Get(2)
+
+      set updateStep = ..GetUpdateStep(className, methodName, isPrimaryOnly)
+      do $$$AssertNotEquals("", updateStep, "Update step with class="_className_" and method="_methodName_" exists.")
+      set timeStampArr(updateStep.UpdateStepClassMethodIdx) = updateStep.TimeStampStart
```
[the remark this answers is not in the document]
isc-eneil: Just meant to uniquely identify the update step, but I didn't realize I couldn't reference the hash that way. I'll just subscript by class name and method name instead

tests/integration_tests/Test/PM/Integration/Update.cls
isc-eneil: Hm I don't want to create a test that will go stale easily if the error message wording changes. Maybe I could check for something like "version" in the error string?

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,263 @@
...
+Parameter ModuleOneVersion As STRING = "1.2.1";
+
+Parameter ModuleTwoVersion As STRING = "2.1.1";
+
+Parameter ModuleThreeVersion As STRING = "3.5.0";
+
+Method OnBeforeAllTests() As %Status
+{
+  // Setup repo pointing to all modules used for tests in this class
+  set tSC = ##class(%IPM.Main).Shell("repo -n integration-tests -fs -path /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/")
```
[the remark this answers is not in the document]
isc-eneil: Thanks for the reminder!

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,238 @@
...
+    set updateStep = ..GetUpdateStep(className, methodName, isPrimaryOnly)
+    do $$$AssertTrue(%IsObject(updateStep), "Update step with class="_className_" and method="_methodName_" exists.")
+  }
+}
+
+Method AssertModuleInstalledCorrectly(
+	moduleName As %String,
+	moduleVersion As %String)
+{
+  set sc = ##class(%IPM.Main).Shell("install -v "_moduleName_" "_moduleVersion)
```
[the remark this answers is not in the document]
isc-eneil: I agree that both should be tested, but not in this class. They belong in install and load test classes (I have some to-dos in the InstallModule class currently).

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,238 @@
...
+/// @argument fromVersion         Version of module to install
+/// @argument toVersion           Version of module to which to update
+/// @argument developerMode       1 if update should be run with -dev flag, 0 otherwise
+Method ModuleNormalUpdateNoMirrorHelper(
+	moduleName As %String,
+	updateStepsToSeed As %DynamicArray,
+	updateStepsToRun As %DynamicArray,
+	fromVersion As %String,
+	toVersion As %String,
+	developerMode As %Boolean = 0)
```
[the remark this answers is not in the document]
isc-eneil: Since install is called during update, I just want to make sure that nothing about the developer mode logic in that class messes with updates. It's more of a fail-safe to prove that dev mode and non-dev mode DO work the same.
Just added the test case to the list in the PR description.

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -156,9 +170,10 @@ Method ModuleNormalUpdateNoMirrorHelper(
     set timeStampArr(className, methodName, isPrimaryOnly) = updateStep.TimeStampStart
   }
 
-  // Update module
-  set devFlag = $select(developerMode: "-dev", 1: "")
-  set sc = ##class(%IPM.Main).Shell("update -v "_devFlag_" "_moduleName_" "_toVersion)
+  // Update module (with appropriate flags)
+  set devFlag = $select(developerMode: " -dev ", 1: "")
+  set pathFlag = $select(pathToModule: " -path="_pathToModule, 1: "")
+  set sc = ##class(%IPM.Main).Shell("update -v"_devFlag_pathFlag_moduleName_" "_toVersion)
```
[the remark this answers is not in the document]
isc-eneil: I like that option. Just implemented.

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,271 @@
...
+  // Update module to a 2.x version
+  set sc = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleTwoVersion)
+  do $$$AssertStatusOK(sc, "Updated module to "_..#ModuleTwoVersion_" without error")
+
+  // Update module to a 3.x version
+  set sc = ##class(%IPM.Main).Shell("update -v "_..#ModuleNoDependencies_" "_..#ModuleThreeVersion)
+  do $$$AssertStatusOK(sc, "Updated module to "_..#ModuleThreeVersion_" without error")
+}
+
+Method TestUpdateToSameVersionTwiceNoMirror()
```
isc-jili, line 81: Sorry for late comment @isc-eneil but recording the conversation we just had offline here: it seems that between running each test method, there should be some cleanup/uninstall mechanism. Because running install on the same module at a lower version would be considered a downgrade, which is not allowed and would error out.

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -0,0 +1,297 @@
...
+
+    set updateStep = ..GetUpdateStep(className, methodName, isPrimaryOnly)
+    do $$$AssertTrue(%IsObject(updateStep), "Update step with class="_className_" and method="_methodName_" exists.")
+    set timeStampArr(className, methodName, isPrimaryOnly) = updateStep.TimeStampStart
+  }
+
+  // Update module (with appropriate flags)
+  set devFlag = $select(developerMode: "-dev ", 1: "")
+  set pathFlag = $select(pathToModule: "-path="_pathToModule_" ", 1: "")
+  set sc = ##class(%IPM.Main).Shell("update -v "_devFlag_pathFlag_moduleName_" "_toVersion)
```
isc-jili, line 191: Another case to try here is passing it a module name which does not match the module.xml in the -path parameter. The code now has validation to check the module name against that in the module.xml. Should verify that in that case, the incorrect module does not get updated.

isc-jili (approved):

PR #872, opened 2025-08-26
Update framework Tests (Part Two)

isc-eneil:
NOTE: This PR should cover all checkbox items in the "In Scope" section. See below for tests that are a) already written or b) out of scope.

To run: `verify zpm -only -DUnitTest.Case=Test.PM.Integration.Update`

Issue: https://github.com/intersystems/ipm/issues/550

### In Scope
1. Create the following modules:
- [x] Module with dependencies
- [x] Module with dependencies added as part of an update
- [x] Module with dependencies removed as part of an update
- [x] Module without any update methods (in either the current or updated module)
- [x] Module with new update method added to an older major version
- [x] Module with changed update method order for the major version to update to
- [x] Module with an update step to run that throws an error (should fail)
- [x] Module with a seeded update step that throws an error (should succeed)
- [x] Module with a dependency whose update step that throws an error
- [x] Module with an `<Invoke>` that throws an error during a lifecycle phase
- [x] Module with a dependency with an `<Invoke>` that throws an error during a lifecycle phase

2. Add the following test cases to the update suite:
- [x] Create a generic update test to be run on all the modules WITH ERRORS and verify that each module update fails
- [x] Running update on a module that has not been installed fails
- [x] Passing in a -path to update with a module name that does not match the given name fails
- [x] Trying to update the same module at the same time in the same namespace fails
- [x] Running update AFTER an an update process was killed mid-update is successful

3. Add the following tests to the install suite:
- [x]  Verify that the UpdateStep table is properly seeded at install/load, regardless of whether the module is provided in a tarball or local directory
- [x] Verify that attempting to run install/load without the force flag fails if a module is already installed
- [x] Verify that rows in the UpdateStep table are properly removed when the module is uninstalled
- [x] Verify that, if trying to uninstall a dependency module (should throw an error), it errors out and no update metadata is removed
- [x] Verify that running install/load when a module is already installed (no matter the version) succeeds in dev mode and fails otherwise

### Already Merged
1. Create the following module:
- Module without dependencies (simplest case)

3. Create a generic update test to be run on all the modules WITHOUT ERRORS that does the following:
- Note: Run it once with -dev flag
- Rows have been added to the UpdateStep table for each update method in the appropriate range of V# classes for 1) the base module and 2) each of its dependencies
- The correct update steps are run (e.g. starting at the correct step within the correct major version)
- Correct versions of dependency modules are installed

4. Create tests for the following cases (to be run on the simplest module): 
- Running update twice on the same module to new version is successful
- Running update with an older version throws an error immediately
- Running update with a version that doesn't exist in any configured repository throws an error immediately
- Running the same update command twice in a row results in a no-op on all update steps the second time
- Running update with a -path flag (so loading the newer version from a local tarball)

### Out of Scope
1. Update CI pipeline to add another IRIS instance which is mirrored. Initialize the main IRIS container as the primary and the new container as the backup.

2. Update tests above to be mirror-aware.
- Check that the correct subset of update steps are run on each mirror member corresponding to its mirror membership types

3. Add logging checks to the generic update test to be run on all modules WITHOUT errors

4. Add a rollback check to the generic update test run on all the modules WITH ERRORS:
- Verify that, in addition to erroring out the update, all update-related logic is rolled back (the old module is still installed and no new UpdateStep objects exist)

5. Add a check for propagation of Developer Mode on install/load/update (-dev should NOT cause dependencies to also be installed in Developer Mode unless those modules exist in configured source control)

### DIFF
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -1214,6 +1214,7 @@ ClassMethod LoadNewModule(
         if '$data(pParams("qstruct")) {
             set qstruct = "ck"_$select(tVerbose:"d",1:"-d")
         }
+        set commandLineModuleName = $get(pParams("ModuleName"))
 
         // Check for valid license key prior to top-level module load
         // This may throw an exception.
@@ -1236,28 +1237,6 @@ ClassMethod LoadNewModule(
 
         set pDirectory = ##class(%File).NormalizeDirectory("",pDirectory)
 
-        set newModuleObj = ..GetModuleObjectFromPath(pDirectory,.foundNewModuleObj)
-        if foundNewModuleObj {
-            // Check that on an update command with -path flag, the module in the path matches the module specified for update.
-            set commandLineModuleName = $get(pParams("ModuleName"))
-            if (commandLineModuleName '= "") && (newModuleObj.Name '= commandLineModuleName) {
-                set msg = $$$FormatText("The specified path does not contain a module matching the specified module to update. Module in path = %1. Specified module = %2.",newModuleObj.Name,commandLineModuleName)
-                $$$ThrowStatus($$$ERROR($$$GeneralError, msg))
-            }
-            // Note that for update/install commands, the operation version validity has already been checked in %IPM.Main.Install().
-            if isLoad {
-                // Check that the module currently installed is not being downgraded.
-                // Check that the load command is not being used to update a module, unless the -force flag is passed.
-                do ..CheckOperationVersionValidity(newModuleObj, .pParams)
-            }
-        }
-
-        if ($get(pParams("NoJournal"),0) '= 0) {
-            // Disable journalling with process-level switch.
-            set tJournalMgr = ##class(%IPM.Utils.JournalManager).%New(0)
-        } elseif tUseTransactions {
-            tstart
-        }
         // This loads the new version of the module into storage. Overrides the module context of the currently installed module, if it exists.
         set tSC = $system.OBJ.Load(pDirectory_"module.xml",$select(tVerbose:"d",1:"-d"),,.tLoadedList)
         $$$ThrowOnError(tSC)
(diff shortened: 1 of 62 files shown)


tests/integration_tests/Test/PM/Integration/InstallModule.cls
```diff
@@ -1,25 +1,187 @@
...
+{
+    // Using update-test modules for the test of making sure update steps are used correctly with zpm "install"
+    set sc = ##class(%IPM.Main).Shell("repo -n update-test-modules -fs -path /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/update-test/")
+    do $$$AssertStatusOK(sc,"Created update-test-modules repo successfully.")
+    return sc
+}
+
+Method OnAfterAllTests() As %Status
+{
+    // Using update-test modules for the test of making sure update steps are used correctly with zpm "install"
```
isc-jili, line 14: Truly a nit but seems redundant to have same line of comment repeat for OnBefore() and OnAfter()

tests/integration_tests/Test/PM/Integration/InstallModule.cls
```diff
@@ -1,25 +1,187 @@
...
+}
+
 Method TestSimpleApp()
 {
-  Set tSC = $$$OK
-  Try {
-    Set tTestRoot = ##class(%File).NormalizeDirectory($Get(^UnitTestRoot))
+  set tSC = $$$OK
+  try {
+    set tTestRoot = ##class(%File).NormalizeDirectory($get(^UnitTestRoot))
```
isc-jili, line 24: I think we are supposed to avoid using "t" prefixes for net new code

tests/integration_tests/Test/PM/Integration/InstallModule.cls
```diff
@@ -1,25 +1,187 @@
...
+    set sc = ##class(%IPM.Main).Shell("install -v update-simple 3.5.0")
+    do $$$AssertStatusOK(sc,"Installed update-simple")
+
+    // Confirm that table is properly seeded for all update steps with that module
+    set updateStepsToSeed = [["UpdatePackage.V1", "MethodB", 0], ["UpdatePackage.V1", "MethodA", 1], ["UpdatePackage.V1", "MethodC", 1],
+                             ["UpdatePackage.V2", "MethodB", 0], ["UpdatePackage.V2", "MethodA", 1], ["UpdatePackage.V2", "MethodC", 1],
+                             ["UpdatePackage.V3", "MethodA", 1]]
+    do ..ConfirmUpdateStepsAreSeeded(updateStepsToSeed)
+
+    /// Defining paths for module tarball
```
isc-jili, line 48: nit: should be "//" instead of "///"

tests/integration_tests/Test/PM/Integration/InstallModule.cls
isc-jlechtne: This method is not new code. Only changes look like auto-formatting to make keywords lower case

isc-jili (approved):

isc-eneil (commented):
Looks good to me @isc-jlechtne

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -136,11 +300,137 @@ Method TestUpdateToSameVersionTwiceNoMirror()
...
+	seededUpdateStepsToNotRun As %DynamicArray,
+	updateStepsToRun As %DynamicArray,
+	fromVersion As %String,
+	toVersion As %String)
+{
+  // Install module
+  do ..AssertModuleInstalledCorrectly(moduleName, fromVersion)
+
+  // Verify that update steps were correctly seeded and capture the TimeStampStart property for comparison
+  set timeStampArr = []
```
isc-jili, line 352: Question: Is this the way to instantiate a subscripted array? This way seems to be a Dynamic Array which uses %Push and %Pop or at least based on https://docs.intersystems.com/iris20252/csp/docbook/DocBook.UI.Page.cls?KEY=GJSON_iteration#GJSON_iteration_pushpop
I'm also not 100% sure so if this works then no worries

src/cls/IPM/Utils/Module.cls
```diff
@@ -1236,10 +1237,33 @@ ClassMethod LoadNewModule(
...
-        if foundNewModuleObj {
+        // This loads the new version of the module into storage. Overrides the module context of the currently installed module, if it exists.
+        set tSC = $system.OBJ.Load(pDirectory_"module.xml",$select(tVerbose:"d",1:"-d"),,.tLoadedList)
+        $$$ThrowOnError(tSC)
+
+        set tFirstLoaded = $order(tLoadedList(""))
+        if (tFirstLoaded = "") {
+            $$$ThrowStatus($$$ERROR($$$GeneralError,"No module definition found."))
+        }
+        if ($order(tLoadedList(tFirstLoaded)) '= "") {
```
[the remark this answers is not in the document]
isc-eneil: Ohhh that makes sense-- so each `<Document>` tag is loaded as a separate item?

PR #879, opened 2025-09-09
fix #779: Pull in module defaults prior to load

### LINKED ISSUE #779: Pull in module defaults prior to load
I should be able to put in module.xml for a top-level module:
```
      <Defaults>
        <Parameter Name="Threads">0</Parameter>
        <Parameter Name="NoTransaction">1</Parameter>
        <Parameter Name="NoLock">1</Parameter>
      </Defaults>
```

And have this respected in load/install. As it stands, we don't do anything with these until ExecutePhases for the package itself. (See %IPM.Storage.Module:GetDefaultParameters.)

isc-dchui:
Fixes #779
<Defaults> in module.xml are now used during load/install

### DIFF
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -1109,6 +1109,57 @@ ClassMethod GetModuleNameFromXML(
     return $$$OK
 }
 
+/// Parses defaults defined in the module.xml that look like this:
+/// ```
+/// <Defaults>
+///     <Parameter Name="ArbitraryDefault">1234</Parameter>
+///     <Parameter Name="NoTransaction">1</Parameter>
+///     <Parameter Name="NoLock">1</Parameter>
+/// </Defaults>
+/// ```
+/// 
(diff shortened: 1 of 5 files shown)


src/cls/IPM/Utils/Module.cls
```diff
@@ -1213,6 +1266,11 @@ ClassMethod LoadNewModule(
     set tSC = $$$OK
     set tInitTLevel = $tlevel
     try {
+        // check module.xml for defaults
+        set tSC = ..GetModuleDefaultsFromXML(pDirectory, .defaults)
+        $$$ThrowOnError(tSC)
+        merge pParams = defaults
```
isc-tleavitt, line 1272: This should be the other way around - currently a "default" will override a user-specified parameter.

To avoid module.xml defaults leaking back out byref, I'd recommend:
merge params = defaults
merge params = pParams

then use params through the rest of the method.

isc-tleavitt (changes requested):

PR #882, opened 2025-09-10
Add -dev flag to update command

isc-jili:
Add -dev flag to update command. Also remove extraneous checking and setting of command in LoadModuleFromDirectory()


src/cls/IPM/Utils/Module.cls
```diff
@@ -476,9 +476,7 @@ ClassMethod LoadModuleFromDirectory(
             // Only propagate selected settings that should apply to dependencies.
             set tParams("DeveloperMode") = $get(pOverrideDeveloperMode,0)
             set tParams("Verbose") = tVerbose
-            if '$get(pParams("Update"),0) {
-                set tParams("cmd") = "install"
```
isc-eneil, line 480: Why does cmd not need to be set to "install" anymore?

isc-eneil (changes requested):

PR #890, opened 2025-09-16
Fix load bug with invalid oref error

isc-kiyer:
Also fix some possible issues with getting null orefs in update steps and accordingly update downstream logic.
Ensure storage globals do not use % since we need storage to be namespace-specific (not go into IRISSYS).

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -2154,9 +2154,8 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
     if $get(tComponents("scheme")) [ "http" {
         set tDirectoryName=..LoadFromRepo(tDirectoryName,.tParams)
     }
-    if ##class(%File).DirectoryExists(tDirectoryName) {
-        set tParams("DeveloperMode") = $get(tParams("DeveloperMode"), 1)
-    } elseif ##class(%File).Exists(tDirectoryName) && (($$$lcase($piece(tDirectoryName,".", *))="tgz") || ($$$lcase($piece(tDirectoryName,".", *-1, *))="tar.gz")) {
+    set extension = $$$lcase($piece(tDirectoryName,".",*))
+    if ##class(%File).Exists(tDirectoryName) && ((extension="tgz") || (extension="tar.gz")) {
         set tTargetDirectory = $$$FileTempDirSys
         if $get(pCommandInfo("data", "Verbose")) {
             write !,"Extracting archive to ",tTargetDirectory


src/cls/IPM/Storage/UpdateStep/PrimaryOnly.cls
```diff
@@ -13,24 +16,24 @@ Index UpdateClassMethodIdx On (ClassName, MethodName) [ Unique ];
...
-/// If object does not exist, this will return an empty object
+/// Returns object of this class corresponding to unique index in storage table
+/// If object does not exist, this will return a new object with the ClassName and MethodName
+/// properties set. <br />
+/// @Argument className     The class name of the update step. <br />
+/// @Argument methodName    The method name of the update step</param>
 ClassMethod GetUpdateStep(
 	className As %String,
-	methodName As %String) As %IPM.Storage.UpdateStep.PrimaryOnly
+	methodName As %String) As %IPM.Storage.UpdateStep.AnyMember
```
isc-jlechtne, line 26: Should be PrimaryOnly

src/cls/IPM/Main.cls
```diff
@@ -2142,15 +2142,13 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
     set tDirectoryName = $get(pCommandInfo("parameters","path"))
     set tSynchronous = $$$HasModifier(pCommandInfo,"synchronous")
     set tForce = $$$HasModifier(pCommandInfo,"force")
-    set moduleName = pCommandInfo("parameters","module")
```
isc-jili, line 2145: @isc-kiyer I remembered why I'm passing the module name into Load(). It's because when Load() calls LoadNewModule(), there is logic in LoadNewModule() to check that on an update command with -path flag, the module in the path matches the module specified for update. That means that moduleName as defined here (passed down from Update()) is compared against newModuleObj.Name (where newModuleObj is the one in -path).
So it's valid to ADD the set pCommandInfo("data","ModuleName") = moduleName in Update(). But we should not REMOVE these 2 lines in Load().

isc-jili (changes requested):
Please see my inline comment regarding passing module name to Load()

src/cls/IPM/Main.cls
isc-jili: So this should actually be set moduleName = pCommandInfo("data","ModuleName")

src/cls/IPM/Storage/UpdateStep/AnyMember.cls
```diff
@@ -2,19 +2,30 @@
...
 ForeignKey HomeModKey(HomeModule) References %IPM.Storage.Module() [ OnDelete = cascade ];
 
-/// Returns %IPM.Storage.UpdateStep.AnyMember object corresponding to unique index in storage table
-/// If object does not exist, this will return an empty object
+/// Returns object of this class corresponding to unique index in storage table
+/// If object does not exist, this will return a new object with the ClassName and MethodName
+/// properties set. <br />
+/// @Argument className     The class name of the update step. <br />
+/// @Argument methodName    The method name of the update step</param>
 ClassMethod GetUpdateStep(
```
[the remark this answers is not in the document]
isc-eneil: Good point @isc-jlechtne . Yes, this breaks our current way we check whether an UpdateStep has been correctly created (or not). I call GetUpdateStep() and then check whether the result is an object. Would a better way be to call `..UpdateClassMethodIdxExists(className, methodName, .stepId)` directly in the test class @isc-kiyer ?

isc-eneil (approved):

src/cls/IPM/Main.cls
isc-jili: Got it thanks @isc-kiyer 
I also tested using your branch hsieo-13521-fix-load-bug and can confirm that this does work.

src/cls/IPM/Main.cls
```diff
@@ -2142,15 +2142,13 @@ ClassMethod Load(ByRef pCommandInfo) [ Internal ]
...
     // In the case of an update command, pCommandInfo has already been populated in ..Update(). The relevant information was then moved to tParams via the merge command above.
     // In the case of a load command, tParams still needs to be populated here.
     if '$get(tParams("Update"),0) {
         set tParams("cmd") = "load"
         set tParams("Load") = 1
     }
-    set tParams("ModuleName") = moduleName
     set tParams("Force") = tForce
     do ##class(%Net.URLParser).Decompose(tDirectoryName, .tComponents)
     if $get(tComponents("scheme")) [ "http" {
```
isc-jili, line 2154: @isc-kiyer unrelated to the bug you're trying to fix, but since you're touching this method, could you change the line
set tParams("DeveloperMode") = $get(tParams("DeveloperMode"), 1)
to become
set tParams("DeveloperMode") = $get(tParams("DeveloperMode"), **0**)
?
This is a bug that has been present in the main branch for a while that I just discovered

isc-jili (changes requested):
Please see my comment re: fixing default for DevMode in Load()

src/cls/IPM/Storage/UpdateStep/AnyMember.cls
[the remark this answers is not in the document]
isc-kiyer: @isc-eneil yes either that or check if the resulting object returns non-null for %Id(). That would return empty for new objects but something for existing.

isc-jili (approved):

PR #891, opened 2025-09-16
Fix v0.10.3 release blockers

isc-dchui:
This addresses a few issues from the v0.10.3 changes:

- Fix SemVer parsing bug (surfaced during regression testing of packages)
- Refactor code that gets module default parameters
- Update Dockerfile to pull from containers.intersystems.com instead of irepo.intersystems.com so GitHub Actions workflows can use anonymous tokens
- Prevent Test Major Packages workflow from breaking due to packages with spaces in their names

Running the Test Major Packages workflow with a 0.10.2 baseline shows no new failing package tests.
[v0.10.3](https://github.com/intersystems/ipm/actions/runs/17774805172/job/50519747348) vs [v0.10.2](https://github.com/intersystems/ipm/actions/runs/17775557983/job/50522230457)

### DIFF
--- a/src/cls/IPM/General/SemanticVersionExpression.cls
+++ b/src/cls/IPM/General/SemanticVersionExpression.cls
@@ -25,18 +25,36 @@ Method ExpressionSet(value As %String) As %Status
   }
   set status = $$$OK
   try {
-    set options = $listfromstring(value,"||")
-    set ptr = 0
-    while $listnext(options,ptr,rangeExpr) {
-      set rangeExpr = $zstrip(rangeExpr,"<>W")
-      set status = ##class(%IPM.General.SemanticVersionExpression.Range).FromString(rangeExpr,.option)
-      if $$$ISERR(status) {
-        quit
+    if value["&&" {
+      set reqs = $listfromstring(value,"&&")
+      set ptr = 0
+      while $listnext(reqs,ptr,rangeExpr) {
+        set rangeExpr = $zstrip(rangeExpr,"<>W")
+        set rangeExpr = $zstrip(rangeExpr, "*", "()")
+        set status = ##class(%IPM.General.SemanticVersionExpression.Range).FromString(rangeExpr,.option)
+        if $$$ISERR(status) {
+          quit
+        }
+        if option.ToResolvedString() = "" {
+          continue
+        }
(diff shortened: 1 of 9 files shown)


src/cls/IPM/General/SemanticVersionExpression.cls
```diff
@@ -57,14 +75,23 @@ Method ToResolvedString() As %String
...
     }
     quit $listtostring(optList," || ")
   }
   if ..AllOf.Count() {
     for i=1:1:..AllOf.Count() {
       set opt = ..AllOf.GetAt(i)
-      set optList = optList_$listbuild("(" _ opt.ToResolvedString() _ ")")
+      set resolvedString = opt.ToResolvedString()
+      if ..AllOf.Count() > 1 {
+        set optList = optList_$listbuild("(" _ resolvedString _ ")")
```
isc-kiyer, line 91: Why can parentheses not always be used? Seems simpler that way right? Both here and in ExpressionSet where you then won't need to strip the parentheses

src/cls/IPM/General/SemanticVersionExpression/Comparator.cls
```diff
@@ -42,6 +42,7 @@ ClassMethod FromString(
 	pExpr As %String,
 	Output pComparator As %IPM.General.SemanticVersionExpression.Comparator) As %Status
 {
+  set pExpr = $zstrip(pExpr, "*", "()")
```
isc-kiyer, line 45: Parsing should be updated to handle parentheses instead of stripping them away everywhere IMO

src/cls/IPM/Storage/Module.cls
```diff
@@ -593,6 +593,7 @@ Method GetDefaultParameters(Output pParams)
 
             // Convert "."-delimited pieces to subscripts in pParams
             set tPieces = $listfromstring(tDefault.Name,".")
+            kill tValue
```
isc-kiyer, line 596: Why is this needed if tValue is set right under it? Or could this be multidimensional?

isc-kiyer (commented):
@isc-dchui looks good! Just the one question

isc-tleavitt (approved):

Issue #494, opened 2024-06-05
How to suppress license check during image build?


isc-tstolker:
Right now loading a package using ZPM during the Docker Image build phase fails due to the license check.
Is there a way to suppress this license check?

isc-tleavitt:
There's currently no way to suppress it. We should add one. Calling this a "bug" rather than enhancement because of the severe impact.

@isc-tstolker what does this look like in the docker logs? Could you add a snippet?

isc-tstolker:
![image](https://github.com/intersystems/ipm/assets/9209532/50fc4f9d-03dd-42eb-a427-1b5ef66a15f0)

isc-tstolker:
In the ZPM ##class(%ZPM.PackageManager.Developer.Utils).CheckLicenseKey() we found this only happens in the so-called programmer mode, but I have not been able to find a way to move away from programmer mode.

isc-tleavitt:
More detail on what can go wrong with no license (based on the implementation of this check in the HS package manager about 6 years ago, internal reference HSAI-587):
Installation of packages is multithreaded. Without a license key activated, a developer experienced a hang during package installation as part of a build process. I believe this was because of background jobs failing to start because the license limit on the system had been exceeded.

bdeboe:
I ran into a different symptom of presumably the same issue when ipm is installed and used as part of a script during a docker build. When there is no license key, the interactive prompt does not get a response and goes into a crazy loop with <ENDOFFILE> errors until I abort the build.

```
=============================================================================
|| Welcome to the Package Manager Shell (ZPM). Version: 0.9.2 ||
|| Enter q/quit to exit the shell. Enter ?/help to view available commands ||
|| No registry configured ||
=============================================================================
zpm:SC>
Your system has no (valid) license key present
Copy a valid license key into /usr/irissys/mgr/
and then enter 'y' to activate it. Enter 'n' to continue without a license key,
or 'q' to quit.
Action > Yes =>
Please enter Yes or No
Action > Yes =>
ERROR activating license:
ERROR #6082: License upgrade error: 'Could not upgrade local key, see messages.log.'.
Your system has no (valid) license key present
Copy a valid license key into /usr/irissys/mgr/
and then enter 'y' to activate it. Enter 'n' to continue without a license key,
or 'q' to quit.
Action > Yes =>
Please enter Yes or No
Action > Yes =>

(a bunch more of this yes/no conversationgoes here)

Please enter Yes or No
Action > Yes =>

[...]

Issue #377, opened 2023-03-17
Need support to check proper IRIS license when installing a package


isc-mzhou:
As we plan to launch new supply chain product through IPM, we would like to check if the IRIS instance has the proper license for the installation. For example, IRIS instance without a proper IRIS for Supply Chain license should not be allowed to install package "isc-supply-chain". See Slide 7 in this ppt file for proposed changes in IRIS for Supply Chain license key, which can be used to verify if an IPM installation should be allowed.  https://intersystemscorporation-my.sharepoint.com/:p:/g/personal/mzhou_intersystems_com/EdAvFHy0DHhNvI_ZWvE8pfQB-7fH0nzf26TGqCc1kpDEfQ?e=ZqsES8

daimor:
So, I suppose it just requires a specific feature flag in license. Then it can be implemented through the additional attribute in SystemRequirements tag

isc-mzhou:
Since this will be applicable beyond supply chain, I will leave it to IPM PM to decide the best way forward. I only provided one possible solution.

daimor:
I have no access to the slide, so, I just guessed

isc-mzhou:
Somehow I cannot find your name in sharepoint to give you access. so I am pasting the slide here:
![image](https://user-images.githubusercontent.com/92470455/225982707-b051c5a7-fffa-4a9c-b6a4-b128ec594e7f.png)

daimor:
Yeah, because I'm not part of the company 

Well, using description part of the license does not look good, it should be through the Feature Flag 

$system.License.GetFeature(X)
https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25SYSTEM.License

On screenshot it shown as Extended Features

isc-mzhou:
Thanks for sharing the link. Looks like ideal solution is to make this work: $System.License.IsAppLicensed("isc-supply-chain") somehow. Please work with Bob on the proper solution, and supply chain will follow the same practice.

isc-tleavitt:
I think this is a case for https://github.com/intersystems-community/zpm-registry/issues/66 rather than changes to IPM itself. cc @evshvarov

kuszewski:
I see two needs here:  

1. ZPM Server needs to know about your organization and licensed products to let you download only the right products.  (As is the case with the WRC downloads and with the InterSystems Container Registry today.

[...]

PR #894, opened 2025-09-17
Fix 885: Make loading of dependencies always synchronous

### LINKED ISSUE #885: Remove multi threaded load of dependencies
Multi-threaded load of module dependencies adds a lot of complexity and often causes more issues with lock contention than speeding up the load. We have. a flag right now -synchronous that makes this synchronous but this should be the default and the multi-threading should be removed.
We should deprecate the flag to indicate that is the default (do not remove yet as it would be backwards incompatible to remove).

isc-kiyer:
Summary of changes:
- Remove InstallContext which was used to keep track of modules being loaded across processes but now everything is in the parent process.
- Remove calls to LoadDependencies() from anywhere other than module installation since the side effect causes more confusion than assistance since the user is not expecting dependencies to change out from under them.
- Remove $$$IsISCWorker references and temp global tracking of modules being loaded in other processes
- Deprecate the synchronous flag (add generic CLI support for deprecating modifiers/parameters)
- Remove logic where repo cache is always rebuild when `GetPackageService()` is called which slows down builds

resolves #885

### DIFF
--- a/src/cls/IPM/CLI/Commands.cls
+++ b/src/cls/IPM/CLI/Commands.cls
@@ -48,7 +48,7 @@ XData Schema [ Internal, MimeType = application/xml ]
     <!-- Parameter required? -->
     <xs:attribute name="required" type="xs:boolean" use="optional" default="false"/>
 
-    <!-- Whether the modifier is deprecated -->
+    <!-- Whether the parameter is deprecated -->
     <xs:attribute name="deprecated" type="xs:boolean" use="optional" default="false"/>
 
     <!-- Description of what this parameter does -->
--- a/src/cls/IPM/Storage/Module.cls
+++ b/src/cls/IPM/Storage/Module.cls
@@ -805,9 +805,7 @@ ClassMethod HasScope(
 /// <li>Version string of the module to be installed</li>
 /// </ul>
 /// <p>
-/// If <var>pDepth</var> is 1, only direct dependencies are resolved.
-/// If it is 2, dependencies of dependencies are also resolved, and so on.
-/// If it is 0, dependencies are resolved to any depth.
+/// <var>pDepth</var> specifies how deep dependencies have been resolved so far. It starts at 1 for direct dependencies of the root module, and increments by 1 for each level of recursion.
 /// <p>
 /// If <var>pForceSnapshotReload</var> is 1, then any cached snapshot of available modules is reloaded from the repository server(s).
 /// <p>
--- a/src/cls/IPM/CLI.cls
+++ b/src/cls/IPM/CLI.cls
@@ -664,11 +664,11 @@ ClassMethod %ParseCommandInput(
                     if (tKey = "") {
                         quit
                     }
-                    // Only include deprecated subscript if command is deprecated
+                    // Only include deprecated subscript if parameter/modifier is deprecated
                     set tIsDeprecated = $get(tCommandStructure(tCommandName,tType,tKey,"deprecated"),0)
                     if tIsDeprecated {
                         set pCommandInfo(tType,tKey,"deprecated") = 1
-                        }
+                    }
                 }
             }
         }


src/cls/IPM/CLI/Commands.cls
```diff
@@ -315,9 +323,10 @@ ClassMethod %GetOneCommandStructure(Output pCommandStructure) [ CodeMode = objec
                                     do %code.WriteLine(" Set pCommandStructure("_tName_","_$$$QUOTE(tMap(tCommChild))_") = "_tDescContent)
                                 }
                             } else {
-                                // Common: name, required, description
+                                // Common to bother modifier and parameter
```
isc-dchui: Nit: typo?

isc-jili:
Just wondering out loud if there is a way to attach "doc" to code changes on Github (probably not). But a reminder to update the Wiki with the new CLI changes once live!

src/cls/IPM/Utils/Module.cls
```diff
@@ -1261,15 +1047,6 @@ ClassMethod LoadNewModule(
         }
         set commandLineModuleName = $get(params("ModuleName"))
 
-        // Check for valid license key prior to top-level module load
-        // This may throw an exception.
-        if '$data($$$ZPMHandledModules($namespace)) {
-            set tLicenseIsValid = ..CheckLicenseKey()
```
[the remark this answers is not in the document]
isc-dchui: Here are two relevant license checking issues for consideration too:
https://github.com/intersystems/ipm/issues/377
https://github.com/intersystems/ipm/issues/494

src/cls/IPM/CLI/Commands.cls
```diff
@@ -48,6 +48,9 @@ XData Schema [ Internal, MimeType = application/xml ]
     <!-- Parameter required? -->
     <xs:attribute name="required" type="xs:boolean" use="optional" default="false"/>
 
+    <!-- Whether the modifier is deprecated -->
```
isc-eneil: Nit: Whether the *parameter* is deprecated

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -918,10 +918,6 @@ Method %Compile(ByRef pParams) As %Status
             $$$ThrowOnError(##class(%Studio.SourceControl.Interface).SourceControlCreate())
         }
 
-        // Load dependencies correctly for the requested phases. (This will be a no-op in most cases.)
-        set tSC = ..Module.LoadDependencies(..PhaseList,.pParams)
```
isc-eneil: @isc-kiyer Can you remind me what the original reasoning for reloading dependencies in each phase was?

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -46,6 +46,7 @@ Method OnBeforeAllTests() As %Status
 Method TestAllNoErrorModulesNoMirror()
 {
   if $system.Mirror.IsMember() {
+    do $$$AssertSkipped("Skipping test since this is a mirror member")
```
isc-eneil: Did you intend to remove this? This test should show as having been skipped (don't want to falsely show it as passing)

src/cls/IPM/Repo/Filesystem/Definition.cls
```diff
@@ -119,7 +115,7 @@ Trigger RootChanged [ Event = UPDATE, Foreach = row/object ]
 }
```
isc-eneil: Add a method description with a brief explanation for why pPurge is always 1 for us in the future

src/cls/IPM/Storage/Module.cls
```diff
@@ -806,67 +797,45 @@ ClassMethod HasScope(
...
-                quit
-            }
-        }
-    } catch e {
-        set tSC = e.AsStatus()
-    }
-    quit tSC
-}
-
+/// Builds a dependency graph for this module, recursively resolving dependencies to the specified depth.
```
isc-eneil: Yay! Thanks for the detailed explanation of this method @isc-kiyer  💯

isc-eneil (changes requested):
I did a first round of review! Looks good @isc-kiyer

tests/integration_tests/Test/PM/Integration/Update.cls
isc-kiyer: This is adding a logging line to indicate the test is skipped instead of skipping silently.

src/cls/IPM/Repo/Filesystem/Definition.cls
isc-kiyer: Its not always 1. That is just the default. The prior behavior was to always try to call BuildCache() when GetPackageService() is called as long as we are not in a worker process. Now we have gotten rid of worker processes altogether so the default becomes 1 which is the equivalent

isc-eneil (approved):

src/cls/IPM/Utils/Module.cls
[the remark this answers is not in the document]
isc-tleavitt: The main reason for license checking was because we'd hit issues on multithreading module load, IIRC. With a single thread it should be fine. I think fine to drop. (Just another benefit here!)

PR #897, opened 2025-09-18
Fix CLI parser issue where -path data gets stuffed into -verbose position

isc-jili:
Bug - Command from CLI being parsed incorrectly:
zpm:USER>update -v -path /path/to/local/module/ update-simple
tParsedCommandInfo="update"
tParsedCommandInfo("data","Verbose")="/path/to/local/module/" tParsedCommandInfo("data","zpm","Verbose")="/path/to/local/module/" tParsedCommandInfo("parameters","module")="update-simple"

Fixed by resetting tDataName on each iteration where a modifier is being set in %ParseCommandInput()


isc-kiyer:
@isc-jili this is a general fix so I think is worth going directly to main. Also, please add unit tests
cc @isc-tleavitt @isc-dchui

PR #898, opened 2025-09-18
Fix 815: Packaging/publishing of module with deployed code will always occur in current namespace

### LINKED ISSUE #815: Publishing with deployed code in a temporary namespace not working
When trying to publish with deployed code, the PrepareDeploy phase is added to list to be executed. This phase creates a temporary namespace with copied repos to run the rest of the steps of publishing in. However, doing this we run into the following error:

`ERROR! Object open failed because 'Name' key value of 'healthshare.testipmmod' was not found`

A workaround was put in place for [issue-774](https://github.com/intersystems/ipm/pull/774) to only add the PrepareDeploy phase if code is in dev mode since we should be able to publish just fine with the code not being edited outside of dev move. But we still need to look into the issue of what is going on in this temporary namespace to get deployed code to publish while in dev mode as well.

isc-dchui:
Fixes #815

### DIFF
--- a/src/cls/IPM/Storage/Module.cls
+++ b/src/cls/IPM/Storage/Module.cls
@@ -682,11 +682,6 @@ ClassMethod ExecutePhases(
     } catch e {
         set tSC = e.AsStatus()
     }
-    if $$$ISERR(tSC) && $data(pParams("PackageDeployNamespace"),tPackageDeployNamespace) &&(tPackageDeployNamespace'="") {
-        set $namespace = $get(pParams("InitNamespace"))
-        do ##class(%IPM.Utils.Build).DeleteNamespace(tPackageDeployNamespace, tVerbose)
-        kill pParams("PackageDeployNamespace"), pParams("InitNamespace")
-    }
     quit tSC
 }
 
--- a/src/cls/IPM/Lifecycle/Module.cls
+++ b/src/cls/IPM/Lifecycle/Module.cls
@@ -138,16 +138,6 @@ Method %Package(ByRef pParams) As %Status
         set tSC = $$$ADDSC(tSC, ##class(%IPM.Utils.File).RemoveDirectoryTree(tExportSubDirectory))
     }
 
-    if '$get(pParams("Package","KeepNamespace")) {
-        set tPackageDeployNamespace = $get(pParams("PackageDeployNamespace"))
-        set tInitNamespace = $get(pParams("InitNamespace"))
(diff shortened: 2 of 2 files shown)


src/cls/IPM/Storage/Module.cls
```diff
@@ -332,7 +332,7 @@ ClassMethod ExecutePhases(
 
         if pIsComplete {
             set tPhases = tLifecycle.GetCompletePhases(pPhases)
-            if (($listfind(tPhases, "Package") || $listfind(tPhases, "Publish")) && tModule.HaveToDeploy() && tModule.DeveloperMode) {
+            if (($listfind(tPhases, "Package") || $listfind(tPhases, "Publish")) && tModule.HaveToDeploy() ) {
                 set tPhases = $listbuild("PrepareDeploy") _ tPhases
```
isc-kiyer, line 336: I think this should remain as we only ever want PrepareDeploy (where we install the module in a separate namespace for deployment purposes) if its dev mode and you don't want the dev namespace to be affected. If its not dev, it means this is running in a build/ci pipeline where we are going to throw away the namespace anyway after deploying

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -584,7 +584,7 @@ Method %Reload(ByRef pParams) As %Status
         set tSkipScoped = 'tDeveloperMode
         set tVerbose = $get(pParams("Verbose"))
         set tRoot = $get(pParams("RootDirectory"))
-        if (tRoot = "") && tDeveloperMode {
+        if (tRoot = "") && (tDeveloperMode || ($get(pParams("PackageDeployNamespace")) '= "")) {
```
isc-kiyer, line 587: Can you explain why this is needed? From my understanding, PrepareDeploy() runs after reload and simply maps everything into the newly created deploy namespace (doesn't actually load anything in).

isc-kiyer (changes requested):
@isc-dchui a comment and a suggestion

src/cls/IPM/Lifecycle/Base.cls
isc-kiyer: Ah ok that makes sense then regarding PrepareDeploy. Follow up: do we want to reload all the resources if deploying in a separate namespace? I thought everything is just mapped via PrepareDeploy? Also, when using PrepareDeploy, does the export with dependencies stuff work i.e. are dependencies also available in the new namespace or just the module?

Yes when in non-dev mode, we should just do it from the existing namespace rather than creating a scratch namespace since the only reason to install a non-dev mode package and then run publish is in a CI pipeline.

PR #912, opened 2025-09-30
Fix some issues with uninstall -all (#58 and #908)

### LINKED ISSUE #58: uninstall allows you to remove dependent class
I suggest showing an alert  if there are modules  that depend on deleted module

### LINKED ISSUE #908: uninstall -f -all fails if dependency resolution fails
`uninstall -f -all` should force uninstallation of all modules regardless of whether BuildDependencyGraph() succeeds.

To reproduce:
`install dsw`
`repo -delete-all`
`uninstall -f mdx2json`
`uninstall -f -all`

isc-dchui:
Fixes #58 
Fixes #908

Rewrites UninstallAll() to use the module and dependency information in %IPM.Storage.Module instead of relying on BuildDependencyGraph().

(Side note: hilariously enough before this fix, `uninstall -f <dependent-module>` succeeded in uninstalling but `uninstall -f -verbose <dependent-module>` did not.)

### DIFF
--- a/src/cls/IPM/Lifecycle/Base.cls
+++ b/src/cls/IPM/Lifecycle/Base.cls
@@ -255,46 +255,46 @@ Method CheckBeforeClean(
 	ByRef pParams,
 	Output pSkip As %Boolean = 0) As %Status
 {
-  set tSC = $$$OK
-  try {
-    set tVerbose = $get(pParams("Verbose"))
-    set tLevel = $get(pParams("Clean","Level"),0)
-    set tRecurse = $get(pParams("Clean","Recurse"),1)
+    set tSC = $$$OK
+    try {
+        set tVerbose = $get(pParams("Verbose"))
+        set tLevel = $get(pParams("Clean","Level"),0)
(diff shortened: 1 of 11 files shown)


src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -255,48 +255,46 @@ Method CheckBeforeClean(
...
-            }
-
-            set tModList = ""
-            for i=1:1:tList.Count() {
-                set tName = tList.GetAt(i).Name
-                if '$data(pParams("Clean","Cycle",tName)) {
-                    set tModList = tModList_$listbuild(tName)
-                }
-            }
+  set tSC = $$$OK
```
isc-kiyer, line 258: @isc-dchui Looks like the local folder settings.json did not get used for formatting rules for some reason? That has tabSize set to 4 which would make the diff without the format differences. Could you adjust the tab size back to 4?

isc-kiyer: Oh yeah probably should remove it or keep it in sync with the vs code settings file

isc-kiyer (approved):

PR #916, opened 2025-10-06
Packaging with Python Dependencies #938 Add flag export-python-deps to package command

### LINKED ISSUE #938: Add flag export-python-deps to package command
Adding to support for packaging with third-party python dependencies.

IPM Changes:
Add -export-python-deps flag to package command
Update the package logic in IPM so that, if the -export-python-deps flag is provided,
1. it downloads all relevant .whl files to <moduleroot>/wheels/ directory
2. it adds PythonWheel tags to the module.xml of the module to be packaged

Testing:
Add a test to verify that requirements.txt behavior with and without overlapping PythonWheels is as expected when the instance does not have access to the internet
PythonWheels should be installed first, before those defined in requirements.txt
Then, if there is no internet...
but the python package has already been installed, then requirements.txt should not throw an error, but should log that is it not able to pull down python package versions from the pip registry
but the python package has NOT been installed, then requirements.txt should throw an error
If there IS internet access, then installing python deps from requirements.txt should proceed normally (overwriting any already-installed python wheels)

isc-jili:
Resolves #938 

This is for HSIEO-13097: Support for Packaging with Python Dependencies.

IPM Changes:
Add -export-python-deps flag to package command
Update the package logic in IPM so that, if the -export-python-deps flag is provided,
it downloads all relevant .whl files to /wheels/ directory and
it adds PythonWheel tags to the module.xml of the packaged module

Testing the following:

- If there is no requirements.txt, PythonWheels should be installed as the default.
- If there is requirements.txt and wheels present, and the system is Online, requirements.txt overrides wheels
- If there is requirements.txt and wheels present, and the system is Offline, wheels overrides requirements.txt
- If there is requirements.txt and no wheels present, and the system is Offline, throws error

Note: to simulate offline environment, --no-index is passed to pip in requirements.txt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,16 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - #938 Added flag -export-python-deps to package command
 
-### Changed
-
-### Deprecated
-
 ### Removed
 - #938 Removed secret flag NewVersion handling in %Publish()
 
 ### Fixed
-
-### Security
+- #943: The `load` command when used with a GitHub repository URL accepts a `branch` argument again
 
 ## [0.10.4] - 2025-10-21
 
--- a/module.xml
+++ b/module.xml
@@ -2,7 +2,7 @@
 <Export generator="Cache" version="25">
 <Document name="ZPM.ZPM"><Module>
   <Name>ZPM</Name>
-  <Version>0.10.4-SNAPSHOT</Version>
+  <Version>0.10.5-SNAPSHOT</Version>
   <ExternalName>IPM</ExternalName>
   <Description>InterSystems Package Manager (IPM) provides development tools and infrastructure for defining, building, distributing, and installing modules and applications.</Description>
   <Keywords>Package Manager</Keywords>
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -2122,7 +2122,7 @@ ClassMethod LoadFromRepo(
     set:$extract(tDirectoryName,*)="/" tDirectoryName=$extract(tDirectoryName,1,*-1)
     set RepoName=$piece($piece($piece(tDirectoryName,"/",*),".git")," ")
     set tCmd="cd "_TempDir_" && git clone "_tDirectoryName
-    if $get(tParams("zpm","Branch"))'="" set tCmd=tCmd_" -b "_tParams("zpm","Branch")
+    if $get(tParams("Branch"))'="" set tCmd=tCmd_" -b "_tParams("Branch")
     $$$ThrowOnError(##class(%IPM.Utils.Module).RunCommandViaZF(tCmd,.tLog,.tErr))
     set tDirectoryName = TempDir_slash_RepoName
     if ($get(tParams("Verbose"))) {


src/cls/IPM/StudioDocument/Module.cls
```diff
@@ -419,6 +419,67 @@ ClassMethod UpdateVersion(
...
+        <xsl:value-of select="$wheelName"/>
+      </xsl:element>
+    </xsl:element>
+
+  </xsl:template>
+</xsl:stylesheet>
+}
+
+/// Compile AppendWheelUnderModuleTransform with concrete values
+ClassMethod CompiledAppendWheelUnderModuleXSLT(wheelName As %String) As %XML.XSLT.CompiledStyleSheet
```
isc-dchui: Nit: should probably be the present tense verb "compile" instead of "compiled"

src/cls/IPM/StudioDocument/Module.cls
```diff
@@ -419,6 +419,67 @@ ClassMethod UpdateVersion(
...
+{
+    set xData = ##class(%Dictionary.CompiledXData).IDKEYOpen($classname(),"AppendWheelUnderModuleTransform",,.sc)
+    $$$ThrowOnError(sc)
+    set s = ""
```
isc-eneil: Nit: use more descriptive variable names so this is more readable!

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -741,7 +741,16 @@ Method InstallPythonRequirements(
         }
         set tSC = ##class(%IPM.Utils.Module).RunCommand(pRoot, command,.stdout)
         $$$ThrowOnError(tSC)
-
+        set wheelsPath = ##class(%File).NormalizeFilename("wheels", pRoot)
```
isc-kiyer: Should this be moved to be part of HandlePythonDependencies instead? Right now, all wheels would always be installed but we only need them during packaging

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1210,6 +1223,45 @@ Method %Export(
...
+    set sc = $$$OK
+    try {
+        set wheelsDir = ##class(%File).NormalizeDirectory("wheels", ..Module.Root)
+        if '##class(%File).DirectoryExists(wheelsDir) {
+            $$$ThrowStatus($$$ERROR($$$GeneralError, "Wheels directory not found."))
+        }
+        set stmt = ##class(%SQL.Statement).%New()
+        set status = stmt.%PrepareClassQuery("%File", "FileSet")
+        if $$$ISERR(status) {
+            write "%Prepare failed:"
```
isc-kiyer: Nit: no need for this write. Simply $$$ThrowOnError() is good

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1210,6 +1223,45 @@ Method %Export(
...
+        }
+        while rset.%Next() {
+            set file = rset.%Get("ItemName")
+            set type = rset.%Get("Type")
+            if (type '= "F") || (file = "") {
+                continue
+            }
+            do ##class(%IPM.StudioDocument.Module).AddPythonWheels(..Module.Name, file)
+        }
+        if (rset.%SQLCODE < 0) {
```
isc-kiyer: No need to check SQLCODE at the end here

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1210,6 +1223,45 @@ Method %Export(
...
+            write "%Prepare failed:"
+            do $system.Status.DisplayError(status)
+            quit
+        }
+        set rset = stmt.%Execute(wheelsDir, "*.whl")
+        if (rset.%SQLCODE '= 0) {
+            write "%Execute failed:", !, "SQLCODE ", rset.%SQLCODE, ": ", rset.%Message
+            quit
+        }
+        while rset.%Next() {
```
isc-kiyer: Check status returned by %Next() and have a throwonerror after the loop

tests/integration_tests/Test/PM/Integration/ProcessPythonWheel.cls
```diff
@@ -3,30 +3,155 @@ Class Test.PM.Integration.ProcessPythonWheel Extends Test.PM.Integration.Base
...
+    try {
+        set dummyPackage = ##class(%SYS.Python).Import("ipm_dummy_package")
+        do $$$AssertSuccess("Successfully imported python-wheel resource")
+    } catch ex {
+        do $$$AssertFailure("Failed to import python-wheel resource: "_ex.AsStatus())
+    }
+}
+
+/// Testing case where only Python wheel and no requirements.txt is present
+Method TestWheelPresent()
```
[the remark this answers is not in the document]
isc-jili: You're right that they are similar. However TestWheelPresent() tests with a wheel from Pypi.org whereas the one above uses a generated wheel. And TestWheelPresent() also does an extra check on the wheel version. So I will keep both as I don't think there's much harm having an extra test case

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1210,6 +1229,45 @@ Method %Export(
...
+            }
+            do ##class(%IPM.StudioDocument.Module).AddPythonWheels(..Module.Name, file)
+        }
+        $$$ThrowOnError(sc)
+    } catch ex {
+        $$$ThrowStatus(ex.AsStatus())
+    }
+}
+
+Method %PrepareDeploy(ByRef pParams) As %Status [ Abstract ]
```
isc-kiyer: Delete PrepareDeploy (seems like an issue from resolving merge conflicts)

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -700,10 +700,11 @@ Method %Reload(ByRef pParams) As %Status
 
 Method InstallPythonRequirements(
 	pRoot As %String = "",
-	ByRef pParams)
+	ByRef pParams,
+	download As %Boolean = 0)
 {
     set cmd = $get(pParams("cmd"))
-    if '$listfind($listbuild("install", "load"), cmd) {
+    if ('download && '$listfind($listbuild("install", "load"), cmd)) {
```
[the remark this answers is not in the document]
isc-eneil: @isc-jili What if we updated the method name to be called `InstallOrDownloadPythonRequirements()` to reflect this?
cc @isc-kiyer @isc-jlechtne

PR #922, opened 2025-10-08
Re-add SystemSetting resource processor

### LINKED ISSUE #850: Add <SystemSetting> Resource Processor back
The <SystemSetting> resource processor is available in v0.7.x releases but disappears from v0.9.x onwards. It is used to set the ^%SYS("CSP") global to a particular value during the Compile phase. For backwards compatibility, this should still be available in future IPM versions.

isc-dchui:
Resolves #850

### DIFF
--- a/src/cls/IPM/ResourceProcessor/SystemSetting.cls
+++ b/src/cls/IPM/ResourceProcessor/SystemSetting.cls
@@ -1,0 +1,35 @@
+Class %IPM.ResourceProcessor.SystemSetting Extends %IPM.ResourceProcessor.Abstract
+{
+
+/// Comma-separated list of resource attribute names that this processor uses
+Parameter ATTRIBUTES As STRING = "Name,Value";
+
+Property Name As %String(MAXLEN = "") [ Required ];
+
+Property Value As %String(MAXLEN = "");
+
+Method OnPhase(
+	pPhase As %String,
+	ByRef pParams,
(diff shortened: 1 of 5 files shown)


src/cls/IPM/ResourceProcessor/SystemSetting.cls
```diff
@@ -0,0 +1,35 @@
+Class %IPM.ResourceProcessor.SystemSetting Extends %IPM.ResourceProcessor.Abstract
```
isc-kiyer, line 1: Can a description be added to explain exactly what this does? Seems to be specific to some CSP settings? Feels like this should be deprecated in favor of the CPF resource processor (candidate for removal in v1.0.0).

src/cls/IPM/ResourceProcessor/SystemSetting.cls
```diff
@@ -0,0 +1,35 @@
...
+
+Property Value As %String(MAXLEN = "");
+
+Method OnPhase(
+	pPhase As %String,
+	ByRef pParams,
+	Output pResourceHandled As %Boolean = 0) As %Status
+{
+  if pPhase = "Compile" {
+    write !, "Resource Reference Name: ", ..ResourceReference.Name
```
isc-kiyer, line 17: Should this only write out in verbose mode?

isc-kiyer (changes requested):
@isc-dchui few comments/questions

PR #932, opened 2025-10-10
Adding new processor class for artifactory

isc-jlechtne:
This PR addresses #935 

This is for creating a new generic resource processor to cover bundling an artifact with a package and deploying it to a final location on install.

Behavior on packaging:
- Always attempts to download the artifact.
- Artifact untarred and moved to a location under the module root

Behavior on activation:
- Check if artifact files exist under the module root. If they don't, download and move them there (as in packaging)
- With untarred files under module root, move them to their final deployed location. This location is specified by `DeployDirectory` or `DevDirectory` if defined and installed in dev mode.

### DIFF
--- a/src/cls/IPM/ResourceProcessor/ArtifactoryTarball.cls
+++ b/src/cls/IPM/ResourceProcessor/ArtifactoryTarball.cls
@@ -154,7 +154,7 @@ Method DownloadArtifact(ByRef params) [ Internal, Private ]
 /// Errors if no valid artifact is found.<br />
 /// @Argument   artifactServer  Server name where the artifact is located.<br />
 /// @Argument   artifactPort    Port to use when connecting to artifactServer.<br />
-/// @Argument   aqlURL          URL endpoint for aql search withing JFrog Artifactory.<br />
+/// @Argument   aqlURL          URL endpoint for aql search within JFrog Artifactory.<br />
 /// @Argument   username        Username used for server authentication.<br />
 /// @Argument   password        Password used for server authentication.<br />
 /// @Argument   bearerToken     Token to use in place of username/password auth.<br />
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - #938 Added flag -export-python-deps to package command
 - #462: The `repo` command for repository configuration now supports secret input terminal mode for passwords with the `-password-stdin` flag
-- #935: Adding a generic JFrog Artifactory resource processor for bundling artifact with a package and deploying it to a final location on install.
+- #935: Adding a generic JFrog Artifactory tarball resource processor for bundling artifact with a package and deploying it to a final location on install.
 
 ### Removed
 - #938 Removed secret flag NewVersion handling in %Publish()


src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -134,107 +122,71 @@ Method DownloadArtifact(ByRef params) [ Internal, Private ]
...
-        // Query syntax error will be categorized as Bad Request (400) and displayed here
+        // Query syntax error will be a bad request and displayed here
         $$$ThrowStatus($$$ERROR($$$GeneralError, "The following error occurred while retrieving artifact info: " _ responseDataStr))
     }
 
     #dim jsonObj As %Library.DynamicAbstractObject
     set jsonObj = ##class(%Library.DynamicAbstractObject).%FromJSON(responseData)
     #dim results As %Library.DynamicArray
     set results = jsonObj.%Get("results")
+    set targetResultObj = results.%Get(0, "")
```
isc-dchui: So this will return the first result if there are multiple from the filter? If so, just thinking a bit about what the correct/best thing to do is. Maybe let the user know at least if not be able to select one?

isc-dchui (commented):
Just a few questions

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -0,0 +1,351 @@
...
+/// For more details on specific values, visit <a href="https://jfrog.com/help/r/jfrog-artifactory-documentation/artifactory-query-language">
+Property Filter As %String(MAXLEN = 2048);
+
+Method OnAfterPhase(phase As %String, ByRef params) As %Status
+{
+    #dim ex As %Exception.AbstractException
+    set sc = $$$OK
+    try {
+        if (phase = "Package") {
+            do ..DownloadArtifact(.params)
```
isc-kiyer: General q: once it gets downloaded into the target dir here, does IPM automatically copy it to the correct place for packaging? Also, we should consider how we want to add unit/integration tests here

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -0,0 +1,351 @@
...
+
+    set localPath = ##class(%Library.File).NormalizeFilename(artifactName, downloadDirectory)
+    if '##class(%Library.File).Exists(localPath) {
+        if verbose {
+            write !, "Update found. Saving to " _ localPath
+        }
+
+        set stream = ##class(%Library.FileBinaryStream).%New()
+        set stream.Filename = localPath
+        $$$ThrowOnError(stream.CopyFrom(responseData))
```
isc-kiyer: Nit: this could be CopyFromAndSave()

isc-kiyer (changes requested):
@isc-jlechtne few comments up but general logic looks good!

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -0,0 +1,351 @@
...
+        if (phase = "Package") {
+            do ..DownloadArtifact(.params)
+        }
+        if (phase = "Activate") {
+            set verbose  = $get(params("Verbose"),0)
+            // If install directory exists, use that
+            // Else, download from artifactory
+            set downloadDirectory = ##class(%File).NormalizeDirectory(..ResourceReference.Module.Root_..ResourceReference.Name)
+            if '##class(%File).DirectoryExists(downloadDirectory) {
+                if $get(params("Artifactory", "Run"), 0) {
```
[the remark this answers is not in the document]
isc-eneil: @isc-kiyer And just to clarify: the warning would explain that in order to pull the artifact, they'd need to set the appropriate variables and then uninstall and install -dev again?

src/cls/IPM/ResourceProcessor/Artifactory.cls
isc-jlechtne: @isc-kiyer Whenever an artifact gets downloaded, it is automatically copied to `..ResourceReference.Module.Root_..ResourceReference.Name`. I believe this is the correct path we want to use for packaging?

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -0,0 +1,351 @@
...
+    set developerMode = $get(params("DeveloperMode"), ..ResourceReference.Module.DeveloperMode)
+    set verbose  = $get(params("Verbose"),0)
+    set downloadDirectory = ##class(%File).NormalizeDirectory(..ResourceReference.Module.Root_..ResourceReference.Name)
+
+    if (verbose) {
+        write !, "Start retrieving artifact..."
+    }
+
+    merge artifactory = params("Artifactory")
+    set useHttps = $get(artifactory("UseHttps"), 0)
```
isc-jlechtne: @isc-tleavitt, @isc-dchui is there any reason why we should have a variable for UseHttps here vs always setting it to true for these artifactory calls?

src/cls/IPM/ResourceProcessor/Artifactory.cls
isc-kiyer: Yes I agree the behavior is correct. Its worth adding some logging in verbose mode though about what were the results that were found (like how many results were found) and the top one that was selected.
Also, worth writing out the full query (POST body) to the current device when in verbose mode.

isc-jlechtne: Am already writing out the body query (earlier above) but will add logging for multiple results found and which one gets selected

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -0,0 +1,366 @@
...
+Method OnExportItem(
+	fullExportPath As %String,
+	itemName As %String,
+	ByRef itemParams,
+	ByRef params,
+	Output itemHandled As %Boolean = 0) As %Status
+{
+    set sc = $$$OK
+    try {
+        if (params("cmd") = "package") {
```
isc-kiyer: No need for this check (actually you don't want it since this would not export if the publish command is run)

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -0,0 +1,364 @@
...
+	ByRef params,
+	Output itemHandled As %Boolean = 0) As %Status
+{
+    set sc = $$$OK
+    try {
+        do ..DownloadArtifact(.params)
+        set downloadDirectory = ##class(%File).NormalizeDirectory(..ResourceReference.Module.Root_..ResourceReference.Name)
+        set exportPath = params("ModuleExportPath")
+        // Copy contents over to package directory with pDeleteBeforeCopy = 1
+        do ##class(%File).CopyDir(downloadDirectory, exportPath,,1)
```
isc-jili: Question: is there a reason the rest of the CopyDir() calls in this class come from %Library.File and this one comes from %File?

src/cls/IPM/ResourceProcessor/Artifactory.cls
```diff
@@ -124,10 +124,10 @@ Method DownloadArtifact(ByRef params) [ Internal, Private ]
...
-    set PMTempDir = ##class(%Library.File).SubDirectoryName(OSTempDir, ..ResourceReference.Name, 1)
+    set PMTempDir = ##class(%File).SubDirectoryName(OSTempDir, ..ResourceReference.Name, 1)
 
-    set artifactPath = ..GetArtifactPath(useHttps, server, port, apiURL, useTokenAuth, username, password, bearerToken, ..Filter, verbose)
-    set artifactName = ..RetrieveArtifact(useHttps, server, port, artifactPath, PMTempDir, verbose)
+    set artifactPath = ..GetArtifactPath(server, port, aqlURL, username, password, bearerToken, ..Filter, verbose)
+    set artifactName = ..RetrieveArtifact(server, port, artifactPath, PMTempDir, verbose)
 
     // Unpack artifact in temp location (so that if anything goes wrong, the true destination does not get affected)
     // and copy artifact to destination
```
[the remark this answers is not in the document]
isc-dchui: That makes sense. So something like "ArtifactoryTarball"?

src/cls/IPM/ResourceProcessor/ArtifactoryTarball.cls
```diff
@@ -0,0 +1,289 @@
...
+        set verbose  = $get(params("Verbose"),0)
+        if (verbose) {
+            write !, "Going to copy extracted contents from "_downloadDirectory_" to "_exportPath
+        }
+        // Copy contents over to package directory with pDeleteBeforeCopy = 1
+        if '##class(%File).CopyDir(downloadDirectory, exportPath, 1, , 1) {
+            $$$ThrowStatus($$$ERROR($$$GeneralError, "Failed to copy to directory: "_exportPath))
+        }
+        set itemHandled = 1
+            if (verbose) {
```
isc-dchui: Nit: indentation

src/cls/IPM/Utils/FileBinaryTar.cls
```diff
@@ -160,17 +160,22 @@ Method ExtractTo(pDest As %String) As %Status
...
             set fullPath = ##class(%File).NormalizeFilename(..name, pDest)
             if $extract(fullPath, * - 7, *) = "IRIS.DAT" {
                 set irisdat = 1
                 set fullPath = $extract(fullPath, 1, * - 8) _ "iris.dat"
             }
-            set fs = ##class(%Stream.FileBinary).%New()
-            set fs.Filename = fullPath
-            do fs.CopyFromAndSave(..fileData)
-            $$$ThrowOnError(fs.LinkToFile(fullPath))
+            set fs = ##class(%Stream.FileBinary).%OpenId(fullPath)
```
isc-kiyer: Nit check %OpenId status and throw on error

src/cls/IPM/Utils/FileBinaryTar.cls
```diff
@@ -160,17 +160,22 @@ Method ExtractTo(pDest As %String) As %Status
                 return sc
             }
         } else {
+            // First create subdirectories for the path if they don't already exist
+            set path = pDest_..name
```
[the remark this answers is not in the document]
isc-jlechtne: @isc-dchui NormalizeFilename() returns an empty string if the directory doesn't exist which is the case here and why I went with creating the directory prior to the NormalizeFilename() call.

isc-dchui (approved):

isc-dchui (approved):

isc-kiyer (approved):

PR #972, opened 2025-11-26
Fix issue with FileCopy on directories breaking without leading slash

### LINKED ISSUE #965: FileCopy sneakily breaks if directory name doesn't start with "/"
Sample module.xml:

```
<?xml version="1.0" encoding="UTF-8"?>

  <Document name="fc-debug.ZPM">
    <Module>
      <Name>fc-debug</Name>
      <Version>1.0.0</Version>
      <Packaging>module</Packaging>
      <FileCopy Name="foo/bar/" Target="${libdir}foo/bar/"/>
    </Module>
  </Document>
</Export>
```

Call `load`, then `publish`, and then `install` will fail with
```
[fc-debug]      Activate FAILURE
ERROR! Error copying directory /usr/irissys/ipm/fc-debug/1.0.0/foo/bar/ to /usr/irissys/lib/foo/bar/
  > ERROR #5021: Directory '/usr/irissys/ipm/fc-debug/1.0.0/foo/bar/' does not exist.`
```

The problem is that the Package phase in `publish` changes the `foo` directory to `.foo`
```
[USER|fc-debug] Package START
Exported to /usr/irissys/mgr/Temp/dirj1zAaG/fc-debug-1.0.0/module.xml
Copying /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/fc-debug/foo/bar/ to /usr/irissys/mgr/Temp/dirj1zAaG/fc-debug-1.0.0/.foo/bar/ as directory
```

isc-dchui:
Fixes #965

### DIFF
--- a/tests/integration_tests/Test/PM/Integration/FileCopy.cls
+++ b/tests/integration_tests/Test/PM/Integration/FileCopy.cls
@@ -1,7 +1,7 @@
 Class Test.PM.Integration.FileCopy Extends Test.PM.Integration.Base
 {
 
-Method Setup() As %Status
+Method OnBeforeAllTests() As %Status
 {
     // Setup repo for FileCopy tests
     set status = ##class(%IPM.Main).Shell("repo -n integration-tests -fs -path /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/")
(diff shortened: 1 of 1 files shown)


tests/integration_tests/Test/PM/Integration/FileCopy.cls
```diff
@@ -149,4 +153,33 @@ Method TestFileCopyScope()
     }
 }
 
+Method TestFileCopyExport()
+{
+    set tSC = $$$OK
```
isc-kiyer: Nit: for net new methods no need to use the `t` prefix

PR #989, opened 2025-12-11
Make default and external name interchangeable for `install` and `update` commands for ORAS repos

### LINKED ISSUE #789: Revisit consistency across registry types for AllVersions in search criteria
Currently, the behavior of `ListModules` across registry types is a bit inconsistent. We have three use cases for ListModules - listing for dependency resolution (which gets everything matching a semver expression - possibly just everything), searching (which has the "all versions" option), and repo "-list-modules" (which doesn't have the "all versions" option and should just show the latest).

Ultimately the full set of versions should be aggregated by the caller (as semantic versions) rather than being part of the ListModules response.

### LINKED ISSUE #951: Allow `unpublish` command to skip user prompt if given `-force` flag
Some of the integration tests publish to a registry and it would be nice to be able to clean this up by unpublishing. Unfortunately `unpublish` currently requires a user to respond. This seems like the perfect case for adding a `-force` flag that skips this step. It should still respect the repo settings about whether unpublish is allowed.

Also there should be better documentation that `unpublish` accepts a repository name using the `unpublish <repository>/<module-name> <module-version>` syntax

### LINKED ISSUE #954: `update` with external name failed
I am trying to update a module, this module has `<Name>` and `<ExternalName>` in module.xml. I have installed the previous version of the module and confirmed it exists using `list-installed`, for example:
```
zpm:XYZ>list-installed
abc_external (abc) 1.0.0
```

When using external name <ExternalName>, for example:
`update abc_external 1.1.0 -path abc.tgz -verbose`
This failed with error message: 
`Module abc_external has not been installed. Please use the install or load commands to install the module before calling update.`

When using the original name <Name>, for example: 
`update abc 1.1.0 -path abc.tgz -verbose`
This does not throw error.

### LINKED ISSUE #959: ExternalName does not work with install
ExternalName and Name should be able to be used interchangeably for load, install, & update. At this time, ExternalName does work interchangeably for uninstall but not the other two. When trying to use ExternalName for install, you get the error `<ExternalName> not found in any repository`. The update command calls install, so the error install exists in update as well.

isc-dchui:
Resolves #959
Resolves #954 
Resolves #789 
Resolves #951 

Note: the interchangeability of default and external name is only implemented for ORAS repos, as the module.xml is in the metadata. For remote repos, it would require pulling the entire module, which would add a significant overhead, so has been left unimplemented.


src/cls/IPM/Storage/Module.cls
```diff
@@ -16,7 +16,7 @@ Property VersionString As %String(MAXLEN = 100, XMLNAME = "Version") [ InitialEx
 /// Does not need comparison method to be code generated because that comparing <property>VersionString</property> is good enough.
 Property Version As %IPM.General.SemanticVersion(ForceCodeGenerate = 0, XMLPROJECTION = "NONE") [ Required ];
 
-Property ExternalName As %String(MAXLEN = 255);
+Property ExternalName As %IPM.DataType.ModuleName;
```
isc-kiyer: Hmm I'm not sure we should do this. We should just validate if trying to use ext name to publish. This could be backwards incomptaible.

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -126,11 +126,73 @@ Method ListModulesFromTagString(
...
 
+        // Filter by module name if requested - check both Name and ExternalName
+        if (searchCriteria.Name '= "") {
+            // Extract the package name without namespace prefix for comparison
+            // Package name from ORAS may be "namespace/modulename" or just "modulename"
+            set packageNameWithoutNamespace = name
+            if (name [ "/") {
+                #; Extract everything after the last "/"
+                for i=$length(name):-1:1 {
+                    if ($extract(name, i) = "/") {
```
isc-kiyer: Does $Piece() not work here such that a loop is needed?

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -126,11 +126,73 @@ Method ListModulesFromTagString(
...
+                    }
+                    if 'found {
+                        continue
+                    }
+
+                    // Check if search name matches either Name or ExternalName (case-insensitive)
+                    set matchesName = 0
+                    set matchesExternalName = 0
+
+                    if ($$$lcase(moduleObj.Name) = $$$lcase(searchCriteria.Name)) {
```
isc-kiyer: Same with `matchesName` and `matchesExternalName` as above. Can make these single declarations.

src/cls/IPM/Repo/Oras/PackageService.cls
```diff
@@ -162,73 +224,68 @@ Method ListModulesFromTagString(
...
-        $$$ThrowOnError(tSC)
-        set response=request.HttpResponse
-        if (response.StatusCode'=200) {
-            // TODO improve error processing
-            set msg = "Error: ORAS namespace is not set and the call to /v2/_catalog endpoint failed"
-            set msg = msg _ $char(10,13) _ "Either set an ORAS namespace or ensure the ORAS server supports the /v2/_catalog endpoint"
-            set msg = msg _ $char(10,13) _ "Response Code: "_response.StatusCode _ " - " _ response.Data.Read()
-            $$$ThrowStatus($$$ERROR($$$GeneralError, msg))
-        }
+    // Use /v2/_catalog to enumerate all packages, then filter appropriately
```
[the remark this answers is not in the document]
isc-dchui: I've added some special handling to cover the case where /v2/_catalog is unavailable but name is specified. External name and default name won't be interchangeable in this case though

tests/integration_tests/Test/PM/Integration/PublishExternalName.cls
```diff
@@ -9,31 +9,247 @@ Parameter Folder = "publish-external-name";
...
+    do $$$AssertStatusOK(sc, "Deleted all repos")
 
-    Set sc = ##class(%IPM.Main).Shell("repo -o -name oras -url http://oras:5000 -publish 1")
-    Do $$$AssertStatusOK(sc, "Added ORAS repo")
+    set sc = ##class(%IPM.Main).Shell("repo -o -name oras -url http://oras:5000 -publish 1")
+    do $$$AssertStatusOK(sc, "Added ORAS repo")
 
-    Set sc = ##class(%IPM.Main).Shell($$$FormatText("load %1", moduleDir))
-    Do $$$AssertStatusOK(sc, "Loaded module")
+    set sc = ##class(%IPM.Main).Shell($$$FormatText("load %1", moduleDir _ "/0.0.1"))
```
isc-eneil: I think trying to install with external name should not be part of this-- I'd suggest adding that to the InstallModule.cls test suite to make the scope of the PublishExternalName() tests limited to only testing publish functionality

src/cls/IPM/Utils/Module.cls
```diff
@@ -328,6 +328,19 @@ ClassMethod GetModuleObjectFromStream(
     return $get(moduleObj)
 }
 
+/// Given a string, tries to correlate it to an instance of <class>%IPM.Storage.Module</class>.
+/// Throws errors.
```
isc-jili: I think it's great to comment explicitly on methods that throw errors. Would it be helpful for consistency to also comment on other methods on this PR that throw errors? For example ListModules() above throws errors but lacks this comment

src/cls/IPM/Main.cls
```diff
@@ -2154,6 +2154,8 @@ ClassMethod ShowModulesForRepository(
             set list(list, "AllVersions") = tRes.AllVersions
         }
     }
+    // Also need to throw an error if there are no results and tSC is set
```
isc-jili: This comment confused me a bit. Is it explaining what the ThrowOnError is doing? Or is it a placeholder for a TODO for future work?

tests/integration_tests/Test/PM/Integration/PublishExternalName.cls
```diff
@@ -9,31 +9,247 @@ Parameter Folder = "publish-external-name";
...
+    set sc = ##class(%IPM.Main).Shell("repo -delete-all")
+    do $$$AssertStatusOK(sc, "Deleted all repos")
 
-    Set sc = ##class(%IPM.Main).Shell("repo -reset-defaults")
-    Do $$$AssertStatusOK(sc, "Reset repos to default")
+    set sc = ##class(%IPM.Main).Shell("repo -reset-defaults")
+    do $$$AssertStatusOK(sc, "Reset repos to default")
+}
+
+Method TestInterchangeableNames()
```
isc-jili: Suggest adding a comment here saying interchangeable names refers to external name VS name so that it's clear for future folks to understand

tests/integration_tests/Test/PM/Integration/PublishExternalName.cls
isc-dchui: Ah good call. I've moved the tests.

src/cls/IPM/Main.cls
```diff
@@ -2506,23 +2508,26 @@ ClassMethod Unpublish(ByRef pCommandInfo) [ Internal ]
...
-                set tMsg = "Are you sure you want to delete all versions of the package """_tModuleName_""" from registry """_tServer.Name_""" ("_tServer.URL_")?"
-            } else {
-                write $$$FormattedLine($$$Red, "Deleting a package version is an irreversible action")
-                set tHelp = "Enter ""Yes"" if you want to delete selected package version."
-                set tMsg = "Are you sure you want to delete the package """_tModuleName_" "_tVersion_""" from registry """_tServer.Name_""" ("_tServer.URL_")?"
-            }
+            set force = $data(pCommandInfo("modifiers","force"))
+            if 'force {
+                if ($$$lcase(tVersion)="all") {
+                    write $$$FormattedLine($$$Red, "Deleting a package and all its versions is an irreversible action")
```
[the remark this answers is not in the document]
isc-dchui: I think the extra user confirmation dialog should be enough but I've added an exclamation point as well

PR #1016, opened 2025-12-31
Fix v0.10.5 release blockers

### LINKED ISSUE #999: Clean up stale ^oddStudioDocument nodes on IPM installation
e.g., for %ZHSLIB.PackageManager.* and %ZPM.PackageManager.*

These are preserved in IRISSYS on upgrade (e.g., across many healthshare versions, we're seeing both of these plus %IPM) even though the corresponding classes no longer exist.

### LINKED ISSUE #1000: Clean up workaround with oddStudioDocument mappings on upgrade
On older versions when upgrading IPM to ensure no studio document conflicts, we recommended a workaround for creating global mappings. These MUST be removed from all namespaces when upgrading IPM.
Came from a debugging session with Ready Computing on 12/23/2025 where they upgraded HS from 2023.1 to 2025.2 and IPM from 0.7.x to 0.10.4 AFTER having upgraded HS.

From https://github.com/intersystems/ipm/issues/433#issuecomment-1852799339:

Here's the minimally invasive workaround:

To enable use of the InterSystems Package Manager (from GitHub / the Open Exchange) in an InterSystems IRIS for Health™ 2023.3 or HealthShare® Health Connect 2023.3 interoperability-enabled namespace, run the following ObjectScript commands in that namespace:

do ##class(%ZPM.PackageManager.Developer.Utils).AddGlobalMapping($namespace,"oddStudioDocument:(BEGIN):(""%ZP"")","IRISLIB")
do ##class(%ZPM.PackageManager.Developer.Utils).AddGlobalMapping($namespace,"oddStudioDocument:(""%ZP""):(""A"")","IRISSYS")
do ##class(%ZPM.PackageManager.Developer.Utils).AddGlobalMapping($namespace,"oddStudioDocument:(""Ens""):(""Ent"")","ENSLIB")
This is equivalent to the following iris.cpf changes – for example, given namespace IPMTEST:

[Map.IPMTEST]
Global_oddStudioDocument=IPMTEST
Global_oddStudioDocument(BEGIN):("%ZP")=IRISLIB
Global_oddStudioDocument("%ZP"):("A")=IRISSYS
Global_oddStudioDocument("Ens"):("Ent")=ENSLIB
These mappings should be removed after the next upgrade; the issue is expected to be resolved in the next release.

### LINKED ISSUE #1007: <Arg> element doesn't understand variables, e.g. ${ipmdir}
Tried to transfer the path where I installed data into the Init method to deal with installed files, but it stopped working. 
Here is my setup:
```xml
<FileCopy Name="data/" Target="${ipmdir}data/"/>
       <Invokes>
        <Invoke Class="app.setup" Method="Init">
        <Arg>${ipmdir}</Arg>
        </Invoke>
      </Invokes>
```
it comes as "{ipmdir}" string to the Init() method.
It used to work previously.

### LINKED ISSUE #1015: Sneaky dependency resolution regressions
1) Dependencies with version requirements of `*` are not being handled correctly.
2) The upper bounds of ranges cannot be directly compared using "Follows" semantics. For example, `1.*` resolves to `>=1.0.0 <2.0.0-0` and `1.0.*` resolves to `>=1.0.0 <1.1.0-0` (see https://github.com/npm/node-semver?tab=readme-ov-file#advanced-range-syntax), but `2.0.0-0` and `1.0.0-0` cannot be directly compared with typical precedence rules (see https://github.com/npm/node-semver?tab=readme-ov-file#prerelease-tags). Instead, they must be numerically compared part by part.

### LINKED ISSUE #1036: `Update` propagates dev mode to dependencies when it shouldn't
When calling the `update` command with the `-dev` flag, it should behave like the `install` command where the parent module is installed in developer mode but the dependencies are not.

isc-dchui:
Fixes #999 
Fixes #1000 
Fixes #1007 
Fixes #1015 
Fixes #1036

### DIFF
--- a/preload/cls/IPM/Installer.cls
+++ b/preload/cls/IPM/Installer.cls
@@ -23,6 +23,7 @@ XData PM [ XMLNamespace = INSTALLER ]
     <ClassMapping Package="%IPM" From="${CURNSRoutineDB}" />
     <RoutineMapping Routines="%IPM.*" Type="ALL" From="${CURNSRoutineDB}" />
   </Configuration>
+  <Invoke Class="IPM.Installer" Method="CleanupOldMappings"/>
   <Import File="${SRCDIR}" Recurse="1"/>
   <Invoke Class="%IPM.Main" Method="UpdateLanguageExtensions" CheckStatus="true" />
   <Invoke Class="IPM.Installer" Method="ZPMInit" CheckStatus="true">
@@ -416,8 +417,7 @@ ClassMethod CreateTempDir() As %String
     return ##class(%File).NormalizeDirectory(tDir)
 }
 
-/// This method is called directly in LoadNewModule() instead of with an <Invoke> in the module.xml
-/// because the old mappings break load/install before any <Invokes> can be run.
+/// When installing a new version of IPM, make sure any old mappings are cleaned up
 ClassMethod CleanupOldMappings() As %Status
 {
     set sc = $$$OK
@@ -465,6 +465,11 @@ ClassMethod CleanupOldMappings() As %Status
         set status = ##class(%IPM.Utils.Module).RemoveGlobalMapping($namespace,"oddStudioDocument:(""Ens""):(""Ent"")","ENSLIB")
         if $$$ISERR(status) {
             set sc = $$$ADDSC(sc, status)
+        }
+        set status = ##class(%IPM.Utils.Module).RemoveGlobalMapping($namespace,"oddStudioDocument",namespace)
+        if $$$ISERR(status) {
+            // this mapping deletion will fail if there are still sub-mappings present
+            // so we can ignore errors here
         }
     }
 
--- a/src/cls/IPM/Lifecycle/Base.cls
+++ b/src/cls/IPM/Lifecycle/Base.cls
@@ -1322,7 +1322,6 @@ Method %Verify(ByRef pParams) As %Status
 
                 // Map the package manager itself equivalently in the VERIFY namespace.
                 $$$ThrowOnError(##class(%IPM.Utils.Build).MapRoutineEquivalently("%IPM.*",tInitNS,,tNewNamespace))
-                $$$ThrowOnError(##class(%IPM.Utils.Build).MapRoutineEquivalently("IPM.*",tInitNS,,tNewNamespace))
                 $$$ThrowOnError(##class(%IPM.Utils.Build).MapPackageEquivalently("%IPM",tInitNS,tNewNamespace))
 
                 // Map globals for %IPM.Repo.Definition and %IPM.Repo.Filesystem.Cache
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -1113,11 +1113,6 @@ ClassMethod LoadNewModule(
             set toVersionString = newModuleObj.VersionString
             // Cannot simply get name from pParams because load command doesn't take in module name
             set moduleCurrent = ##class(%IPM.Storage.Module).NameOpen(newModuleObj.Name)
-
-            // If installing new version of IPM, clean up old mappings
-            if newModuleObj.Name=$$$IPMModuleName {
-                $$$ThrowOnError(##class(IPM.Installer).CleanupOldMappings())
-            }
         }
         if moduleCurrent '= "" {
             set fromVersion = moduleCurrent.Version
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ lock contention by bypassing IRIS compiler.
 - #999, #1000: Installing IPM cleans up stale mappings used in old versions of IPM
 - #1007: The `${ipmDir}` expression now works in the `<Arg>` of an `<Invoke>`
 - #1015: Fix dependency resolution bugs where `*` as the version requirement and intersecting ranges wouldn't work properly.
+- #1036: The `update` command no longer propagates developer mode to dependencies
 
 ### Deprecated
 - #828: The `CheckStatus` flag for `<Invoke>` action has been deprecated. Default behavior is now to always check the status of the method if and only if the method signature returns %Library.Status


src/cls/IPM/Storage/Module.cls
```diff
@@ -1774,6 +1777,7 @@ Method %Evaluate(
     set customParams("packagename") = ..Name
     set customParams("version") = ..VersionString
     set customParams("verbose") = +$get(pParams("Verbose"))
+    set customParams("ipmDir") = ##class(%Library.File).NormalizeDirectory($system.Util.DataDirectory() _ "ipm/" _ ..Name _ "/" _ ..VersionString)
```
isc-kiyer: Be sure to add this to the wiki indicating what version this variable became available!

src/cls/IPM/Utils/Module.cls
```diff
@@ -274,17 +274,20 @@ ClassMethod LoadModuleFromDirectory(
...
-                zkill tParams("Install")
-            }
-        } else {
-            // Propagate all params except developer mode to dependencies
-            set tParams("DeveloperMode") = $get(pOverrideDeveloperMode,0)
+        if '$get(pParams("Install")) {
+            set tParams("DeveloperMode") = pOverrideDeveloperMode
+        }
+
+        if $get(pParams("Update")) {
```
isc-kiyer: The behavior for install and update should be identical. Also, we need to ensure IgnoreInstalled and UpdateSnapshots are also not passed to dependencies (in addition to DeveloperMode).
It would be good to have a complex dependency test case which currently would result in infinitely installing dependencies if IgnoreInstalled is passed as 1.

isc-kiyer (changes requested):
@isc-dchui few minor comments. Looks good!

src/cls/IPM/Utils/Module.cls
isc-dchui: This fix actually realigns Install and Update to both avoid propagating dev mode to dependencies. In the code flow, `pParams("Update")` is set during the Update command, and `pParams("Install")` during the Install command if the Update flag is not set.

The test for install is Test.PM.Integration.LoadModuleWithDeps.TestParamPropagation and the one for update is Test.PM.Integration.Update.TestDevModePropagation


I believe @isc-jili is tackling the IgnoreInstalled and UpdateSnapshots issues

preload/cls/IPM/Installer.cls
```diff
@@ -416,4 +416,37 @@ ClassMethod CreateTempDir() As %String
     return ##class(%File).NormalizeDirectory(tDir)
 }
 
+/// This method is called directly in LoadNewModule() instead of with an <Invoke> in the module.xml
+/// because the old mappings break load/install before any <Invokes> can be run.
+ClassMethod CleanupOldMappings() As %Status
```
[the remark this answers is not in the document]
isc-dchui: I tried that and it didn't work. For details, I set up the old mappings using
```
do ##class(%IPM.Utils.Module).AddGlobalMapping($namespace,"oddStudioDocument:(BEGIN):(""%ZP"")","IRISLIB")
do ##class(%IPM.Utils.Module).AddGlobalMapping($namespace,"oddStudioDocument:(""%ZP""):(""A"")","IRISSYS")
do ##class(%IPM.Utils.Module).AddGlobalMapping($namespace,"oddStudioDocument:(""Ens""):(""Ent"")","ENSLIB")
```
then called `load /home/irisowner/zpm` to install IPM again, which resulted in this error:
```
ERROR! User defined document 'ZPM.ZPM' not supported. No user defined document class in this namespace.
  > ERROR #6315: Errors reporting importing XML subelement in file '/home/irisowner/zpm/module.xml' at line '3' offset '28', skipping this item.
```

It seems the problem is that the mappings break IPM before any of the lifecycle stages can happen, which is when an `<Invoke>` would occur.

Good call on checking the other namespaces though, I've added that change.

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1323,6 +1322,7 @@ Method %Verify(ByRef pParams) As %Status
 
                 // Map the package manager itself equivalently in the VERIFY namespace.
                 $$$ThrowOnError(##class(%IPM.Utils.Build).MapRoutineEquivalently("%IPM.*",tInitNS,,tNewNamespace))
+                $$$ThrowOnError(##class(%IPM.Utils.Build).MapRoutineEquivalently("IPM.*",tInitNS,,tNewNamespace))
```
isc-eneil: Why is this necessary @isc-dchui ?

isc-eneil (commented):
@isc-dchui Looks good to me! Just one suggestion and one question.

preload/cls/IPM/Installer.cls
```diff
@@ -416,4 +416,59 @@ ClassMethod CreateTempDir() As %String
...
+        for {
+            set key = $order(^oddStudioDocument(key))
+            quit:key=""
+            if (key["%ZPM.PackageManager") || (key["%ZHSLIB.PackageManager") {
+                kill ^oddStudioDocument(key)
+            }
+        }
+
+        // These mappings were suggested as a temporary workaround for IRIS 2023.x,
+        // but whose underlying issues have been resolved in 2024.1 and later.
```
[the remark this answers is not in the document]
isc-dchui: We might miss the case where a customer has these mappings, upgrades to a new version of IRIS first, and then tries to upgrade IPM.

preload/cls/IPM/Installer.cls
```diff
@@ -416,4 +417,63 @@ ClassMethod CreateTempDir() As %String
...
+    // Iterate through each namespace and remove old mappings
+    set namespace = ""
+    for {
+        set namespace = $order(nsList(namespace))
+        if namespace = "" {
+            quit
+        }
+        set $namespace = namespace
+        // check if IPM is installed in this namespace
+        if '$$$comClassDefined("%IPM.Utils.Module") {
```
isc-kiyer: Nit: I would check for %IPM.Main in general since its the entry point at part of the public API of IPM so is unlikely to change

isc-kiyer (approved):

Issue #1017, opened 2026-01-01
uninstall command executes on implicit context when argument is missing


AshokThangavel:
The `uninstall` command does not strictly enforce the requirement for a module name argument or the `-all` flag, despite what is stated in the `help` documentation. Instead, it appears to target modules based on an internal "contextual stack" (likely the most recently listed or loaded modules). This leads to accidental uninstallation of applications when a user executes the command without parameters.

```objectscript
zpm:LEARNING>list
IPM (zpm)                      0.10.4
webfslog (web-fslog)           1.0.2
embeddedpy-bridge              1.0.1
DocDBUI (docdb-ui)             1.0.4
FHIRServerLogs (irisfsrepolog) 1.0.3
testify                        1.0.2
interop-lookuptable            1.0.2
zpm:LEARNING>uninstall
 
[LEARNING|Testify]      Clean START
[LEARNING|Testify]      Unconfigure START
[LEARNING|Testify]      Unconfigure SUCCESS
[LEARNING|Testify]      Clean SUCCESS
zpm:LEARNING>uninstall
 
[LEARNING|Interop-LookupTable]  Clean START
[LEARNING|Interop-LookupTable]  Unconfigure START
[LEARNING|Interop-LookupTable]  Unconfigure SUCCESS
[LEARNING|Interop-LookupTable]  Clean SUCCESS
zpm:LEARNING>uninstall
 
ERROR! Module '' not found.
```
### **Actual Behavior**

* IPM automatically selects a module (e.g., `testify`) and begins the uninstallation process:
* Running `uninstall` again continues to "pop" modules off the list until the stack is empty, at which point it finally throws `ERROR! Module '' not found.`

### **Expected Behavior**

According to the `help uninstall` documentation:

> `uninstall [flags] [<module>]`
> ○ **module**: Name of module to uninstall

The command should fail with an error such as **"ERROR: Module name required"** unless the `-all` flag is explicitly provided. It should not guess which module to delete based on the previous command's output.

PR #1018, opened 2026-01-02
fix(cli): require module name for uninstall

AshokThangavel:
### Description

This PR resolves an issue where the `uninstall` command would proceed without a module name. In such cases, the command would default to the "implicit context" (the last active module), leading to accidental uninstallation of applications.

fixes: #1017

### Verification Results

**Unit Test Passed:** `TestUninstallWithoutModuleName`

1. Verified that `uninstall` (no args) now returns a `NotOK` status.
2. Verified that `uninstall <name>` still functions correctly.


src/cls/IPM/Main.cls
```diff
@@ -2448,7 +2448,10 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
         $$$ThrowOnError(##class(%IPM.Utils.Module).UninstallAll(tForce,.tParams))
         return
     } else {
-        set tModuleName = pCommandInfo("parameters","module")
+        set tModuleName = $get(pCommandInfo("parameters","module"))
+        if (tModuleName = "") {
+            $$$ThrowOnError($$$ERROR($$$GeneralError, "Module name is required for uninstall."))
```
isc-kiyer: Nit: add to the message "when not using the -all flag"

tests/unit_tests/Test/PM/Unit/CLI.cls
```diff
@@ -313,4 +313,17 @@ Method TestListPython() As %Status
     quit sc
 }
 
+Method TestUninstallWithoutModuleName()
+{
+    do $$$LogMessage("Run Uninstall command with out module name")
+    do ..RunCommand("unsintall")
```
isc-kiyer: This should check appropriate error was thrown

isc-kiyer (changes requested):
@AshokThangavel Looks good! Few minor comments. Also, this needs a changelog update

src/cls/IPM/Main.cls
isc-dchui: Adding it to the error message itself would also be handy (and I think what @isc-kiyer meant)

tests/unit_tests/Test/PM/Unit/CLI.cls
```diff
@@ -219,10 +219,13 @@ Method CompareModifiers(
 	}
 }
 
-Method RunCommand(pCommand As %String)
+// This method returns the %Status for checking the success or failure of executed commands.
```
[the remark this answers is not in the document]
AshokThangavel: I've updated to three slashes

src/cls/IPM/Main.cls
AshokThangavel: Thank you. I've updated the error message to be more descriptive: `A module name must be specified for uninstall unless the -all flag is used`

PR #1021, opened 2026-01-05
Publishing a module with python dependencies

isc-jlechtne:
Addresses #1024 as well as adding the `-export-python-deps` flag to the "publish" command

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - #996: Ensure COS commands execute in exec under a dedicated, isolated context
+- #1002: When listing configured repositories, only show the TokenAuthMethod when a token is defined.
 - #1024: Modules with PythonWheels have wheels packaged correctly under -export-python-deps
 
 ## [0.10.5] - 2026-01-15
--- a/src/cls/IPM/Repo/Http/Definition.cls
+++ b/src/cls/IPM/Repo/Http/Definition.cls
@@ -90,8 +90,6 @@ Method Display()
     }
     if (..Token '= "") {
         write !,$char(9),"Token: ",..Padding(0),$case(..Token,"":"<unset>",:"<set>")
-    }
-    if (..Token '= "") || (..Password '= "") || (..Username '= "") {
         write !,$char(9),"Token Auth Method: ",..Padding(2),..TokenAuthMethod
     }
 }


isc-eneil (dismissed):

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1242,13 +1248,25 @@ Method %Export(
 }
 
 /// Downloads Python wheels for Python dependencies into the /root/wheels directory of the module. Also updates module.xml with the Python Wheel resources.
-Method ExportPythonDependencies(ByRef pParams)
+Method ExportPythonDependencies(
+	module As %IPM.Storage.Module,
```
isc-jlechtne: Actually don't need to update this to pass along the module parameter. Instead of calling `do ..ExportPythonDependencies(pyDepMod, .pParams)` in %Export(), can call `do pyDepMod.ExportPythonDependencies(.pParams)`. Then can still reference values in this method via ..Module, also makes more sense in the idea of it being an instance method (this new implementation acts more as a ClassMethod)

src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -1198,11 +1194,18 @@ Method %Export(
         set tParams($$$DeployedProjectIndex) = tDeployedProject
 
         set tModuleName = ""
+        set exportPythonDependencies = $get(pParams("ExportPythonDependencies"), 0)
         for {
             set tModuleName = $order(tResourceArray(tModuleName))
             if (tModuleName = "") {
                 quit
             }
+            if exportPythonDependencies {
```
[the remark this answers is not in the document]
isc-jlechtne: @isc-eneil tried this out and it caused the tests to fail so am going to keep this as is.

tests/integration_tests/Test/PM/Integration/ProcessPythonWheel.cls
```diff
@@ -190,25 +192,83 @@ Method TestPackageWithPythonDeps() As %Status
...
     // Cleanup
     set sc = ##class(%IPM.Main).Shell("uninstall package-with-python-deps")
     do $$$AssertStatusOK(sc, "Uninstalled module")
-    */
+
+    set sc = ##class(%IPM.Main).Shell("repo -delete -n localrepo")
+    do $$$AssertStatusOK(sc, "Deleted local filesystem repo")
+}
+
+/// Omitting -export-python-deps flag should package without the python
```
isc-eneil: Nit: "without the python wheel resources"

isc-dchui (approved):

isc-jili (approved):

PR #1027, opened 2026-01-08
feat: add -bypass-deps (-bd) flag

### LINKED ISSUE #1026: Add -bypass-deps flag to skip ObjectScript dependency resolution
**Description:**
Currently, the `load` , `install` and `update` commands automatically resolve and install all dependencies listed in a module's `module.xml`. While this ensures environment integrity, there are several "Expert Mode" scenarios where a user needs to skip this step.

**Precedent:**
This enhancement follows the pattern already established by the `-bypass-py-deps` flag (which skips Python dependencies). Adding `-bypass-deps` provides a consistent experience for ObjectScript dependencies.

**Example Usage:**

```objectscript
zpm:USER>load /path/to/module -bypass-deps

```

AshokThangavel:
## Description
This PR introduces a new flag, `-bypass-deps`, for the `load`, `install` and `update` commands. This allows users to skip the IPM dependency in installation phase. This is the ObjectScript equivalent of the existing `-bypass-py-deps` flag. and `-bpd` alias for skip python dependency.

fixed: #1026

## Changes

* **Command Definitions:** Added the `bypass-deps` modifier to the `load` and `install` commands in the XML metadata.
* **Lifecycle Logic:** Modified the dependency resolution logic to check for the `BypassDeps` parameter. If set, the system logs the bypass and skips the graph-building and resolution phases.
* **Consistency:** Aligned naming conventions with `-bypass-py-deps` to ensure a predictable CLI experience.
## Testing Performed

1. **Baseline Test:** Ran `load <path>` without the flag. Verified all dependencies (e.g., `irisjwt`) were initialized and activated.
2. **Bypass Test:** Ran `load <path> -bypass-deps`. Verified the "Building dependency graph" was skipped and only the target module was reloaded.


isc-dchui (commented):
Looks like the tests are failing on GitHub CI:
```
FAILED Test.PM.Unit.CLI:TestByPassIPMDepsModifier: AssertStatusOK - Module demo-module1 is loaded without -bypass-deps => ERROR #5001: Could not find satisfactory version of objectscript-math in any repositories. Required by: demo-module1 1.0.0: 0.0.5
FAILED Test.PM.Unit.CLI:TestByPassIPMDepsModifier: AssertTrue - Dependency module 'objectscript-math' should be installed when loading without -bypass-deps.
FAILED Test.PM.Unit.CLI:TestByPassIPMDepsModifier: AssertTrue - Dependency module 'LocalizationLab' should be installed when loading without -bypass-deps.
FAILED Test.PM.Unit.CLI:TestByPassIPMDepsModifier: AssertTrue - Dependency module 'irisjwt' should be installed when loading without -bypass-deps.
```

I'd say typically it's easier to just build your own modules and dependencies for integration tests like these. See `tests\integration_tests\Test\PM\Integration\_data\uninstall` for an example

AshokThangavel:
Hi @isc-dchui
I agree—if a user skips dependencies that aren't already present, the code will indeed fail at runtime. However, this flag is intended as an 'Expert Mode' for scenarios where the resolution phase itself is the blocker:

* **Development Speed:** During active local development where dependencies are already loaded, skipping the graph-building and resolution phase significantly speeds up the `load/install` cycle.
* **Consistency:** This ensures the ObjectScript side of the house has the same flexibility already provided by the `-bypass-py-deps` flag.

To ensure users are aware of the risk, I've added a log message: `"Skipping installation of IPM dependencies because -bypass-deps is set."`

PR #1050, opened 2026-01-23
Add `update -force-dependencies` flag

### LINKED ISSUE #430: ModuleDependencies and Version
I'm finding the overral reliability of ModuleDependencies, Versions to be unreliable.

I have tried using <Version>*</Version> as a way of indicating that when I install Module A which is dependent on Module B with Version *.  My expectation is that when I install Module A it will 
1. check to see if Module B is installed.  If it is not installed ZPM will install Module B.  This part works,
2. Check if Module B is install, if it is but the remote registry has a higher version then I feel like it should install the latest version of Module B.  This part is not working
3. If I defined in Module A a dependency on Version 2, even though I have Module B Version 2 installed but it is on Version 1 I run into issues like

Could not find satisfactory version of rclib-rcipm in any repositories. Required by: rclib-wellbase: 2.1.0; rclib-ccdprofile, rclib-arcgis, rclib-logviewer, rclib-hs: *; rclib-fhir: 1.3.2

isc-dchui:
Resolves #430 

This is an initial plan for how to address the question of updating complex dependencies that the current system fails to handle. See the UPDATE_FORCE_FLAG_PLAN.md file for details and the included example modules for the problem structure.

### DIFF
--- a/IMPLEMENTATION_CHECKLIST.md
+++ b/IMPLEMENTATION_CHECKLIST.md
@@ -1,0 +1,164 @@
+# Implementation Checklist: `-force-dependencies` Flag
+
+## Phase 1: Core Flag Support
+
+### 1.1 Command Definition
+- [ ] Add `-force-dependencies` modifier (alias `-force-deps`) to `update` command in Main.cls XData (line ~157)
+  - Modifier: `force-dependencies` with alias `force-deps`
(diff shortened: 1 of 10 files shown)


isc-dchui:
Superseded by #1136.

PR #1080, opened 2026-02-27
Adding zpm "ci" command to install from a lock file

isc-jlechtne:


src/cls/IPM/General/LockFile.cls
```diff
@@ -116,4 +116,88 @@ ClassMethod GetRepo(repoName As %String) As %IPM.Repo.Definition [ Internal ]
...
+    // Install repositories (if they don't already exist)
+    set repoIter = repositories.%GetIterator()
+    while repoIter.%GetNext(.repoName, .repoVals) {
+        if ##class(%IPM.Repo.Definition).ServerDefinitionKeyExists(repoName) {
+            if (verbose) {
+                write !, "Repo: "_repoName_" already exists, skipping creating new one from lock file"
+            }
+            continue
+        }
+        elseif (verbose) {
```
isc-dchui: Nit: change to one line like `} elseif (verbose) {`

src/cls/IPM/Main.cls
```diff
@@ -1031,8 +1050,10 @@ ClassMethod ShellInternal(
                 do ..ModuleVersion(.tCommandInfo)
             } elseif (tCommandInfo = "information") {
                 do ..Information(.tCommandInfo)
-			} elseif (tCommandInfo = "history") {
-				do ..History(.tCommandInfo)
+            } elseif (tCommandInfo = "history") {
+                do ..History(.tCommandInfo)
+            } elseif (tCommandInfo = "ci") {
+                do ..CleanInstall(.tCommandInfo)
```
isc-dchui: It's a bit confusing to me to have "ci" stand for "CleanInstall", since 1) CI=continuous integration, and 2) it doesn't mention the lock file like "LockFileInstall" would for example.

But taking a step back, what's the benefit of having a new command versus using a flag on `install`/`load`?

src/cls/IPM/Main.cls
```diff
@@ -2415,10 +2435,32 @@ ClassMethod Install(
...
+    }
+    $$$ThrowOnError(log.Finalize($$$OK, devMode))
+}
+
+ClassMethod CleanInstall(ByRef commandInfo) [ Internal ]
+{
+    set moduleName = $get(commandInfo("parameters","module"))
+    set version = $get(commandInfo("parameters","version"))
+    set verbose = $get(commandInfo("data","Verbose"))
+    set log = ##class(%IPM.General.HistoryTemp).CleanInstallInit(moduleName)
```
isc-dchui: Also add a test to check that the history log is being populated correctly (see Test.PM.Integration.History)

isc-dchui (changes requested):
Good overall! Just a few small things. Also you'll want to add a description.

src/cls/IPM/Main.cls
isc-jlechtne: The initial plan was to have creation of a lock file be tied to install/load but installing from it its own method. However, the more I think about this, I do think you are correct. Functionally, I don't think there is anything that requires a whole new method as opposed to the new flag you suggested. I'll make the switch for the next PR iteration (and we can always back track if a reason comes up to).

As for the naming part: The reason we called it ci/CleanInstall for consistency since that is the same name used in other package managers.

src/cls/IPM/Main.cls
```diff
@@ -2415,10 +2435,32 @@ ClassMethod Install(
...
+
+    // TODO: Add "path"? (see Update() for more info of calling install v load)
```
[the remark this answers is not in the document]
isc-jlechtne: It is being done. This was just a reminder to myself when putting up the first commit and PR for comments that I still need to account for that case.

src/cls/IPM/Main.cls
isc-jlechtne: Update: This has since been resolved by switching from a new "ci" command to a "-from-lockfile" flag on "install"

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -167,19 +162,23 @@ Method Test05ModuleDependenciesAllRepositoryTypes()
...
+        // Now that modules have been published to their respective repositories, run the test
+        do ..AssertInstallCreatesLockFileAsExpected(..#ModuleG)
+
+        // Remove all repositories except lock-file-edge (where Module G comes from) to confirm all types get created correctly when installing from lock file
+        do ##class(%IPM.Main).Shell("repo -delete -name lock-file-base")
+        do ##class(%IPM.Main).Shell("repo -delete -name "_remoteRepo)
+        do ##class(%IPM.Main).Shell("repo -delete -name "_orasRepo)
+
+        // Install from Module G's lock file
+        do ..AssertInstallFromLockFileAsExpected(..#ModuleG)
```
isc-jlechtne, line 175: One thing I noticed here is that the creation of the remote and oras repositories do not include username/password values as environment variables are not set. The repositories and test do still seem to pass but does anyone know if there's a way to set them in this test or should we add them as variables to the docker-compose.yml file?

src/cls/IPM/General/LockFile.cls
```diff
@@ -116,4 +116,93 @@ ClassMethod GetRepo(repoName As %String) As %IPM.Repo.Definition [ Internal ]
...
+        set version = depVals.version
+        set repository = depVals.repository
+
+        // Call Install() on dependency modules but set flag "LockFileInstallStarted" so we don't try installing from the dependency module's lock file
+        set commandInfo = "install"
+        set commandInfo("data", "Verbose") = verbose
+        set commandInfo("parameters","module") = repository_"/"_depName
+        set commandInfo("parameters", "version") = version
+        set commandInfo("data", "LockFileInstallStarted") = 1
+        set commandInfo("data","FromLockFile") = 1
```
isc-dchui: Do we need both "LockFileInstallStarted" and "FromLockFile"?

src/cls/IPM/Repo/Filesystem/Definition.cls
```diff
@@ -357,11 +357,21 @@ ClassMethod ScanDirectory(
...
+    set modifiers("path") = lockFileValues.%Get("root")
+    set modifiers("depth") = lockFileValues.%Get("depth")
+    set modifiers("read-only") = lockFileValues.%Get("readOnly")
+}
+
 XData LockFileMapping
 {
 <Mapping xmlns="http://www.intersystems.com/jsonmapping">
   <Property Name="LockFileType" FieldName="type" />
-  <Property Name="OverriddenSortOrder" FieldName="overriddenSortOrder" />
```
isc-dchui, line 364: Out of curiosity, what's the reason we're removing this?

src/cls/IPM/Repo/Oras/Definition.cls
```diff
@@ -154,11 +154,34 @@ Method GetPublishingManager(ByRef status)
...
+	lockFileValues As %DynamicObject,
+	Output modifiers)
+{
+    set modifiers("oras") = ""
+    set modifiers("name") = lockFileValues.%Get("name")
+    set modifiers("read-only") = lockFileValues.%Get("readOnly")
+    set modifiers("url") = lockFileValues.%Get("url")
+    set modifiers("namespace") = lockFileValues.%Get("orasNamespace")
+
+    // The following variables are set as system level variables for us to get
```
isc-dchui, line 167: This should end up being documented somewhere else too. Definitely in the wiki and maybe even in the help text from `zpm "help"`

tests/integration_tests/Test/PM/Integration/LockFile.cls
isc-dchui: I believe the ORAS repo doesn't require username/password and the remote one is being configured with it (see line 142)

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -310,13 +414,35 @@ ClassMethod AreLockFilesEqual(
 	expectedLockFilePath As %String) As %Boolean
```
isc-dchui, line 421: Should be `As %Status`

src/cls/IPM/General/LockFile.cls
```diff
@@ -116,4 +116,93 @@ ClassMethod GetRepo(repoName As %String) As %IPM.Repo.Definition [ Internal ]
...
+                write !, "Repo: "_repoName_" already exists, skipping creating new one from lock file"
+            }
+            continue
+        } elseif (verbose) {
+            write !, "Creating repo: "_repoName_" from lock file"
+        }
+        do ##class(%IPM.Repo.Definition).CollectServerTypes(.types)
+        set repoClass = types(repoVals.type)
+        set repoVals.name = repoName
+        do $classmethod(repoClass, "LockFileValuesToModifiers", repoVals, .modifiers)
```
isc-dchui, line 154: Should kill modifiers between iterations of the loop so they don't leak between repos

src/cls/IPM/General/LockFile.cls
```diff
@@ -116,4 +116,93 @@ ClassMethod GetRepo(repoName As %String) As %IPM.Repo.Definition [ Internal ]
...
+        if ##class(%IPM.Repo.Definition).ServerDefinitionKeyExists(repoName) {
+            if (verbose) {
+                write !, "Repo: "_repoName_" already exists, skipping creating new one from lock file"
+            }
+            continue
+        } elseif (verbose) {
+            write !, "Creating repo: "_repoName_" from lock file"
+        }
+        do ##class(%IPM.Repo.Definition).CollectServerTypes(.types)
+        set repoClass = types(repoVals.type)
```
isc-dchui: Should handle the error thrown if `repoVals.type` is not in `types`

src/cls/IPM/General/LockFile.cls
```diff
@@ -116,4 +116,93 @@ ClassMethod GetRepo(repoName As %String) As %IPM.Repo.Definition [ Internal ]
...
+    while repoIter.%GetNext(.repoName, .repoVals) {
+        if ##class(%IPM.Repo.Definition).ServerDefinitionKeyExists(repoName) {
+            if (verbose) {
+                write !, "Repo: "_repoName_" already exists, skipping creating new one from lock file"
+            }
+            continue
+        } elseif (verbose) {
+            write !, "Creating repo: "_repoName_" from lock file"
+        }
+        do ##class(%IPM.Repo.Definition).CollectServerTypes(.types)
```
isc-dchui: Doesn't need to be inside this while loop

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -167,19 +162,23 @@ Method Test05ModuleDependenciesAllRepositoryTypes()
...
-        do ##class(%IPM.Main).Shell("uninstall -r "_moduleName)
-        do $$$AssertStatusOK(sc, "Uninstalled "_moduleName_" and dependencies at the end of the test")
         do ##class(%IPM.Main).Shell("repo -delete -name "_remoteRepo)
         do $$$AssertStatusOK(sc, "Removed "_remoteRepo_" at the end of the test")
         do ##class(%IPM.Main).Shell("repo -delete -name "_orasRepo)
         do $$$AssertStatusOK(sc, "Removed "_orasRepo_" at the end of the test")
-    } catch (ex) {
-        set sc = ex.AsStatus()
+    } catch e {
+        do $$$AssertStatusOK(e.AsStatus(),"An exception occurred in Test05ModuleDependenciesAllRepositoryTypes.")
```
isc-dchui: Update this error message with the new method name

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -248,43 +269,126 @@ Method Test07ModuleNoLockFile()
...
+Method Test10LoadFromLockfile()
 {
     try {
         set moduleName = ..#ModuleK
-        do ..AssertInstallCreatesLockFileAsExpected(moduleName)
+
+        set sc = ##class(%IPM.Main).Shell("install " _ moduleName)
+        do $$$AssertStatusOK(sc, "Successfully installed " _ moduleName)
+
+        set packagePath = "/home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/lock-test/package/pack"
```
isc-dchui, line 318: A bit brittle: this will break if running outside of containers

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -248,43 +269,126 @@ Method Test07ModuleNoLockFile()
...
 {
     try {
-        set moduleName = ..#ModuleL
-        do ..AssertInstallCreatesLockFileAsExpected(moduleName)
+        set moduleName = ..#ModuleM
+
+        // First install without a lock file and confirm that the latest version of module H was chosen
+        set sc = ##class(%IPM.Main).Shell("install " _ moduleName)
+        do $$$AssertStatusOK(sc, "Successfully installed " _ moduleName _ " and dependencies using ""install""")
+        zpm "list-installed"
```
isc-dchui: Leftover debug code?

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -248,43 +269,126 @@ Method Test07ModuleNoLockFile()
...
+        set modH = ##class(%IPM.Storage.Module).NameOpen(..#ModuleH)
+        do $$$AssertEquals(modH.VersionString, "2.1.0", "Version 2.1.0 of dependency module H was installed via ""install""")
+
+        // Uninstall prior to installing from a lock file
+        set sc = ##class(%IPM.Main).Shell("uninstall -r " _ moduleName)
+
+        // Now install using module M's lock file
+        set sc = ##class(%IPM.Main).Shell("install -from-lockfile " _ moduleName)
+        do $$$AssertStatusOK(sc, "Successfully installed " _ moduleName _ " and dependencies using its lock file")
+        zpm "list-installed"
```
isc-dchui: Same as above?

src/cls/IPM/Main.cls
```diff
@@ -2394,36 +2396,40 @@ ClassMethod LoadInternal(
...
-        set tTargetDirectory = tFinalDirectory
-    }
-
-    set dotModules = ##class(%File).NormalizeDirectory(".modules", tTargetDirectory)
-    set tmpRepoMgr = ##class(%IPM.General.TempLocalRepoManager).%New(dotModules, 1)
-    set tSC = ##class(%IPM.Utils.Module).LoadNewModule(tTargetDirectory, .tParams, , pLog)
-    set tSC = $$$ADDSC(tSC, tmpRepoMgr.CleanUp())
-    $$$ThrowOnError(tSC)
+    if ($get(pCommandInfo("data", "FromLockFile"))) {
+        do ##class(%IPM.General.LockFile).InstallFromLockFile(tTargetDirectory, .tParams)
```
isc-dchui: I believe the `load` path will not install the parent module since that's only in the "else" path and it slips past test10 accidentally.

tests/integration_tests/Test/PM/Integration/LockFile.cls
```diff
@@ -248,43 +269,126 @@ Method Test07ModuleNoLockFile()
...
 /// Uses Module K
-Method Test09Module2DependenciesSameTransient()
+Method Test10LoadFromLockfile()
 {
     try {
         set moduleName = ..#ModuleK
-        do ..AssertInstallCreatesLockFileAsExpected(moduleName)
+
+        set sc = ##class(%IPM.Main).Shell("install " _ moduleName)
+        do $$$AssertStatusOK(sc, "Successfully installed " _ moduleName)
```
isc-dchui, line 316: Should also verify that the modules and dependencies have been actually installed

isc-dchui (changes requested):
A few things! Also still needs a description

src/cls/IPM/General/LockFile.cls
isc-dchui, line 1: Out of curiosity, why not move this file into the `.../IPM/General/LockFile` directory?

tests/integration_tests/Test/PM/Integration/LockFile.cls
isc-jlechtne: Still should be %Boolean, but will update the catch block to `$$$ThrowOnError` instead of returning a status

src/cls/IPM/General/LockFile.cls
isc-jlechtne: This was to be consistent with other directory paths under `.../IPM/General/`. `LogManager`, `SemanticVersion`, and `SemanticVersionExpression` all have class files directly under `General/` with directories to hold other classes that are related

src/cls/IPM/Main.cls
isc-jlechtne: Good catch I missed that! Will make sure to add that checking in Test10 as well as the fix here

src/cls/IPM/General/LockFile.cls
isc-jlechtne: We need different indicators to make sure we are only installing from a lock file of the base module (i.e. don't get to a dependency module and try installing its lock file too).

However, doing this in multiple parameters can be confusing. I have updated this to just have the `"FromLockFile"` parameter. When set to `1`, then we do want to install from the lock file. This lock file method now sets it to `0`, which indicates that we are already installing from a lock file so will ignore that in down the line modules.

tests/integration_tests/Test/PM/Integration/LockFile.cls
isc-jlechtne: It will, but our tests have a number of places where /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/... is hard coded in which this is consistent with. And the tests for now are always run in the container.

I think this should be kept as is for this issue and if we want to refactor, refactor all those references together in their own issue.

Issue #1084, opened 2026-03-03
0.10.6 regression on installation due to missing Python wheels


isc-dchui:
A clean install of IPM 0.10.6 won't fully work because the python wheels are no longer packaged (#1021). The logic should be amended such that on package, all `<PythonWheel>` resources in the module.xml are automatically included and the -export-python-deps flag will only consider the extra ones in the requirements.txt

PR #1087, opened 2026-03-04
Add -export-python-deps flag to publishing of IPM itself

isc-dchui:
Part of the solution to #1084. Will also need a manual run of the "publish to registry" workflow to get a working version of IPM 0.10.6 properly published.


isc-dchui:
@isc-kiyer I've added a call to run the unit tests in the "Clean install" section of the workflow that previously only did a smoke test. The unit tests should take no more than a minute to run.

PR #1107, opened 2026-03-26
Fix 1097: Nested Tests

### LINKED ISSUE #1097: Python wheel test failure + silent CI failure
Our CI has caught but somehow suppressed a failing test. This is pulled from the "Test and Verify" of https://github.com/intersystems/ipm/actions/runs/22909009044/job/66475799854

```
FAILED Test.PM.Integration.ProcessPythonWheel:TestWheelAndReqsPresentOnline: AssertEquals - reqVer== "1.6.4" was '1.6.2'

[USER|ZPM]	Verify SUCCESS
```
Running this test locally fails both the test and the verify phase as expected.

I did a bit of digging and it is a weird issue. It started cropping up on the CI of an unrelated issue (PR #1072), but when I try running the test on commits prior to that, it now fails despite passing CI earlier. Claude suggests the problem is pip related and somehow the old dist-info directories aren't being cleaned up properly and the solution is deleting them manually.

So this is a two part bug: 1) the failing test, and 2) why this failure isn't bubbling up in the CI.

isc-dchui:
## Description
- Fixes #1097
- Fixes the suppression of test failures.
    - This stems from nested `test`/`verify` calls, e.g. in the scopes integration test, we call `scope-test verify -only`, so when `zpm verify -only` is called, we end up nesting.
    - The test resource processor wasn't built to handle nesting due to its use of a single process-private global `^||%UnitTest.Manager.LastResult` to track test runs. This is `kill`ed at the start of the running of tests, which causes problems with nesting: the parent's results are killed by the child
    - The fix is to instead use a different process-private global `^||%UnitTest.Manager.AllResultsCount` that collates individual instantiations.
    - There are two integration test classes that use nested tests: `Scopes` and `FileCopy`.
    - Since this bug is dependent on test ordering, i.e. will only occur if a failing test is before a nested test, it didn't occur previously.
    - Note: `^||%UnitTest.Manager.LastResult` has been completely removed since IPM will no longer use it, so any custom code built on top of this will break

## Testing
Ran `zpm verify -only` and the tests pass. Also made sure if there is a failing test, the `verify` phase will correctly fail. For nested tests, made sure that if the nested test fails, it will be properly reported in the summary at the end.

## Checklist

- [x] Changelog entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #1057: Fix IPM not cleaning up after itself on self-uninstall
 - #1122: Packaging should recognize resources in dependency modules set to deploy
 - #1119: The update command should check version requirements using post-update values instead of what's currently installed
+- #1097: The Test resource processor now supports nested tests and installing Python dependencies from requirements.txt will correctly override wheels
 
 ## [0.10.6] - 2026-02-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #1057: Fix IPM not cleaning up after itself on self-uninstall
 - #1122: Packaging should recognize resources in dependency modules set to deploy
 - #1119: The update command should check version requirements using post-update values instead of what's currently installed
-- #1097: The Test resource processor now supports nested tests and installing Python dependencies from requirements.txt will correctly override wheels
+- #1097: The Test resource processor now supports nested tests
 
 ## [0.10.6] - 2026-02-24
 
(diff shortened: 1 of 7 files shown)


isc-tleavitt (commented):
I want to discuss this in our next IPM meeting.

I *think* the effect of this will be that zpm "install" of the same package will upgrade Python dependencies based on semvers in requirements.txt where previously it wouldn't unless you specifically told it to.

Other possible concern - running processes while the install runs. But that's probably an issue any time your upgrading even without explicitly deleting directory chains first - possibly something that could be addressed in a venv-ish solution.

This is tied to the challenges with Python dependency management more generally.

Especially where our target here is fixing CI problems, I'm hesitant to take on a significant user facing change as part of it. But we shouldn't need to solve the whole pseudo-venv / python dependency management problem; maybe this is a reasonable incremental step.

isc-tleavitt (approved):

Issue #1119, opened 2026-04-10
IPM update dependency version conflict


isc-jlechtne:
## Describe the bug
Found an issue with update when the dependencies of the base module also get updated. Issue found when updating a module from 2.0.0 to 2.1.0. A dependency in 2.1.0 gets bumped to `6.2.0+snapshot` while 2.0.0 requires `^5.4.1`. Trying to perform this update encountered the following error:

> Requested version (dependency 6.2.0+snapshot) does not satisfy the requirements of other modules installed in the current namespace (dependency: ^5.4.1).

## To Reproduce
1. Have modules A, B, & C. Module A depends on B which in turn depends on C
2. Pre update versions:
- Module A: 2.0.0, requires B 1.0.0
- Module B: 1.0.0, requires C ^5.4.1
- Module C: 5.6.45+snapshot
3. Post update versions
- Module A: 2.1.0+snapshot, requires B ^2.0.0
- Module B: 2.0.0+snapshot, requires C ^6.1.15
- Module C: 6.2.0+snapshot
4. Run `zpm "install <Module A> 2.0.0"`
5. Run `zpm "update <Module A> 2.1.0+snapshot"` (this should error)

## Expected behavior
Expected behavior is for Module A and all of its dependencies to successfully update as the post-update versions do satisfy requirements.

isc-dchui:
Sounds related to #430.

isc-dchui:
Addressed by #1120

PR #1120, opened 2026-04-10
Update should check version requirements using post-update values instead of what's currently installed

isc-jlechtne:
## Description
Addresses issue #1119 

Source of the fix is centered around the `%IPM.Utils.Module::GetRequiredVersionExpression()` method. It uses a SQL call to determine what versions are installed and what relies on them for checking version requirements. It also accepts a list of modules to exclude from this check. This fix adds a dependency graph to that list. When given a dependency graph, the method will now add the modules from the exclusion list to add them manually after the SQL call. This addresses the update case since the dependency graph will have the required version expressions associated with what update will install as opposed to what's currently installed.

## Testing
PR includes a new integration test to the update suite to check for this case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #1052: In a namespace with mapped IPM, the `info` command works again and the intro message displays the IPM version and where its mapped from
 - #1102: %IPM.Storage.QualifiedModuleInfo:%New() will now copy over version properties when passed in a resolvedReference
 - #1112: Packaging a module with a globals resource now respects SourcesRoot, placing the exported file at the correct path in the tarball
-- #1119: Update should check version requirements using post-update values instead of what's currently installed
+- #1119: The update command should check version requirements using post-update values instead of what's currently installed
 
 ## [0.10.6] - 2026-02-24
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #1052: In a namespace with mapped IPM, the `info` command works again and the intro message displays the IPM version and where its mapped from
 - #1102: %IPM.Storage.QualifiedModuleInfo:%New() will now copy over version properties when passed in a resolvedReference
 - #1112: Packaging a module with a globals resource now respects SourcesRoot, placing the exported file at the correct path in the tarball
+- #1119: Update should check version requirements using post-update values instead of what's currently installed
 
 ## [0.10.6] - 2026-02-24
 
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -351,7 +351,7 @@ ClassMethod GetModuleObjectFromString(
 /// A list of modules to exclude may be provided (for example, if these modules would be updated at the same time).
 /// 
 /// If provided a dependency graph, will use versions defined there instead of what the SQL call returns.
-/// This is especially important for update, where we want to check requirements with post-update versions as opposed to what's currently installed.
+/// This is especially important for the update command, where we want to check requirements with post-update versions as opposed to what's currently installed.
 ClassMethod GetRequiredVersionExpression(
 	moduleName As %String,
 	excludeModules As %List = "",
--- a/src/cls/IPM/Utils/Module.cls
+++ b/src/cls/IPM/Utils/Module.cls
@@ -114,7 +114,7 @@ ClassMethod LoadModuleReference(
 
         // Ensure requested versions match those required by other modules in the namespace, excluding versions currently being installed
         // (the requirements of such modules are already known to be satisfied)
-        set tSC = ..GetRequiredVersionExpression(pModuleName,,.tExpression,.tSourceList)
+        set tSC = ..GetRequiredVersionExpression(pModuleName,,.pDependencyGraph,.tExpression,.tSourceList)
         if $$$ISERR(tSC) {
             quit
         }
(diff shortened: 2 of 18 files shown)


src/cls/IPM/Utils/Module.cls
```diff
@@ -350,35 +350,76 @@ ClassMethod GetModuleObjectFromString(
 /// Returns a semantic version expression capturing all version requirements for a given module name in the current namespace.
 /// A list of modules to exclude may be provided (for example, if these modules would be updated at the same time).
```
isc-dchui, line 351: Adding a comment here explaining the use of the dependency graph would be nice.

src/cls/IPM/Utils/Module.cls
```diff
@@ -350,35 +350,76 @@ ClassMethod GetModuleObjectFromString(
...
-            set pExpression = pExpression.And(tVersionExpr)
-            set pSourceList = pSourceList_$listbuild($listtostring(tResult.%Get("ModuleNames"),", ")_": "_tVersion)
+        while result.%Next(.sc) {
+            $$$ThrowOnError(sc)
+            set version = result.%Get("Version")
+            $$$ThrowOnError(##class(%IPM.General.SemanticVersionExpression).FromString(version,.versionExpr))
+            set expression = expression.And(versionExpr)
+            set sourceList = sourceList_$listbuild($listtostring(result.%Get("ModuleNames"),", ")_": "_version)
         }
-        $$$ThrowOnError(tSC)
```
isc-dchui, line 375: I believe we need the $$$ThrowOnError up here to catch errors that are returned and do not enter the while loop.

src/cls/IPM/Utils/Module.cls
```diff
@@ -350,35 +350,76 @@ ClassMethod GetModuleObjectFromString(
...
+                // If this is a new required version expression, add a new item to the list
+                if (newVersion) {
+                    $$$ThrowOnError(##class(%IPM.General.SemanticVersionExpression).FromString(version, .versionExpr))
+                    set expression = expression.And(versionExpr)
+                    set sourceList = sourceList _ $listbuild(requiringModuleName _ ": " _ version)
+                }
+            }
+        }
+
+        $$$ThrowOnError(sc)
```
isc-dchui: See above comment.

src/cls/IPM/Utils/Module.cls
```diff
@@ -1251,7 +1292,7 @@ ClassMethod LoadDependencies(
 
     set reloadSnapshots = +$get(pParams("UpdateSnapshots"))
     set permitDowngrade = +$get(pParams("PermitDowngrade"))
-    set ignoreInstalled = +$get(pParams("IgnoreInstalled"))
+    set ignoreInstalled = +$get(pParams("zpm", "IgnoreInstalled"))
```
isc-dchui: Can you explain the rationale behind this change and why it doesn't apply to the other parameters above it?

isc-dchui (commented):
First round of feedback!

src/cls/IPM/Utils/Module.cls
isc-jlechtne: This really shouldn't be a part of this review, and I will take it out.

The origin was me noticing when debugging that the -DIgnoreInstalled=1 flag showed up here as pParams("zpm", "IgnoreInstalled"), not pParams("IgnoreInstalled") for every case except for the base module on "install". Added the "zpm" so that every call used it, but it didn't end up helping me with the bug and I'm less confident it's something that even needs to be changed so no reason for me to touch this code.

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -662,4 +662,65 @@ Method TestDevModePropagation()
     do $$$AssertEquals(dep.DeveloperMode, 0)
 }
 
+/// Test case to make sure that update doesn't fail due to version errors that get resolved as a part of update
+/// i.e. - Module A 1.0.0 requires Module B ^1.0.0
+///      - Module A 2.0.0 requires Module B ^2.0.0
+///      - Pre-Update: Module A 1.0.0, Module B 1.3.0
```
isc-jili: I think these pre and post versions in the comment don't match what is being set in the vars below

src/cls/IPM/Utils/Module.cls
```diff
@@ -349,36 +349,79 @@ ClassMethod GetModuleObjectFromString(
...
+                    quit
+                }
+                // Name of module which requires this one + the version expression string it requires
+                set requiringModuleName = $piece(key, " ")
+                set version = dependencyGraph(moduleName, key)
+
+                // Iterate over sourceList; if the required version expression is equivalent to this one, add this module name to that version
+                set newVersion = 1
+                for i=1:1:$listlength(sourceList) {
+                    if $find($listget(sourceList, i), version) {
```
isc-jili, line 405: Is is possible that $find is too loose of a search? ie. would there be a case where the version incorrectly matches only a substring?

isc-jili (changes requested):
Left a couple comments but looks good @isc-jlechtne !

src/cls/IPM/Utils/Module.cls
```diff
@@ -349,36 +349,79 @@ ClassMethod GetModuleObjectFromString(
 
 /// Returns a semantic version expression capturing all version requirements for a given module name in the current namespace.
 /// A list of modules to exclude may be provided (for example, if these modules would be updated at the same time).
+/// 
+/// If provided a dependency graph, will use versions defined there instead of what the SQL call returns.
+/// This is especially important for update, where we want to check requirements with post-update versions as opposed to what's currently installed.
```
isc-jili: nit: "for the update command"

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - #1052: In a namespace with mapped IPM, the `info` command works again and the intro message displays the IPM version and where its mapped from
 - #1102: %IPM.Storage.QualifiedModuleInfo:%New() will now copy over version properties when passed in a resolvedReference
 - #1112: Packaging a module with a globals resource now respects SourcesRoot, placing the exported file at the correct path in the tarball
+- #1119: Update should check version requirements using post-update values instead of what's currently installed
```
isc-jili: nit: "The update command"

isc-jili:
I am realizing actually that even though we framed this issue as "update command" related, GetRequiredVersionExpression() is called by many more functionalities than just update. So I have 2 questions:
1. With regards to wording, I am not sure if we should tweak places that this change explicitly refer to the update command to be more general?
2. Should we add a test case for another situation that this would also fix (theoretically) like #430 to kill 2 birds with one stone?

isc-dchui:
> I am realizing actually that even though we framed this issue as "update command" related, GetRequiredVersionExpression() is called by many more functionalities than just update. So I have 2 questions:
> 
> 1. With regards to wording, I am not sure if we should tweak places that this change explicitly refer to the update command to be more general?
> 2. Should we add a test case for another situation that this would also fix (theoretically) like [ModuleDependencies and Version #430](https://github.com/intersystems/ipm/issues/430) to kill 2 birds with one stone?

For 2, there's a sample module in PR #1050 you can copy over.

src/cls/IPM/Utils/Module.cls
isc-jlechtne: The syntax of one element in sourceList looks like `"module-a, module-b: ^1.0.0"` for a case where both module-a and module-b require version ^1.0.0 of whatever module is being evaluated (the case here being adding some module-x to this string if it also requires ^1.0.0) so there won't be a case of this incorrectly passing since that would require a module name to have the identical version string in it.

Could alternatively do `$piece($listget(sourceList, i), " ", *) = version` as a way to completely isolate the version expression but splitting up the string on the delimiter should be unnecessary for this case.

isc-jili (approved):

isc-cborbonm (approved):
looks good to me!

Issue #1122, opened 2026-04-17
Deploy="true" does not work for dependencies


isc-jlechtne:
## Describe the bug
Found an issue where having packages/classes in a dependency's module.xml with `Deploy="true"` does not work. After install, those resources are still readable and writable. The base module does correctly have deployed code for resources that specify it.

## To Reproduce
1. Have a module with a resource marked with `Deploy="true"` that is a dependency of another module
2. Install the base module which includes this dependency
3. Package the base module, make sure to include the `-export-deps 1` flag to include the dependency
4. Load module from the packaged tarball

## Expected behavior
The resources installed marked with deploy should be deployed on load from the packaged tarball.

isc-dchui:
Done in #1123

PR #1123, opened 2026-04-17
Packaging should recognize resources in dependency modules set to deploy

isc-jlechtne:
Addresses #1122 

- If `-export-deps` provided, have that also set the recurse flag in packaging (for deploying dependency resources)
- Create a Deployed.xml studio document for dependency modules that have deployed resources in addition to the base module

### DIFF
--- a/src/cls/IPM/ResourceProcessor/Default/Document.cls
+++ b/src/cls/IPM/ResourceProcessor/Default/Document.cls
@@ -280,10 +280,9 @@ Method OnExportDeployedItem(
     try {
         set tVerbose = $get(pParams("Verbose"),0)
         set tDeveloperMode = $get(pParams("DeveloperMode"),0)
-        set isDepAndRecurse = $get(pParams("Recurse"),0) && $find(pFullExportPath, ".modules")
         set executingPhases = $listlength(..ResourceReference.Module.Lifecycle.PhaseList)
-        if (tDeveloperMode || ('..ResourceReference.Deploy) || ('executingPhases && 'isDepAndRecurse)) {
-            // Only export deployed items if we say to, are in a non-developer-mode context,
+        if (tDeveloperMode || ('..ResourceReference.Deploy) || ('executingPhases && '$get(pParams("Recurse"),0))) {
+            // Only export deployed items if they are marked for deploy, we are in a non-developer-mode context,
             // and are currently executing module lifecycle phases or this is a dependency module with the recurse flag set
             // OnExportItem may also be called from a source control extension context, and in this case PhaseList will be empty.
             quit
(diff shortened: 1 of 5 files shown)


src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -280,8 +280,11 @@ Method OnExportDeployedItem(
     try {
         set tVerbose = $get(pParams("Verbose"),0)
         set tDeveloperMode = $get(pParams("DeveloperMode"),0)
-        if (tDeveloperMode || ('..ResourceReference.Deploy) || ('$listlength(..ResourceReference.Module.Lifecycle.PhaseList))) {
-            // Only export deployed items if we say to, are in a non-developer-mode context, and are currently executing module lifecycle phases.
+        set isDepAndRecurse = $get(pParams("Recurse"),0) && $find(pFullExportPath, ".modules")
+        set executingPhases = $listlength(..ResourceReference.Module.Lifecycle.PhaseList)
+        if (tDeveloperMode || ('..ResourceReference.Deploy) || ('executingPhases && 'isDepAndRecurse)) {
+            // Only export deployed items if we say to, are in a non-developer-mode context,
```
isc-eneil: Nit: clarify in this comment what it means to "say to" export deployed items

src/cls/IPM/ResourceProcessor/Default/Document.cls
```diff
@@ -280,8 +280,11 @@ Method OnExportDeployedItem(
     try {
         set tVerbose = $get(pParams("Verbose"),0)
         set tDeveloperMode = $get(pParams("DeveloperMode"),0)
-        if (tDeveloperMode || ('..ResourceReference.Deploy) || ('$listlength(..ResourceReference.Module.Lifecycle.PhaseList))) {
-            // Only export deployed items if we say to, are in a non-developer-mode context, and are currently executing module lifecycle phases.
+        set isDepAndRecurse = $get(pParams("Recurse"),0) && $find(pFullExportPath, ".modules")
```
isc-eneil: I'm curious why you are checking for the .modules dir within the export path rather than checking for the ExportDependencies parameter, for example?

isc-kiyer: I think we may want to consider removing the PhaseList check altogether instead. We should only care about source control classes when in dev mode and since we never deploy in dev mode, this isn't an issue. We used to support deploying in dev mode so that was likely the issue.

isc-jlechtne: That's a good point, can just use the Recurse param since that'll satisfy cases where the module doesn't have a phaselist but is being processed here

isc-kiyer (approved):

isc-jili (approved):

PR #1124, opened 2026-04-20
Add iriscli and ipm container utility scripts

### LINKED ISSUE #1110: OS/Container utility scripts
Internally we have some OS level, container scripts to drop straight into the IRIS shell or the IPM shell from terminal. Essentially wrappers around `docker-exec -it <container> bash` + `iris session iris` + `zpm`. We should move them into IPM itself. We should also add documentation for container patterns.

isc-dchui:
## Description
- Resolves #1110 
- Adds two container utility scripts
    - `iriscli` launches an interactive IRIS terminal (essentially wrapper for `iris session iris`), e.g.`docker compose exec -it iris iriscli`
    - `ipm` allows running of IPM commands, e.g. `docker compose exec iris ipm list`
- Auto-installs the scripts in Unix/Linux to `~/bin/` and `~/.local/bin/` so they will work both inside and outside the dev containers 

## Testing
Ran the scripts manually from the host machine. Example:
```
PS C:\Users\dchui\Documents\ipm\ipm2> docker exec -it ipm2-iris-1 ipm list

Node: 122fb10eb60e, Instance: IRIS

USER>
IPM (zpm)                     0.10.7-SNAPSHOT
vscode-per-namespace-settings 1.0.0
testcoverage                  4.1.2
```

## Checklist

- [x] Changelog entry added.

### DIFF
--- a/ipm
+++ b/ipm
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+# ipm: convenience wrapper to run a ZPM command in an IRIS session
+# Usage examples:
+#   ipm isc.genai test -only
+#   ipm -U PLAZA isc.genai test -only
+
+ARGS=()
+CMD=()
+
+# Allow namespace via env like iriscli
+if [ -n "$IRIS_NAMESPACE" ]; then
+  ARGS+=( -U "$IRIS_NAMESPACE" )
+elif [ -n "$IRISNAMESPACE" ]; then
+  ARGS+=( -U "$IRISNAMESPACE" )
+fi
+
+# Optional leading -U <ns>
+if [ "$1" = "-U" ] && [ -n "$2" ]; then
+  ARGS+=( -U "$2" )
+  shift 2
+fi
+
+# Remaining args compose the ZPM command string
+while [[ $# -gt 0 ]]; do
+  CMD+=( "$1" )
+  shift
+done
+
+if [ ${#CMD[@]} -eq 0 ]; then
+  echo "Usage: ipm [ -U <NAMESPACE> ] <zpm command and args...>" >&2
+  exit 1
+fi
+
+# Join CMD array into a single string
+CMDSTR="${CMD[*]}"
+# Escape quotes for ObjectScript string literal
+CMDSTR_ESC=${CMDSTR//\"/\"\"}
+
+# Build an OS snippet and feed to iris session
+(
+  echo "zpm \"${CMDSTR_ESC}\":1:1"
+  echo "halt"
+) | iris session ${ISC_PACKAGE_INSTANCENAME:-iris} "${ARGS[@]}"
--- a/ipm
+++ b/ipm
@@ -1,44 +1,73 @@
 #!/bin/bash
 
-# ipm: convenience wrapper to run a ZPM command in an IRIS session
-# Usage examples:
-#   ipm isc.genai test -only
-#   ipm -U PLAZA isc.genai test -only
+# ipm: convenience wrapper to run an IPM command in an IRIS session
+#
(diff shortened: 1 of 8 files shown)


ipm
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+# ipm: convenience wrapper to run a ZPM command in an IRIS session
```
isc-kiyer: Would it make sense for this to use iriscli instead of duplicating some of the logic there?

isc-kiyer (changes requested):
@isc-dchui This is great! Just the one note. Seems like for now this is just for general dev use but is this something we would consider distributing with IPM so that if it is loaded into an instance, the scripts are available? We could say it only works for unix based systems to make life a bit easier for us to maintain

isc-tleavitt (changes requested):
Agreed with @isc-kiyer. We should install to ~/.local/bin and need to make sure mode is correct (+x)

ipm
isc-dchui: There's not quite enough code overlap for it to be worth it in this case.

isc-dchui:
Added the installing to both `~/bin/` and `~/.local/bin/` in the install script and removed the lines from Dockerfile to keep installation in one place. This should work both inside and outside the container. `~/.local/bin/` is Unix/Linux standard and `~/bin/` is used by containers.

preload/cls/IPM/Installer.cls
```diff
@@ -476,4 +476,68 @@ ClassMethod CleanupOldMappings() As %Status
...
+/// Always returns $$$OK — script installation is best-effort and must not block the install.
+ClassMethod InstallScripts() As %Status
+{
+    // Scripts are bash; skip entirely on non-Unix platforms.
+    if '$$$isUNIX {
+        quit $$$OK
+    }
+    try {
+        // The scripts are staged under ${ipmdir}bin/ by <FileCopy> in module.xml.
+        set module = ##class(%IPM.Storage.Module).NameOpen("ZPM", , .openSC)
```
isc-kiyer: Nit: use macro for module name

iriscli
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+ARGS=()
```
isc-kiyer, line 23: Nit: add usage examples here and perhaps a help modifier to see command usage

iriscli
```diff
@@ -0,0 +1,34 @@
...
+done
+
+if [ -n "$file" ]; then
+    (
+        for param in ${PARAMS[@]}; do
+            echo "Set params(\$i(params)) = \"${param//\"/\"\"}\""
+        done
+        egrep -v '^(;|#|//)|^$' $file; 
+        echo halt
+    ) | iris session $ISC_PACKAGE_INSTANCENAME "${ARGS[@]}"
```
isc-kiyer: Would be nice to be able to specify the instance name in the command since this env var isn't set for non container envs

isc-kiyer (changes requested):
@isc-dchui few more small notes

iriscli
isc-dchui: Added more documentation and a `-h`/`--help` flag

iriscli
isc-dchui: Added a `-i <instance>` flag

preload/cls/IPM/Installer.cls
isc-dchui: We can't actually use the macro when bootstrapping from just the installer because the include file isn't available yet, so did the next best thing of adding a Parameter 'MODULENAME"

isc-tleavitt (approved):

isc-kiyer (approved):

PR #1129, opened 2026-04-30
Fix: IPM Update Resource Conflict

### LINKED ISSUE #1128: IPM Update Resource Conflict
## Describe the bug
This is similar in concept to the issue #1119 except with resources instead of dependency versions. This comes up in a case where a resource is moved from one module to another. 

> ERROR! Resource 'MyClass.CLS' is explicitly part of module 'module-b', and cannot also be in module 'module-c'

## To Reproduce
Using the same module setup as in #1119, but adding a resource `MyClass.CLS` to the scenario:
1. Have modules A, B, & C. Module A depends on B which in turn depends on C
2. Pre update setup:
- Module A (2.0.0)
- Module B, has resource `MyClass.CLS`
- Module C
3. Post update versions
- Module A (2.1.0+snapshot)
- Module B
- Module C, has resource `MyClass.CLS`
4. Run `zpm "install <Module A> 2.0.0"`
5. Run `zpm "update <Module A> 2.1.0+snapshot"` (this should error)

## Expected behavior
Expected behavior is for Module A and all of its dependencies to successfully update as `MyClass.CLS` is only a part of module-c in the post-update environment (no longer with module-b).

isc-jlechtne:
Fixes #1128 

Updates update test with the following resources:
- Adds pre-update resources to dependency modules `module-a` (`Class1.cls`) and `module-b` (`Class2.cls`)
- For post-update versions: moved resources so that `Class1.cls` is in `module-c` and `Class2.cls` is in `module-a`
- Creates an intermediate `module-a v2.0.5+snapshot` with `Class1.cls`

Steps added:
1. After install of pre-update versions, attempt to update to `module-a v2.0.5+snapshot`. This will fail due to `Class1.cls` belonging to both `module-a` and `module-c`
2. Update to `module-a v2.1.0+snapshot` has `Class1` and `Class2` moving around modules, which will no longer fail due to this fix


src/cls/IPM/Storage/Module.cls
```diff
@@ -1371,6 +1364,12 @@ Method GetResolvedReferences(
         }
 
         if pLockedDependencies {
```
isc-cborbonm, line 1366: question: why did this get moved to only when pLockedDependencies is set?

src/cls/IPM/Storage/ResourceReference.cls
```diff
@@ -344,6 +352,21 @@ ClassMethod NonNullResourceNameOpen(
     quit tResult
 }
 
+/// When given a name for a resource, set its MarkedForUpdate flag
+/// See property description for more details on the usage of MarkedForUpdate
+ClassMethod MarkResourceForUpdate(
+	resourceName As %String,
+	marking As %Boolean = 1)
```
isc-cborbonm, line 358: is there a use case for this method when marking would be set to 0?

isc-cborbonm (commented):
just left a couple of comments! but it looks great to me 😃

src/cls/IPM/Storage/ResourceReference.cls
```diff
@@ -54,6 +54,14 @@ Property Children As array Of %String(MAXLEN = 1, XMLPROJECTION = "NONE");
 
 Index Children On Children(KEYS);
 
+/// Tells us if this resource belongs to a module that will be processed as a part of the "Update" command
```
isc-eneil: Wording suggestion: "Indicates whether this ResourceReference belongs to a module currently being processed as part of the update command."

src/cls/IPM/Storage/ResourceReference.cls
```diff
@@ -54,6 +54,14 @@ Property Children As array Of %String(MAXLEN = 1, XMLPROJECTION = "NONE");
...
+/// If set to true, %OnBeforeSave() will not error if another resource of the same name tries to be saved
+///     and will instead save that resource with the other module as new
+/// This is useful for an "Update" command where a resource was moved from one module to another.
```
isc-eneil: I would condense this comment for simplicity but otherwise looks good!

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -704,9 +707,15 @@ Method TestDependencyVersionsUpdated()
...
-    // Uninstall module-d then try the update again
+    // Uninstall module-d since we don't need it anymore
     set sc = ##class(%IPM.Main).Shell("uninstall -v " _ modD)
     do $$$AssertStatusOK(sc, "Successfully uninstalled " _ modD)
+
+    // Try updating module-a to a version that will fail with same resource as in module-c
+    set sc = ##class(%IPM.Main).Shell("update -v " _ modA _ " " _ modAErrorVersion)
+    do $$$AssertStatusNotOK(sc, "Updating " _ modA _ " to version " _ modAErrorVersion _ " fails due to same resource in modules a and c")
+
+    // Not do an update that should succeed
```
isc-eneil: Typo: Now?

tests/integration_tests/Test/PM/Integration/Update.cls
```diff
@@ -704,9 +707,15 @@ Method TestDependencyVersionsUpdated()
     set sc = ##class(%IPM.Main).Shell("update -v " _ modA _ " " _ modAPostVersion)
     do $$$AssertStatusNotOK(sc, "Updating " _ modA _ " fails due to version errors with previously installed " _ modD)
 
-    // Uninstall module-d then try the update again
+    // Uninstall module-d since we don't need it anymore
     set sc = ##class(%IPM.Main).Shell("uninstall -v " _ modD)
     do $$$AssertStatusOK(sc, "Successfully uninstalled " _ modD)
+
+    // Try updating module-a to a version that will fail with same resource as in module-c
```
isc-eneil: Is this supposed to fail because the resource was not moved, but instead duplicated across modules on update? If so, I'd suggest clarifying that in the comment.

isc-eneil:
@isc-jlechtne mind adding your test cases to the PR description? Thanks!
Also, I took a look and left a few nits about comments. Looks good!

isc-dchui (commented):
The only thing not already mentioned is that "addresses" is not actually one of the magic github closing words so won't auto-link the issue. You'll need "fixes" or "resolves". Also more detail in the description would be nice!

src/cls/IPM/Storage/Module.cls
isc-jlechtne: pDependencyGraph is only used within this if-statement. Felt unnecessary to me to build the dependency graph earlier if the execution never got to the point where we actually use it

src/cls/IPM/Storage/ResourceReference.cls
isc-jlechtne: Not currently but for what the method is doing I felt it made sense to have the capability work both ways.

tests/integration_tests/Test/PM/Integration/Update.cls
isc-jlechtne: Correct

isc-jlechtne: Also just realized I didn't include the 2.0.5+snapshot erroring version of module-a here. Missed it because the update is still failing as expected, but not in the way we want it to.

isc-cborbonm (dismissed):
looks good to me :)

isc-cborbonm (dismissed):

isc-dchui (dismissed):

isc-dchui (approved):

isc-dchui (changes requested):
Sorry misclick. Looks like your merge isn't complete

isc-cborbonm (dismissed):

isc-dchui (approved):

isc-cborbonm (approved):

PR #1146, opened 2026-05-11
On install/load, add a file systems permissions check

### LINKED ISSUE #1053: Checking directory permissions for copying assets on install
Improvement: If a terminal or the windows console is started in a user context who lacks sufficient file system permissions to write assets, a dedicated error or warning should be displayed prior to the module installation. Currently now exception is thrown during install. 
I have found the corresponding audit event:

Image

isc-dchui:
## Description
- Resolves #1053 
- Adds the check on both the `install` and `load` command.
- Replaces $$$FileTempDirSys macro with a utility function wrapper that also handles errors cleanly (%IPM.Utils.File:CreateTempDirectory)

## Testing
Manually tested by changing the permissions of the `/usr/irissys/ipm` directory and verifying the error appears when trying to install.
```
zpm:USER>install isc.json

ERROR! Insufficient file system permissions to write to '/usr/irissys/ipm/'. Cannot install module 'isc.json'.
```

## Checklist

- [x] Changelog entry added.


src/cls/IPM/Utils/Module.cls
```diff
@@ -226,6 +226,15 @@ ClassMethod LoadModuleFromArchive(
     try {
         set tVerbose = $get(pParams("Verbose"))
 
+        // Check base IPM install location
+        set ipmBaseDir = ##class(%Library.File).NormalizeDirectory(##class(%SYSTEM.Util).DataDirectory() _ "ipm")
```
isc-kiyer: Don't we have a macro/a setting to be configured for the ipmDir? we should use that right?

src/cls/IPM/Main.cls
```diff
@@ -2351,6 +2351,12 @@ ClassMethod LoadInternal(
     set extension = $$$lcase($piece(tDirectoryName,".",*))
     if ##class(%File).Exists(tDirectoryName) && ((extension="tgz") || (extension="tar.gz")) {
         set tTargetDirectory = $$$FileTempDirSys
+        if tTargetDirectory = ("-" _ -tTargetDirectory) {
```
isc-kiyer: Can you explain this if condition? Its a little odd to see negation of what seems like it should be a string? Is it an integer when there is no permission?

isc-kiyer (changes requested):
@isc-dchui some small questions

isc-tleavitt (dismissed):
Agreed with Keshav on comment needed re: weird negation, otherwise looks good.

src/cls/IPM/Utils/Module.cls
isc-dchui: Not yet--it's in #1031 but I can drive that forward

src/cls/IPM/Main.cls
isc-dchui: Right, it's a negative integer code when there's an error, so it's checking for that. It's a bit ugly/confusing but is used in %IPM.Repo.Oras.PublishService:PublishModule:
`if tDir = ("-" _ -tDir) { // $zu(140,17) can return a negative number if it fails`

Breaking it down a bit:
tTargetDirectory can either be an actual path like `\foo\bar\` or a negative integer code like `-13`. The coercion, negation, and comparison then check if it's a negative integer or a string.

isc-kiyer: @isc-dchui sorry I should have been more clear. Please add this description in the code so its apparent what the code is doing (I figured it was the return code but the variable naming makes it confusing)

src/cls/IPM/Utils/Module.cls
isc-dchui: Changed to use customizable module root

src/cls/IPM/Main.cls
isc-dchui: I refactored it entirely by replaced the $$$FileTempDirSys macro with a utility function wrapper that also handles errors clearly (%IPM.Utils.File:CreateTempDirectory)

isc-kiyer (dismissed):

isc-tleavitt (approved):

PR #975, opened 2025-12-03
BuildDependencyGraph performance enhancements

isc-kiyer:
Resolves #481. Improve the speed of BuildDependencyGraph massively by reworking the algorithm. Details noted in changelog.
Add ability to reduce semantic version expressions that are ANDed together.

Performance benchmark of old vs new algorithm:

**Few dependencies (internal ref - HealthConnect.Common):**
```
===== ORIGINAL BuildDependencyGraph =====
Starting original build dep graph...
.900729 seconds elapsed.

===== ITERATIVE BuildDependencyGraph =====
Starting iterative build dep graph...
.403868 seconds elapsed.

===== COMPARISON =====
Original time:     .900729 seconds
Iterative time:    .403868 seconds
Iterative improv:  55.16%
```

**Medium number of dependencies (internal ref: HealthShare.Common):**
```
===== ORIGINAL BuildDependencyGraph =====
Starting original build dep graph...
50.244799 seconds elapsed.

===== ITERATIVE BuildDependencyGraph =====
Starting iterative build dep graph...
.735122 seconds elapsed.

===== COMPARISON =====
Original time:     50.244799 seconds
Iterative time:    .735122 seconds
Iterative improv:  98.54%
```

**Many dependencies (internal ref: HealthShare.SysIndex):**
```
===== ORIGINAL BuildDependencyGraph =====
Starting original build dep graph...
221.710365 seconds elapsed.

===== ITERATIVE BuildDependencyGraph =====
Starting iterative build dep graph...
.994169 seconds elapsed.

===== COMPARISON =====
Original time:     221.710365 seconds
Iterative time:    .994169 seconds
Iterative improv:  99.55%
```

PR #1153, opened 2026-05-12
Scope information in output array of BuildDependencyGraph()

### LINKED ISSUE #1152: BuildDependencyGraph() should include scope if dependency is scoped
## Describe the problem
The output of %IPM.Storage.Module:BuildDependencyGraph() does not include information for scoped dependencies.

## Describe the solution
The output of BuildDependencyGraph() includes the scope for scoped dependencies.

isc-cborbonm:
## Description
- Closes #1152: Adds information about scoped dependencies in the output array of BuildDependencyGraph()

## Testing
1. New test modules under build-dependency-graph-scoped-dep
2. New test class Test.PM.Integration.BuildDependencyGraphScopedDeps:
a. checks that non-scoped dependencies have an empty string in the dependency info list
b. checks that scoped dependencies see their scope (namely, "test" for this example) in the dependency info list


isc-dchui (commented):
Can I ask what the context of this change is? What is the bigger picture change that this enables? (And why not bundle this change with that one?)

isc-cborbonm:
> Can I ask what the context of this change is? What is the bigger picture change that this enables? (And why not bundle this change with that one?)

Great question! The bigger picture is to allow a user who calls BuildDependencyGraph() to see at a glance which transitive dependencies are scoped and what their scope is (if one exists). The rest of the change isn't within IPM code itself but rather a different application that makes use of IPM (hence not bundling the change in this PR).

tests/integration_tests/Test/PM/Integration/BuildDependencyGraphScopedDeps.cls
```diff
@@ -0,0 +1,52 @@
+Class Test.PM.Integration.BuildDependencyGraphScopedDeps Extends Test.PM.Integration.Base
```
isc-dchui, line 1: It would also be good to test `Scope="Verify"`

isc-cborbonm: done! thanks for the suggestion

tests/integration_tests/Test/PM/Integration/BuildDependencyGraphScopedDeps.cls
```diff
@@ -0,0 +1,56 @@
...
+    do $$$AssertStatusOK(sc,"Deleted "_..#REPONAME_"repo successfully.")
+    quit sc
+}
+
+/// BuildDependencyGraph should include scope information for scoped transitive dependencies
+Method TestDependencyGraphScopedDependency()
+{
+    do $$$AssertStatusOK(##class(%IPM.Utils.Module).LoadModuleFromDirectory(..RepoPath_..#MODNAME), "Successfully loaded module from directory.")
+    set module = ##class(%IPM.Storage.Module).NameOpen(..#MODNAME,,.sc)
+    do $$$AssertStatusOK(sc, "Succcessfully opened module object.")
```
isc-jili: nit: Typo in "successfully"

isc-jili (dismissed):
One typo found but otherwise looks great! @isc-cborbonm

src/cls/IPM/Storage/Module.cls
```diff
@@ -1142,7 +1142,7 @@ Method ProcessSingleDependencyIterative(
                 localObj.Version.Satisfies(searchExpr) &&
                 ((version = "") || (version = localObj.VersionString))
             if installedVersionValid && '(localObj.Version.IsSnapshot() && pForceSnapshotReload) {
-                set pDependencyGraph(pDep.Name) = $listbuild(pDepth,"",localObj.VersionString,pDep.DisplayName)
+                set pDependencyGraph(pDep.Name) = $listbuild(pDepth,"",localObj.VersionString,pDep.DisplayName,pDep.Scope)
```
isc-kiyer, line 1147: Include scope in the description of the BuildDependencyGraph method as well when it describes the structure of the dep graph array

isc-kiyer (commented):
@isc-cborbonm just the one small note. Looks good!

src/cls/IPM/Storage/Module.cls
isc-cborbonm: good point, thank you!

tests/integration_tests/Test/PM/Integration/BuildDependencyGraphScopedDeps.cls
isc-cborbonm: good catch, thank you!

isc-dchui (commented):
Oh one more thing, to get GitHub to auto-link the issue, you must put the issue number immediately after the [magic closing word](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests), e.g. `Closes #1152`. ("Addresses" is not one of these magic words unfortunately.)

Of course, you're free to manually link the issue instead.

src/cls/IPM/Storage/Module.cls
```diff
@@ -1130,7 +1132,7 @@ Method ProcessSingleDependencyIterative(
         set serverName = ""
         set version = ""
         if $data(pDependencyGraph(pDep.Name),depInfo) {
-            set $listbuild(previousDepth,serverName,version) = depInfo
+            set $listbuild(previousDepth,serverName,version,scope) = depInfo
```
isc-dchui: Just noticed this, but it looks like we're missing index 4 (display name) here?

isc-kiyer: Doesn't seem like scope is getting used so it shouldn't be needed either here. Can just keep the first 3

isc-cborbonm: good point! thank you

src/cls/IPM/Storage/Module.cls
```diff
@@ -1204,7 +1206,7 @@ Method ProcessSingleDependencyIterative(
             }
 
             set pDependencyGraph(pDep.Name) = $listbuild(
-                pDepth, qualifiedReference.ServerName, moduleObj.VersionString, qualifiedReference.Deployed, qualifiedReference.PlatformVersion,pDep.DisplayName
+                pDepth, qualifiedReference.ServerName, moduleObj.VersionString, qualifiedReference.Deployed, qualifiedReference.PlatformVersion,pDep.DisplayName,pDep.Scope
```
isc-dchui: Wait, why does this setting of pDependencyGraph have an extra argument of deployed in there? I know this predates your change, but I'm a bit confused by it.

isc-cborbonm: absolutely no clue 😃

isc-dchui: So I asked Claude to do some investigative journalism and here's what it found:

> ## Dependency graph format inconsistency in `BuildDependencyGraph`
> 
> The `pDependencyGraph` array written by `BuildDependencyGraph` has three different
> entry formats depending on which code path populated it:
> 
> | Path | `$list(data,1)` | `$list(data,2)` | `$list(data,3)` | `$list(data,4)` | `$list(data,5)` | `$list(data,6)` |
> |---|---|---|---|---|---|---|
> | Already installed (line 1145) | depth | `""` | version | DisplayName | — | — |
> | Exact-match from repo (line 1206) | depth | serverName | version | **Deployed** | **PlatformVersion** | DisplayName |
> | Fuzzy-match from repo (line 1241) | depth | serverName | version | DisplayName | — | — |
> 
> ### Root cause
> 
> This is a remnant of the iterative BFS rewrite in #975. The exact-match path was
> ported from the old recursive implementation, which embedded `Deployed` and
> `PlatformVersion` in the graph for downstream consumption. The new iterative design
> no longer reads those positions back out of the graph (the inverted graph pipeline
> only reads positions 2–3), so they became dead weight — but the format was never
> cleaned up, leaving `DisplayName` at position 4 on two paths and position 6 on the
> third.
> 
> ### Current impact
> 
> No active bug: `ConstructInvertedDependencyGraph` only reads `$list(data, 2, 3)`, so
> nothing downstream is broken today. However, any code added to read `DisplayName`
> from the graph would silently get a `Deployed` boolean (`0`/`1`) for exact-match
> dependencies.

@isc-kiyer what's your sense of the best thing to do here? This path-dependent inconsistency smells like a footgun and we should probably fully normalize it. Right now the caller has to know which path has been taken to know what the actual format will be.

isc-kiyer: @isc-dchui @isc-cborbonm  Yeah I agree we should normalize it. I don't know anything downstream using the platform version and deployed flags but things are using display name. I would vote to remove those

isc-cborbonm: thanks @isc-kiyer ! updated the review to delete platform version and deployed flags

isc-dchui (approved):
Looks good! Thanks for addressing all the issues that popped up!

isc-jili (approved):

PR #1193, opened 2026-06-05
Auto-detect the configured Python version and look in PATH as a fallback on Linux

### LINKED ISSUE #1191: DetectPipCaller does not detect python executable with version
A common Python layout on Linux is e.g.:
/usr/lib64
-rwxr-xr-x   1 root  root    7131424 Aug 21  2025 libpython3.12.so.1.0
drwxr-xr-x  34 root  root       8192 Mar 20 15:43 python3.12

With PythonRuntimeLibrary set to /usr/lib64/libpython3.12.so.1.0

%IPM.Lifecycle.Base:DetectPipCaller looks for an executable named "python" in the directory of PythonRuntimeLibrary. We should look for the specific version too - e.g. $Piece(PythonRuntimeLibrary,".so.1.0") if not empty.

Without this, it's a requirement to either:
set ^IPM.settings("PipCaller") = "/bin/python3.12"
set ^IPM.settings("UseStandalonePip") = 0
 
Or install IPM, run:
zpm "config set PipCaller /usr/lib64/python3.12"
zpm "config set UseStandalonePip 0"

Then nuke and reinstall. (Which are both gross options.)

isc-dchui:
## Description
- Fixes #1191 
- Looks for the Python executable associated with PythonRuntimeLibrary config value by searching `/usr/bin/`, `/usr/lib/`, and PATH
- Only handles Linux but setting PipCaller IPM config will still work on Windows and Mac

## Testing
New unit test for the version name parsing. Tested the intermediate logic with a Python 3.11 installation in the container.

### DIFF
--- a/src/cls/IPM/Lifecycle/Base.cls
+++ b/src/cls/IPM/Lifecycle/Base.cls
@@ -816,6 +816,9 @@ ClassMethod ResolvePipCaller(ByRef pParams) As %List
     return ..DetectPipCaller(tUseStandalonePip, $get(pParams("Verbose"), 0))
 }
 
+/// Auto-detects a pip caller when PipCaller is not configured. Tries, in order:
+/// flexible Python (PythonRuntimeLibrary), irispip.exe (Windows),
+/// python3/python in PATH, pip3/pip in PATH. Throws if nothing is found.
 ClassMethod DetectPipCaller(
 	pUseStandalonePip As %Boolean,
 	pVerbose As %Boolean = 0) As %List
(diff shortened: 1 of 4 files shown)


src/cls/IPM/Lifecycle/Base.cls
```diff
@@ -917,7 +901,128 @@ ClassMethod DetectPipCaller(
...
+                return path
+            }
+        }
+    }
+    return ""
+}
+
+/// Attempts to find a pip caller using PythonRuntimeLibrary from iris.cpf (IRIS 2024.2+).
+/// Returns a $list command suitable for RunCommand, or "" if unavailable or not found.
+ClassMethod DetectFlexiblePythonCaller(useStandalonePip As %Boolean, verbose As %Boolean = 0) As %List
```
isc-kiyer, line 929: A lot of this feels quite brittle for future proofing. Is the python runtime not something we can just always get from the CPF settings/ask the user to always set those? Beyond that, why do we need to override it in IPM at all?

isc-dchui: The problem is that the Python runtime setting says where the library is, but not where the Python executable is. That could be in any one of multiple places on Linux: /usr/lib/ or /usr/lib64/ or /usr/bin/ or /usr/local/bin/ or anywhere else in PATH.

isc-dchui: For more context, the only time we directly use this Python executable is to call the attached version of pip to either install wheels or install from the requirements.txt file.

isc-kiyer (approved):

PR #1194, opened 2026-06-08
Fix dependency resolution bug, semver intersection and bug and have better error message for dependency incompatibilities

### LINKED ISSUE #1077: Request: better error message for semantic versioning incompatibility
HS recently encountered the following error message in the logs of our CI pipeline:

```
ERROR #5001: No overlapping versions in AND operation for >=6.5.0 <7.0.0-0 and >=7.0.0 <8.0.0-0
```

That particular pipeline run loaded six modules and all of their dependencies. This error message was not thrown in the course of any lifecycle phase but rather, I suspect, from `%IPM.Storage.Module::BuildDependencyGraph()`. Without a pointer to a module, it is hard to address this.

Request: add to the semantic versioning error messages so that they include the name of the module in question (and ideally some form of dependency tree too, but that's a nice-to-have).

### LINKED ISSUE #1192: Dependency resolution will ignore incompatible dependencies if compatible version found after incompatible
## Describe the bug
When the refactor of BuildDependencyGraph() was done in 0.10.5 to improve performance, there was a regression introduced. It would over ambitiously ignore resolution failures expecting a higher stack level to hit the failure but with the refactor, there is no higher stack level since the resolution is done iteratively instead of using recursion. This results in incompatibilities possibly getting ignored. For example if modules A and B both depend on C but A depends on ^1.1.0 and B depends on ^2.0.0, this is not caught if multiple further modules say have ^1.1.0 and are resolved after B. In that case, the incompatibility from B is dropped.

## To Reproduce
1. Construct test modules where one module installs many other modules. Each of those modules has a shared dependency but they have incompatible dependency versions.
2. Run zpm "install < basemodule>"
3. See that no error is thrown and the version of the most dependent module that is installed is the version that matches the most dependents instead of throwing an error.

## Expected behavior
Better error handling. Add good integration tests.

## System information (please complete the following information):
 - IPM version: 0.10.7

isc-kiyer:
## Description
- Fixes #1192
- Fixes #1077 

## Testing
- Added unit tests for semver bug
- Added integration tests for dependency resolution

## Checklist

- [x] Changelog entry added.


src/cls/IPM/Storage/Module.cls
```diff
@@ -1289,6 +1293,30 @@ Method ProcessSingleDependencyIterative(
...
+/// Combines two version expressions with AND.
+/// Throws a descriptive error listing all conflicting requirements if the expressions are incompatible.
+/// <p>
+/// <b>Parameters:</b><br>
+/// <var>pSearchExpr</var> - Accumulated version expression to AND into<br>
+/// <var>pNewExpr</var> - New version expression to AND with<br>
+/// <var>pDepName</var> - Name of the dependency being resolved (for error message)<br>
+/// <var>pConstraintList</var> - %List of "source: constraint" strings accumulated so far (for error message)<br>
+ClassMethod AndConstraint(
+	pSearchExpr As %IPM.General.SemanticVersionExpression,
```
isc-cborbonm: nit: I thought that for brand new methods we were trying to phase out the "p" prefix out of parameter names

isc-cborbonm (dismissed):
slay test cases! 😃

tests/integration_tests/Test/PM/Integration/BuildDependencyGraph.cls
```diff
@@ -1,11 +1,528 @@
...
+    set $listbuild(tDepth) = tGraph("complex-e")
+    do $$$AssertEquals(tDepth, 3, "complex-e at depth 3.")
+}
+
+/// Builds the graph for complex-root 2.0.0, which requires complex-b ^2.0.0.
+/// complex-b 2.0.0 requires complex-c ^3.0.0 while complex-a 1.0.0 requires complex-c ^2.0.0.
+/// The ^2.0.0 AND ^3.0.0 constraint combination is unsatisfiable; verifies that the error
+/// message names complex-c and identifies both complex-a and complex-b as the conflicting sources.
+/// Test data: build-dependency-graph/ (complex-root 2.0.0).
+Method TestComplexIncompatibleDiamondConflicts()
```
isc-dchui: Seems like this case is already being covered by TestIncompatibleTransitiveConstraintsThrowDescriptiveError and is redundant

src/cls/IPM/Storage/Module.cls
```diff
@@ -1289,6 +1293,30 @@ Method ProcessSingleDependencyIterative(
...
 
+/// Combines two version expressions with AND.
+/// Throws a descriptive error listing all conflicting requirements if the expressions are incompatible.
+/// <p>
+/// <b>Parameters:</b><br>
+/// <var>pSearchExpr</var> - Accumulated version expression to AND into<br>
+/// <var>pNewExpr</var> - New version expression to AND with<br>
+/// <var>pDepName</var> - Name of the dependency being resolved (for error message)<br>
+/// <var>pConstraintList</var> - %List of "source: constraint" strings accumulated so far (for error message)<br>
+ClassMethod AndConstraint(
```
isc-dchui: This is essentially a wrapper for And() that provides better error messages right? Should we also migrate the other callers of the old And() method?

tests/integration_tests/Test/PM/Integration/BuildDependencyGraph.cls
```diff
@@ -1,11 +1,528 @@
...
+}
+
+Method OnAfterOneTest() As %Status
+{
+    // Uninstall all installed modules and repo
+    set sc = ##class(%IPM.Main).Shell("uninstall -all")
+    do $$$AssertStatusOK(sc, "Uninstalled all modules successfully.")
+
+    set sc = ##class(%IPM.Main).Shell("repo -delete -name "_..#REPONAME)
+    do $$$AssertStatusOK(sc, "Deleted "_..#REPONAME_" repo successfully.")
```
isc-dchui: Out of curiosity, why do we need to create and delete the repository after each test?

isc-kiyer: Good catch! Just the uninstall of modules needs to be after each test. I'll move repo set up and teardown to before/after all tests

src/cls/IPM/Storage/Module.cls
isc-kiyer: Hmm good point. There aren't many other callers of it so I'll just enhance the original And() method

isc-cborbonm (dismissed):
looks good to me :)

isc-jili (dismissed):
Looks good to me!

isc-dchui (commented):
Looks like CI needs some fixing:
1) 2024.3 and 2025.1 have expired licenses so you'll need to remove them from the list with something like what I did here:
https://github.com/intersystems/ipm/pull/1193/changes#diff-11afbb4eae27849c9f06ecefe10854fd951a54476c5f9ddc0284c0d7f0ea0c6c

2) "Clean install current version and test new one" is timing out so you'll need to extend "timeout-minutes" in .github/workflows/main.yml

Otherwise looks good!

isc-dchui (approved):

isc-cborbonm (approved):

PR #1211, opened 2026-07-29
change ShellScript() to return a status

### LINKED ISSUE #1186: Ability to write all output from any IPM shell command directly to a file instead of the current device
## Describe the problem
When using IPM shell commands with CPF merge, CPF merge does not provide the current device write outputs in a meaningful manner so it makes sense to persist all output to the current device in a log file. This would also be generally useful for devs to share outputs of any IPM commands with each other without needing to copy and paste it from terminal.

## Describe the solution
A global flag (like -verbose) with a place to export the current device write outputs to.
Additionally, it would be good to have a persistent modifier amongst the universal settings of a default location to write log files to (we can default this to somewhere under the DataDirectory()/ipm?). 

Also consider adding output of commands to the history table so we can see the log tied to specific command runs.

## Additional context
Requested in internal reference TRIAGE-6911

isc-cborbonm:
Resolves #1186: change ShellScript() to return a status.

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -4074,7 +4074,7 @@ ClassMethod ShellScript(
 	pCommand As %String,
 	pOutputLogFile As %String = "",
 	pAppendToLog As %Boolean = 0,
-    pterminateOnError As %Boolean = 1) As %Status
+    pTerminateOnError As %Boolean = 1) As %Status
 {
     set tSC = $$$OK
     set tFileOpen = 0
@@ -4106,7 +4106,7 @@ ClassMethod ShellScript(
         close pOutputLogFile
     }
 
-    if (pterminateOnError && $$$ISERR(tSC)) {
+    if (pTerminateOnError && $$$ISERR(tSC)) {
         // Set ERRORLEVEL to 1
         do $system.Process.Terminate(,1)
     }
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -4073,7 +4073,8 @@ ClassMethod History(ByRef pCommandInfo)
 ClassMethod ShellScript(
 	pCommand As %String,
 	pOutputLogFile As %String = "",
-	pAppendToLog As %Boolean = 0) As %Status
+	pAppendToLog As %Boolean = 0,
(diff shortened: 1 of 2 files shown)


src/cls/IPM/Main.cls
```diff
@@ -4105,10 +4105,7 @@ ClassMethod ShellScript(
         close pOutputLogFile
     }
```
isc-kiyer, line 4108: So this is a backwards incompatible change. You should have an arg to determine whether the process should terminate or not and default to 1. Then when its not 1, we can return status

isc-cborbonm: oooh okay I'll do that

isc-eneil (dismissed):

src/cls/IPM/Main.cls
```diff
@@ -4073,7 +4073,8 @@ ClassMethod History(ByRef pCommandInfo)
 ClassMethod ShellScript(
 	pCommand As %String,
 	pOutputLogFile As %String = "",
-	pAppendToLog As %Boolean = 0)
+	pAppendToLog As %Boolean = 0,
+    terminateOnError As %Boolean = 1) As %Status
```
isc-kiyer: Nit: when editing an existing method, stick to the style of that method (use the p prefix here)

isc-jili (approved):

isc-dchui (commented):
Since this PR is linked to #1186, could you also please spawn off a new issue about tee-ing the IPM output in general?

isc-cborbonm:
> Since this PR is linked to #1186, could you also please spawn off a new issue about tee-ing the IPM output in general?

Yeah! I could also unlink it, so that it doesn't close the issue. Would that be preferable?

isc-dchui:
@isc-cborbonm It depends on the scope. If this change fixes a legitimate distinct problem, which it sounds like it does, then I think it should have an issue attached.

isc-cborbonm:
> @isc-cborbonm It depends on the scope. If this change fixes a legitimate distinct problem, which it sounds like it does, then I think it should have an issue attached.

Created new issue https://github.com/intersystems/ipm/issues/1213 to implement the teeing output feature! 😃

isc-dchui (approved):

PR #1221, opened 2026-08-13
fix: installer runs synchronously

### LINKED ISSUE #1215: In IPM installer replace async setup with sync setup
## Describe the problem
Async setup seems like more trouble than its worth because it messes with the logging order. 

## Describe the solution
Replace it with a synchronous setup.

isc-tleavitt:
Fixes #1215 

From working session with Dominic today. This seems to... just work? Maybe? (We'll see what CI says.)


isc-kiyer:
@isc-tleavitt Seems to not work? Unless CI tests need some updates?
cc @isc-dchui

PR #1223, opened 2026-08-17
Various sync fixes from testing with a real (messy) package

### LINKED ISSUE #1222: Sync command fails with module.xml <SourcesRoot>.</SourcesRoot>
v0.10.10-SNAPSHOT (current latest main branch):
I've been trying out the new zpm "sync" command (via Claude) and running into some issues which stem from my module.xml setting SourcesRoot to "." - I think this is due to some past bug/issue and maybe isn't necessary so will investigate a fix on my end in parallel.

Disclaimer, the following came from Claude Code (a major intended audience for zpm "sync"):

>   The chain:
> 
>   1. module.xml:10 sets <SourcesRoot>.</SourcesRoot>.
>   2. %IPM.ResourceProcessor.Default.Package:64-67 (and Default.Document:507-512) build relative paths as SourcesRoot _ "/" _ Directory _ "/" _ item, so with
>   SourcesRoot="." every path comes out ./-prefixed:
>     - OnItemRelativePath("pkg.isc.genai.tools.file.Utils.cls") → ./cls/pkg/isc/genai/tools/file/Utils.cls
>     - GetSyncDirectory() → ./cls
>   3. %IPM.Storage.FileHash.NormalizePath:374-384 translates \→/, collapses //, strips leading / — but does not collapse ./. Verified: ./cls/…/Utils.cls survives
>   normalization, and NormalizePath("./cls/…") = NormalizePath("cls/…") → 0.
>   4. The disk walk (WalkAndHashDirs) and the hash baseline key paths root-relative with no ./: of 1148 stored paths, 0 start with ./ and 904 start with cls/.
>   5. So in SyncBuildReverseIndex both mapping routes miss: the per-child mapping writes reverseIndex("./cls/…"), which no changed path ever equals; and the
>   directory prefix-scan compares prefix ./cls/ against allFiles keys of the form cls/…, matching nothing.
>   6. The storedPaths fallback that exists precisely for this is defeated by a second, independent mismatch: RelPathToDocName("cls/…/Utils.cls") →
>   pkg.isc.genai.tools.file.Utils.**CLS**, while ResolveChildren keys (hence docToResource) are pkg.isc.genai.tools.file.Utils.**cls**. Exact-subscript lookup →
>   miss.
>   7. ComputeChanges still flags the file (it has a baseline row). SyncRoutePathSet:341-343 silently continues on an unrouted path → syncByResource empty →
>   Dispatching to 0 resource(s) → loadItems=0 → step 7 skips SyncCompile. Step 9 nonetheless runs CommitChanges and prints Updated: cls/…/Utils.cls, then
>   MarkSyncOK.
> 
>   Net effect: the edit is never loaded, is reported as synced, and its hash is committed — so the next sync says "Nothing to sync" and it's dropped permanently.
> 
>   The contrast confirms it: the resources that genuinely sync in this module are the six /internal/testing/unit_tests ones, whose path-style names bypass the
>   SourcesRoot prefix and return GetSyncDirectory() = internal/testing/unit_tests — no ./, so they route. That's exactly the set of files we saw actually update.
> 
>   Fixes, in priority order
> 
>   1. NormalizePath should collapse ./ (strip a leading ./ and any /./ segment). One-place fix that repairs both the per-child mapping and the prefix scan for every
>   module with SourcesRoot="." — likely a common manifest idiom.
>   2. Canonicalize document names before using them as subscripts (uppercase the extension on both sides, or route both through one canonicalizer) so the
>   storedPaths fallback works.
>   3. Fail loudly: a modified path that routes to no resource must not print Updated: and must not have its hash committed — it should emit a warning and be
>   re-detected next sync. Defects 1 and 2 were only catastrophic because of this; a warning here would have made it a two-minute diagnosis instead of a false green.

### LINKED ISSUE #1224: Sync doesn't detect and reload on revert of failed code
## Describe the bug
zpm "sync" reports "nothing to sync" rather than reloading a corrected class.

## To Reproduce
1. Break a class definition that's part of a dev-mode package (e.g., add an extra trailing } after a method)
2. zpm "sync" - should report an error
3. Undo the breaking change
4. zpm "sync" again - reports "Nothing to sync."

## Expected behavior
The second zpm "sync" should reload + recompile the modified item.

## System information
 - IPM version: 0.10.10-SNAPSHOT

isc-tleavitt:
Fixes #1222 
Fixes #1224 
Fixes unfiled issue where running sync -test on a package with multiple unit test resources would run the synced test once per resource. (Ignoring Package filter on unit test resources.)


isc-tleavitt:
I've verified this fixes the issue for my target package. The fix is clean and well-tested.

isc-tleavitt:
Switching back to draft, going to fix #1224 and whatever else I find too.

isc-tleavitt:
I think this is good now and I'd like to get it merged to better have sync available for experimentation.

isc-tleavitt:
Actually intentionally skipping the changelogs here since they're fixes to an unreleased feature.

src/cls/IPM/General/Sync/Pipeline.cls
```diff
@@ -130,7 +147,25 @@ ClassMethod Run(
...
         // Done before running tests below so a test failure doesn't prevent the file-sync
         // outcome from being committed and reported — file sync and test results are independent.
-        $$$ThrowOnError(##class(%IPM.Storage.FileHash).CommitChanges(module, .modified, .deleted, processDeletes))
+        $$$ThrowOnError(##class(%IPM.Storage.FileHash).CommitChanges(module, .modified, .deleted, processDeletes, runTests))
+        // Past this point the change set is recorded as synced, so the error handler below must not
+        // flag it for retry: anything that fails from here on (a failing test in step 10) says
+        // nothing about whether the files landed.
+        set committed = 1
+
+        // Committing the hashes above means change detection will never look at these files again,
```
isc-dchui: Nit: I tend to find comments describing what isn't done to be confusing and would prefer just what is being done and why. Essentially these first two lines are a bit unneeded

src/cls/IPM/General/Sync/Pipeline.cls
```diff
@@ -170,6 +211,17 @@ ClassMethod Run(
 
     } catch e {
         set sc = e.AsStatus()
+        // Step 9 never ran, so no hash was committed and the baseline still describes the last
+        // successfully-synced content — while the namespace holds whatever this attempt left behind.
+        // Fixing the file by reverting it would hash back to the baseline and go unnoticed, so flag
+        // what was being applied and let the next sync force it through (issue #1224).
```
isc-dchui: Nit: remove reference to specific intra-release issue since it just adds clutter

src/cls/IPM/Storage/FileHash.cls
```diff
@@ -17,6 +17,20 @@ Property PathHash As %String(MAXLEN = 40) [ Required ];
...
+/// failed attempt left behind. Reverting the file to that content would otherwise make it invisible
+/// to change detection; this flag keeps it in the change set until a sync commits it. Rows written
+/// before this property existed read as "", which is correctly falsy.
+Property SyncFailed As %Boolean [ InitialExpression = 0 ];
+
+/// Set when a sync applied this path but did not run its tests — a plain <code>sync</code> over a
+/// file owned by a test resource that runs in the "test" phase. The sync commits the new hash, so
+/// change detection would never look at that file again and the edit would stay untested; this flag
+/// keeps it in the change set until a sync with -test runs its tests. Rows written before this
+/// property existed read as "", which is correctly falsy.
```
isc-dchui: Nit: seems silly to reference backwards compatibility with itself in a feature that hasn't been released yet for both the SyncFailed and TestPending comments

isc-dchui (commented):
Good bug catches! Looks good overall but I tend to find LLM comments/documentation to be excessively verbose and off in a particular backwards-looking describing-what-used-to-happen way as opposed to a describing what is currently implemented way, if that makes sense.

I usually just ask Claude to write comments with concision in mind and make it undergo a verbosity check pass.

src/cls/IPM/General/Sync/Pipeline.cls
isc-tleavitt: Rewritten to state what happens: "The hashes are committed, so flag the test-owned paths to keep them in the change set: the next sync with -test forces them through and runs their tests (see FileHash's TestPending property)."

src/cls/IPM/General/Sync/Pipeline.cls
isc-tleavitt: Removed the issue reference.

src/cls/IPM/Storage/FileHash.cls
isc-tleavitt: Dropped the "rows written before this property existed" sentence from both SyncFailed and TestPending.

isc-tleavitt:
Did a verbosity pass over the comments this PR adds, cutting the "here's what would otherwise happen" framing in favor of describing current behavior. Trimmed:

- `Pipeline.cls`: the `ComputeChanges` out-param note, the unrouted-paths note, the TestPending flag block, the deferred-test warning, and `SyncRunTests` step 2 / `SyncCollectTestPaths` docs
- `FileHash.cls`: Pass 3/4/5 comments, `NormalizePath`'s "." explanation, and the `SyncFailed`/`TestPending` property docs
- `Test.cls`: the sibling-package compile note, the testspec grammar note, and `OnSyncRunTests`' doc

Comment-only — `git diff -U0 | grep -v '^[+-]\s*//'` is empty over `src/`. f05afa9

isc-tleavitt:
> Did a verbosity pass over the comments this PR adds, cutting the "here's what would otherwise happen" framing in favor of describing current behavior. Trimmed:
> 
> * `Pipeline.cls`: the `ComputeChanges` out-param note, the unrouted-paths note, the TestPending flag block, the deferred-test warning, and `SyncRunTests` step 2 / `SyncCollectTestPaths` docs
> * `FileHash.cls`: Pass 3/4/5 comments, `NormalizePath`'s "." explanation, and the `SyncFailed`/`TestPending` property docs
> * `Test.cls`: the sibling-package compile note, the testspec grammar note, and `OnSyncRunTests`' doc
> 
> Comment-only — `git diff -U0 | grep -v '^[+-]\s*//'` is empty over `src/`. [f05afa9](https://github.com/intersystems/ipm/commit/f05afa97b0be48ed6465db6e992cf431fcb9b720)

This was Claude on auto mode... experimenting here, clearly.

isc-tleavitt:
Interesting - Claude wasn't using AGENTS.md which directs properly here. I'm tacking on a fix for that too.

isc-dchui (approved):

PR #1235, opened 2026-09-02
Fix two issues around uninstalling IPM

### LINKED ISSUE #1227: After uninstalling and reinstalling IPM, IPM commands will display "IPM has been uninstalled from this namespace. Exiting shell..."
## Describe the bug
After uninstalling and reinstalling IPM, IPM commands will display "IPM has been uninstalled from this namespace. Exiting shell..." and then quit the shell. You can re-enter the shell using "zpm" though.

## Additional context
This is from the ^||%IPM.SelfUninstallFlag process-private global not being reset on install. Starting a new process is the workaround.

### LINKED ISSUE #1228: Unable to fully uninstall IPM unless another module is present
## Describe the bug
PR #1106 added the ability to uninstall IPM and also remove the globals tracking which version has been installed. This is required for downgrading IPM. However, this prompt only appears if a non-IPM module is also installed.

```
zpm:USER>uninstall zpm

The following modules are currently installed:
  - objectscript-math (0.0.5)

Do you want to fully remove all modules and IPM metadata (except history log)?
This is required for downgrading IPM. [y/N]:
```

## To Reproduce
1. Install a version of IPM like 0.10.9
2. Uninstall IPM
3. Install IPM 0.10.8 and see how it fails with the warning against downgrades
4. Reinstall IPM 0.10.9
5. Install a different module
6. Uninstall again and see how the prompt for deleting IPM metadata appears: say yes
7. Install IPM 0.10.8

## Expected behavior
The IPM metadata prompt should appear regardless of whether another module is installed.

isc-dchui:
## Description
- Fixes #1227
- Fixes #1228

## Testing
Manually tested by trying to uninstall IPM without another module installed and the prompt to delete metadata appeared. Reinstalling IPM in the same process also allowed me to use it without being kicked out of the shell after every command

## Checklist

- [x] Changelog entry added.

### DIFF
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -1124,6 +1124,8 @@ ClassMethod ShellInternal(
 
         // Check if IPM has uninstalled itself - exit gracefully if so
         if $get(^||%IPM.SelfUninstallFlag, 0) {
+            // Process-private, so clear it once acted on to keep later shells in this process working.
+            kill ^||%IPM.SelfUninstallFlag
             write !,!,$$$FormattedLine($$$Yellow, "IPM has been uninstalled from this namespace. Exiting shell..."),!
             return
         }
@@ -2662,29 +2664,26 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
         if (tModuleName = $$$IPMModuleName) {
             $$$ThrowOnError(..CheckModuleNamespace())
 
-            // Check if other modules are installed
-            set hasModules = 0
+            // List the other modules that a full cleanup would remove
             set rs = ##class(%SQL.Statement).%ExecDirect(,
                 "SELECT Name, VersionString FROM %IPM_Storage.ModuleItem WHERE Name != ?",
                 $$$IPMModuleName)
-            if rs.%Next() {
-                set hasModules = 1
-
-                // Show list of installed modules
-                write !,"The following modules are currently installed:"
+            set hasModules = 0
+            while rs.%Next() {
+                if 'hasModules {
+                    write !,"The following modules are currently installed:"
+                    set hasModules = 1
+                }
                 write !,"  - ",rs.Name," (",rs.VersionString,")"
-                while rs.%Next() {
-                    write !,"  - ",rs.Name," (",rs.VersionString,")"
-                }
-                write !
+            }
+            write:hasModules !
 
-                if 'tForce {
-                    // Dialog: Ask about full cleanup (downgrade scenario)
-                    write !,"Do you want to fully remove all modules and IPM metadata (except history log)?"
-                    write !,"This is required for downgrading IPM. [y/N]: "
-                    read fullCleanupResponse
-                    set tParams("Clean","FullCleanup") = $case($zconvert(fullCleanupResponse,"L"),"y":1,"yes":1,:0)
-                }
+            if 'tForce {
+                // Dialog: Ask about full cleanup (downgrade scenario)
+                write !,"Do you want to fully remove all modules and IPM metadata (except history log)?"
+                write !,"This is required for downgrading IPM. [y/N]: "
+                read fullCleanupResponse
+                set tParams("Clean","FullCleanup") = $case($zconvert(fullCleanupResponse,"L"),"y":1,"yes":1,:0)
             }
 
             if $get(tParams("Clean","FullCleanup")) {
@@ -2693,7 +2692,7 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
                 $$$ThrowOnError(##class(%IPM.Utils.Module).UninstallAll(tForce,.tParams))
             }
 
-            // Set flag to indicate self-uninstall - will be checked by Interactive() loop
+            // Set flag to indicate self-uninstall - will be checked by ShellInternal() loop
             set ^||%IPM.SelfUninstallFlag = 1
         }
         set tRecurse = $$$HasModifier(pCommandInfo,"recurse") // Recursively uninstall unneeded dependencies
--- a/src/cls/IPM/Main.cls
+++ b/src/cls/IPM/Main.cls
@@ -2668,22 +2668,23 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
             set rs = ##class(%SQL.Statement).%ExecDirect(,
                 "SELECT Name, VersionString FROM %IPM_Storage.ModuleItem WHERE Name != ?",
                 $$$IPMModuleName)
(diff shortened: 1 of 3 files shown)


src/cls/IPM/Main.cls
```diff
@@ -2662,29 +2664,26 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
         if (tModuleName = $$$IPMModuleName) {
             $$$ThrowOnError(..CheckModuleNamespace())
 
-            // Check if other modules are installed
-            set hasModules = 0
+            // List the other modules that a full cleanup would remove
             set rs = ##class(%SQL.Statement).%ExecDirect(,
```
isc-kiyer, line 2668: Should have `$$$ThrowSQLIfError(rs.%SQLCODE,rs.%Message)`

src/cls/IPM/Main.cls
```diff
@@ -2662,29 +2664,26 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
...
-                }
-                write !
-
-                if 'tForce {
-                    // Dialog: Ask about full cleanup (downgrade scenario)
-                    write !,"Do you want to fully remove all modules and IPM metadata (except history log)?"
-                    write !,"This is required for downgrading IPM. [y/N]: "
-                    read fullCleanupResponse
-                    set tParams("Clean","FullCleanup") = $case($zconvert(fullCleanupResponse,"L"),"y":1,"yes":1,:0)
+            set hasModules = 0
```
isc-kiyer, line 2673: You can check this by checking %SQLCODE. if its 100, then there were no records returned

src/cls/IPM/Main.cls
```diff
@@ -2662,29 +2664,26 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
...
+            while rs.%Next() {
```
isc-kiyer: Should check status here and $$$ThrowOnError(sc) outside the loop

src/cls/IPM/Main.cls
```diff
@@ -2662,29 +2664,26 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
...
+                    set hasModules = 1
                 }
+                write !,"  - ",rs.Name," (",rs.VersionString,")"
+            }
+            write:hasModules !
+
+            if 'tForce {
+                // Dialog: Ask about full cleanup (downgrade scenario)
+                write !,"Do you want to fully remove all modules and IPM metadata (except history log)?"
+                write !,"This is required for downgrading IPM. [y/N]: "
```
isc-kiyer: Use %Library.Prompt here so default answer can be provided (default should be no?)

isc-kiyer (changes requested):
@isc-dchui few small notes

src/cls/IPM/Main.cls
isc-dchui: Not sure I fully understand this. Isn't %SQLCODE=100 if either there are no records or all the records have been iterated through? The conditional write here only cares about the former and not the latter

isc-kiyer: Correct. However, when no %Next() has been called, if SQLCODE is 100, it means no records were returned by the query (or else you would get 0 as the return value)

isc-dchui: Hm, I was under the impression that SQLCODE before %Next() is called just determines if the query executed successfully, not if there are any results.

> The %Execute() method sets the [%SQL.StatementResult](https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25SQL.StatementResult) class properties [%SQLCODE](https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25SQL.StatementResult#PROPERTY_%25SQLCODE) and [%Message](https://docs.intersystems.com/irislatest/csp/documatic/%25CSP.Documatic.cls?LIBRARY=%25SYS&CLASSNAME=%25SQL.StatementResult#PROPERTY_%25Message) for all SQL statements. Successful execution of the statement sets %SQLCODE to 0. This does not mean that the statement successfully retrieved results. Similarly, %Execute() does not set %SQLCODE to 100 if the statement retrieves no results. The check for results, and subsequent setting of %SQLCODE to 0, 100, or a negative error value occurs as you fetch the results one row at a time, such as by using the %Next() method.

https://docs.intersystems.com/irislatest/csp/docbook/DocBook.UI.Page.cls?KEY=GSQL_dynsql#GSQL_dynsql_executemethod

src/cls/IPM/Main.cls
```diff
@@ -1124,6 +1124,8 @@ ClassMethod ShellInternal(
 
         // Check if IPM has uninstalled itself - exit gracefully if so
         if $get(^||%IPM.SelfUninstallFlag, 0) {
+            // Process-private, so clear it once acted on to keep later shells in this process working.
```
isc-kiyer, line 1127: Would be nice to add a test for this case

src/cls/IPM/Main.cls
```diff
@@ -2662,29 +2664,27 @@ ClassMethod Uninstall(ByRef pCommandInfo) [ Internal ]
...
 
-                // Show list of installed modules
-                write !,"The following modules are currently installed:"
-                write !,"  - ",rs.Name," (",rs.VersionString,")"
-                while rs.%Next() {
-                    write !,"  - ",rs.Name," (",rs.VersionString,")"
+            set hasModules = 0
+            while rs.%Next(.sc) {
+                if 'hasModules {
+                    write !,"The following modules are currently installed:"
```
isc-kiyer, line 2676: We should probably add a line indicating no modules currently installed if there are none present and if so, do not need to prompt right?

isc-kiyer (commented):
@isc-dchui 2 minor notes

src/cls/IPM/Main.cls
isc-dchui: It would be, but it's unclear to me how we could add a test for uninstalling IPM itself

src/cls/IPM/Main.cls
isc-dchui: The prompt is for deleting all the IPM metadata and the bug this PR is trying to fix is that currently the prompt doesn't appear if there's no modules installed and thus no way of deleting the metadata.

I'm not sure adding a line saying there are no modules installed is particularly useful. It's there when there are modules because deleting IPM metadata will also delete metadata for those modules. If a user chooses not to delete the metadata, reinstalling IPM will automatically adopt those modules again.

isc-kiyer (approved):

PR #1243, opened 2026-09-14
Logging framework api skeleton

isc-cborbonm:
Adds the initial Log() method skeleton to IPM.General.LogManager.

Validates the supplied log level against supported levels.
Emits an error for invalid levels.
Leaves a TODO for routing messages to namespace-specific log files.


src/cls/IPM/General/LogManager.cls
```diff
@@ -3,13 +3,28 @@ Include %IPM.Common
...
+
+Method Log(
+    message As %String,
+    level As %String)
+{
+    if '$listfind($listfromstring(..#logLevels), level) {
+        do ..Log("Log has incorrect logLevel "_level_". Please ensure log level is one of: "_..#logLevels_$$$NL_"Attempted message to log was: "_message, "ERROR")
+    }
+
+    // TODO: Add the message to the LogFile(s) going in that instance in the same namespace
```
isc-eneil, line 25: One more thing that you can already calculate is the current timestamp. 
Also, for the time being, you can also do a `Write timeStamp_" ["_level_"] "_message`. It will be helpful as we start calling the skeleton to know if it's correctly being called!

isc-eneil (changes requested):

src/cls/IPM/General/LogManager.cls
isc-cborbonm: done!

isc-eneil: I see the timestamp but not the write!

isc-cborbonm: so true I'll fix that!

isc-eneil (approved):

Issue #1240, opened 2026-09-10
IPM needs a structured logging framework for (at the very least) module deployment


isc-eneil:
## Link to the discussion thread
https://github.com/intersystems/ipm/discussions/1239

## Describe the problem
Currently, IPM logging consists of writing to standard output, which is redirected to the terminal when using the zpm shell. IPM also now has the history command which holds onto a timestamped trace of all IPM commands run in an ipm-enabled namespace. There is also the possibility of logging to a file via the %IPM.Main::ShellScript() method.

As @AshokThangavel noted in a few issues that he created (https://github.com/intersystems/ipm/issues/1048 https://github.com/intersystems/ipm/issues/1012), as IPM's functionality expands to include installation and updating of packages in production environments, it needs a built-in framework for structured logging to a persistent file that can be inspected.

### Overview
Extend the %IPM.General.LogManager class to support writing output to an unstructured log file intended for humans and a structured, AI-readable JSON log file. All log lines will be logged to both log files, in a different format. Logging to these two log files will occur by default on the `install` , `update` , and `uninstall` commands (to start) and can be turned off using the `-no-log` modifier. The log file path can be changed via the `-output-file` modifier. Log files for each of these command will be linked to the corresponding history log entry (used for auditing). Documentation will be added to make IPM logging APIs clear to developers of module update steps.

### User Flow

1. Run update -v myModule 2.1.1 inside of MODULE_NS namespace
2. Two logs will be created:
- /mgr/ipm-logs/MODULE_NS/update_myModule_1787775895.json
- /mgr/ipm-logs/MODULE_NS/update_myModule_1787775895.txt

update_myModule_1787775895.json will look something like:
```
{
	"ipm": 0.10.9,
	"schema": 1.0.0,
	"command" : "update -v myModule 2.1.1",
	"log" : [
	 	{ "level": "INFO", "message": "Running zpm \"update -v myModule 2.1.1\"", "timestamp": "1787775895"},
  		{ "level": "INFO", "message": "Preparing to update module myModule to version 2.1.1", "timestamp": "1787775909"},
	   	{ "level": "INFO", "message": "Loading from /home/irisowner/zpm/tests/integration_tests/Test/PM/Integration/_data/update-test/myModule-2.x/myModule/", "timestamp": "1787775910"},
	    { "level": "INFO", "message": "Load started on 07/09/2026 16:14:58", "timestamp": "1787775911"}, 
[...]

PR #1250, opened 2026-09-17
Log schema + log file creation

isc-eneil:
## Description
- #1240
- This PR includes simply the log file schema (addition of several properties to the %IPM.General.LogManager singleton), as well as two methods-- `InitializeLog()` to create two logs files and update the properties in the class, and `WriteLogLine()` to add a log line of a particular level to each log.
- Out-of-scope: integrating the log schema or logging into the rest of the IPM code base.

## Testing
Dev testing with Claude, pending further testing upon integration with testing branch.


ipm-log.schema.json
```diff
@@ -0,0 +1,66 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "$id": "https://raw.githubusercontent.com/intersystems/ipm/main/schemas/ipm-log/1.0.0",
+    "title": "IPM Log Record",
+    "description": "Schema (version 1.0.0) for a single record in an IPM JSON Lines (NDJSON) log file. Each line of a '.jsonl' log is one JSON document: the first line is a header record and every subsequent line is a log entry.",
```
isc-jlechtne: Nit: typo '.jsonl' instead of '.json'

ipm-log.schema.json
```diff
@@ -0,0 +1,66 @@
...
+        },
+        {
+            "$ref": "#/$defs/entry"
+        }
+    ],
+    "$defs": {
+        "header": {
+            "type": "object",
+            "description": "First record of a log file; identifies the IPM and log schema versions and the command being logged.",
+            "properties": {
```
isc-jlechtne, line 19: Should there be a `"schema"` property as well? In the design doc that's the third of the "headers" in each file (and one that I put in the tests to check for)

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,112 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
-    message As %String,
-    level As %String)
+/// Absolute path to the JSON Lines (NDJSON) log file
+Property JSONFilePath As %String;
+
+/// Absolute path to the TXT log file
+Property TxtFilePath As %String;
+
+/// Starting Unix timestamp for this log
+Property Timestamp As %Integer;
```
isc-jlechtne: Should have a `TimeStampStart` and `TimeStampEnd`

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,112 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
+
+/// The module object for this operation
+Property Module As %IPM.Storage.Module;
```
isc-jlechtne, line 31: Should this be a relationship/foreign key instead of a property? @isc-kiyer deferring this question to you since you know more about the IPM codebase and what type this should be

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,112 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
+
+/// Initialize log files and write headers.
+/// @Argument command       The command verb being logged (e.g., "install", "load", "update", "uninstall")
```
isc-jlechtne: Is command the entire string or just the action? This sounds like it's just the action, but the property description implies it's the whole string

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,112 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
+        // Get IPM version. GetVersion() returns per-module info in versionInfo("zpm") as $lb(namespace, version).
+        $$$ThrowOnError(##class(%IPM.Main).GetVersion("zpm", .versionInfo))
+        set ipmVersion = $listget($get(versionInfo("zpm")), 2)
+
+        // Write TXT header
+        set txtStream = ##class(%Stream.FileCharacter).%New()
+        set txtStream.Filename = ..TxtFilePath
+        do txtStream.WriteLine("IPM Version: " _ ipmVersion)
+        // TODO: Replace hardcoded schema version with global IPM log schema setting
+        do txtStream.WriteLine("Schema Version: 1.0.0")
```
isc-jlechtne: Make schema version a property

src/cls/IPM/General/LogManager.cls
isc-eneil: @isc-jlechtne This is a good question. My instinct is that, since the `%IPM.Storage.Module` object does not need any knowledge about the log, a two-way property type such as a Relationship is unnecessary. It might be necessary to add a relationship to a History log entry, however.

src/cls/IPM/General/LogManager.cls
isc-eneil: Another good question. It should be the full command as passed to the IPM shell, and then the `InitializeLog()` method should strip away everything but the command itself (e.g. install or update) for the purposes of creating a file na,e

ipm-log.schema.json
isc-eneil: I removed this from the header because the way JSON schemas work, is that you have to point to a particular publicly-hosted JSON schema in "$schema" key/value pair at the top of a JSON file, so that will take the place of the "schema" key/value pair with just the version.

src/cls/IPM/General/LogManager.cls
isc-jlechtne: @isc-eneil After our conversation offline yesterday, I think we can get rid of this property altogether. It's only used in InitializeLog() right now for setting the module name, but we have access to that value via `commandInfo("parameters", "module")`

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,118 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
+        // Get IPM version. GetVersion() returns per-module info in versionInfo("zpm") as $lb(namespace, version).
+        $$$ThrowOnError(##class(%IPM.Main).GetVersion("zpm", .versionInfo))
+        set ipmVersion = $listget($get(versionInfo("zpm")), 2)
+
+        // Write TXT header
+        set txtStream = ##class(%Stream.FileCharacter).%New()
+        set txtStream.Filename = ..TxtFilePath
+        do txtStream.WriteLine("IPM Version: " _ ipmVersion)
+        do txtStream.WriteLine("Command: " _ ..Command)
+        do txtStream.WriteLine("")
```
isc-jlechtne: This extra blank line is a mismatch from the design doc example. Not sure if that was intentional or not

isc-jlechtne:
Looks good to me overall. Only change I think is necessary before merging to the main logging-framework branch is removing the module property

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,118 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
-    // TODO: Add the message to the LogFile(s) going in that instance in the same namespace
+/// Write a log entry to both the JSON Lines and TXT files.
+/// @Argument level         The log level: TRACE, INFO, WARNING, or ERROR
+/// @Argument message       The human-readable log message to append
+Method WriteLogLine(level As %String, message As %String) As %Status
+{
+    try
+    {
+        if '$listfind($listfromstring(..#logLevels), level) {
+            do ..WriteLogLine("Log has incorrect logLevel "_level_". Please ensure log level is one of: "_..#logLevels_$$$NL_"Attempted message to log was: "_message, "ERROR")
```
isc-jili: The signature to WriteLogLine() is level, message but the params passed in here are level, message

src/cls/IPM/General/LogManager.cls
```diff
@@ -12,20 +12,118 @@ Property Messages As list Of %IPM.General.LogManager.Message;
...
+        // Namespaces are case-insensitive; '%' (e.g. %SYS) is path-hostile and can't collide once mapped to '_'.
+        set normalizedNamespace = $translate($zconvert(..%namespace, "L"), "%", "_")
+
+        // Create log directory: <IRIS-mgr>/<namespace>/ipm-logs/
+        set mgr = ##class(%Library.File).ManagerDirectory()
+        set logDir = ##class(%File).NormalizeDirectory(mgr _ normalizedNamespace _ "/ipm-logs/")
+        $$$ThrowOnError(##class(%File).CreateDirectoryChain(logDir))
+
+        // Set file paths with timestamp
+        set baseFileName = commandInfo _ "_" _ module.Name _ "_" _ ..TimeStampStart
```
isc-jili: Is commandInfo actually a string? I thought it was an indexed array or something

isc-jili (changes requested):
@isc-eneil looks good! Left some small comments

